PR #2, opened 2023-12-19
feat: add person as delegated type

al3rez:


krschacht (approved):
👍 solid start! Let’s soon add basic tests with minitest. We can do that in a subsequent PR. From what I see so far it would just be creating a user, asserting it’s a type of Person. Or maybe asserting the fixtures are properly associated to one another (user / person and tombstone / person)

also, we should encrypt that password field soon using the new acriverecoed column encryption. But again, later PR. This one looks good to go!

al3rez:
> 👍 solid start! Let’s soon add basic tests with minitest. We can do that in a subsequent PR. From what I see so far it would just be creating a user, asserting it’s a type of Person. Or maybe asserting the fixtures are properly associated to one another (user / person and tombstone / person)
> 
> also, we should encrypt that password field soon using the new acriverecoed column encryption. But again, later PR. This one looks good to go!

it's better to not use fixtures, they become headache very quickly.

krschacht:
@al3rez I’ve had very good luck using fixtures. As we start to use them I can help give guidance of what pitfalls to avoid. I’ve seen them go wrong too on projects that didn’t set them up well. It’s all about creating fixtures in a way that tells a logical story so that it’s easy to follow the relationship between the core entities, and it’s about not trying to create fixtures for every case — just for the common cases. Then individual tests create one-off records as needed and/or they modify existing fixtures as needed. When done this way, fixtures save a lot of extra boilerplate when writing tests. We just need to get them set up well.

PR #10, opened 2023-12-25
Stub out the models for OpenAI API interface

krschacht:
The core models within the OpenAI API are: Assistant, Thread, Message, File, Run, and Step. I tried to mirror all these names, but Thread and File are reserved words in ruby so I renamed "Thread" to `Conversation` and I renamed File to `Document`.

For each of these, I looked at the full API response to determine how they named each of the attributes and what the associations are between them. I generated these models with scaffolding such as this (I didn't record all the commands I did):

```
bin/rails g scaffold Assistant user:references model:string name:string description:string instructions:string tools:string
bin/rails g scaffold Conversation user:references assistant:references title:string
bin/rails g scaffold Message conversation:references role:string text:string document_id:integer run_id:string
bin/rails g scaffold Document user:references assistant:references message:references filename:string purpose:string bytes:integer
...
```

I did a video walkthrough of this PR to unpack why I made some of the decisions that I did. Give this a listen and let me know if you have any thoughts on this!
https://share.getcloudapp.com/v1uBKvxO


app/assets/builds/tailwind.css
```diff
@@ -1 +0,0 @@
...-light{font-weight:300}.font-semibold{font-weight:600}.text-black{--tw-text-opacity:1;color:rgb(0 0 0/var(--tw-text-opacity))}.text-black\/90{color:#000000e6}.text-gray-300{--tw-text-opacity:1;color:rgb(209 213 219/var(--tw-text-opacity))}.text-teal-600{--tw-text-opacity:1;color:rgb(13 148 136/var(--tw-text-opacity))}.text-white{--tw-text-opacity:1;color:rgb(255 255 255/var(--tw-text-opacity))}.shadow-md{--tw-shadow:0 4px 6px -1px #0000001a,0 2px 4px -2px #0000001a;--tw-shadow-colored:0 4px 6px -1px var(--tw-shadow-color),0 2px 4px -2px var(--tw-shadow-color)}.shadow-md,.shadow-sm{box-shadow:var(--tw-ring-offset-shadow,0 0 #0000),var(--tw-ring-shadow,0 0 #0000),var(--tw-shadow)}.shadow-sm{--tw-shadow:0 1px 2px 0 #0000000d;--tw-shadow-colored:0 1px 2px 0 var(--tw-shadow-color)}.hover\:bg-black\/80:hover{background-color:#000c}.hover\:bg-gray-400:hover{--tw-bg-opacity:1;background-color:rgb(156 163 175/var(--tw-bg-opacity))}.hover\:bg-gray-700:hover{--tw-bg-opacity:1;background-color:rgb(55 65 81/var(--tw-bg-opacity))}.hover\:bg-teal-700:hover{--tw-bg-opacity:1;background-color:rgb(15 118 110/var(--tw-bg-opacity))}.focus\:outline-none:focus{outline:2px solid #0000;outline-offset:2px}
```
krschacht, line 1: This was in our gitignore, but somehow this file was also within our repo. I must have accidentally added it at one point so removing it from git history but it'll still be in our dev environments

Gemfile
```diff
@@ -29,7 +29,7 @@ gem "stimulus-rails"
 gem "tailwindcss-rails"
 
 # Build JSON APIs with ease [https://github.com/rails/jbuilder]
-gem "jbuilder"
+# gem "jbuilder"
```
krschacht, line 32: For now we are not anticipating any JSON API replies to our controller endpoints. Having this in here clutters our controllers and it causes scaffold to generate `*.jbuilder` files

app/models/document.rb
```diff
@@ -0,0 +1,7 @@
+class Document < ApplicationRecord
+  belongs_to :user
+  belongs_to :assistant, optional: true
+  belongs_to :message, optional: true
+
+  enum purpose: %w[ fine-tune fine-tune-results assistants assistants_output ].index_by(&:to_sym)
```
krschacht, line 6: I find it easier to deal with enums when the underlying field is a string rather than an integer. It makes declaring them a little more gnarly with the index_by() thingy, but besides that, this is just a normal activerecord enum field.

al3rez (approved):

PR #13, opened 2023-12-27
First pass at pixel-perfect chat experience

krschacht:
This is a design of the core home/chat experience. The core things that it accomplishes:

- [x] A responsive layout of the core areas of the page
- [x] Color scheme in both light & dark mode
- [x] All the hover states and scrolling sections
- [x] Icons & tooltips
- [x] The tentative addition of the Daisy UI tailwind component library. The only thing currently being used in here is tooltips, so if we don't end up using other components, we can just copy out this one element. But I suspect I'll soon want to start using their modals, drop-down menus, form elements, etc.

The next things I plan to work on, which are not in here. I'll do these in a follow-up PR:

- [ ] There is no horizontal control to explicitly close the left sidebar
- [ ] The text area to type your chat message does not expand vertically as the text starts wrapping.
- [ ] Improve the fonts
- [ ] Get the GPT name drop-down menu working


app/views/conversations/_gpt_sidebar.html.erb
```diff
@@ -0,0 +1,13 @@
+  <div class="h-10 flex pr-2 rounded-lg hover:bg-zinc-400 hover:bg-opacity-25 justify-between items-center group cursor-pointer mb-1 mr-5">
+    <div id="profile" class="ml-2 mr-2 w-7 h-7 bg-white rounded-full"></div>
+    <div class="flex-1 text-white truncate">
+      <%= title %>
+    </div>
+    <div class="<%= defined?(selected) && selected ? 'flex' : 'hidden' %> group-hover:flex gap-2 pl-2">
+      <% if !defined?(settings) || settings %>
+        <%= icon "ellipsis-horizontal", variant: :micro, size: 18, class: "inline ml-1 text-gray-400 invisible group-hover:visible hover:text-white cursor-pointer", title: "More" %>
+      <% end %>
```
robacarp, line 9: I don't understand this `if !defined?(settings) || settings` -- to me that will always be true. If settings isn't defined you have `if ! false || nil` which is true. If settings is defined, you have `if ! true || obj` which is true.

app/views/conversations/_gpt_sidebar.html.erb
```diff
@@ -0,0 +1,13 @@
+  <div class="h-10 flex pr-2 rounded-lg hover:bg-zinc-400 hover:bg-opacity-25 justify-between items-center group cursor-pointer mb-1 mr-5">
+    <div id="profile" class="ml-2 mr-2 w-7 h-7 bg-white rounded-full"></div>
+    <div class="flex-1 text-white truncate">
+      <%= title %>
+    </div>
+    <div class="<%= defined?(selected) && selected ? 'flex' : 'hidden' %> group-hover:flex gap-2 pl-2">
```
robacarp, line 6: > `<%= defined?(selected) && selected ? 'flex' : 'hidden' %>`

This is one area where tailwind gets messy.

app/views/conversations/show.html.erb
```diff
@@ -1,15 +1,100 @@
-<div class="mx-auto md:w-2/3 w-full flex">
-  <div class="mx-auto">
-    <% if notice.present? %>
-      <p class="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-lg inline-block" id="notice"><%= notice %></p>
-    <% end %>
+<div class="flex flex-row h-screen">
```
robacarp, line 1: Since `flex-row` is the default, I don't think I've ever seen it unless it was part of some responsive UI pivot between column and row layouts. Why include it?

app/views/conversations/show.html.erb
```diff
@@ -1,15 +1,100 @@
...
+
+    <div id="narrow-header" class="flex-none h-14 text-lg text-center bg-white dark:bg-zinc-800 md:hidden">
+      <div class="flex-1 mb-2 border-b border-zinc-300 dark:border-zinc-600 text-black dark:text-white dark:bg-graphite flex items-center justify-center">
+        <%= render 'gpt_header', conversation: @conversation %>
+      </div>
+    </div>
+
+    <div id="right-content-container" class="flex flex-col flex-grow overflow-auto">
+      <div id="right-content" class="flex-grow overflow-auto">
+        <div class="w-full md:max-w-none lg:max-w-[700px] xl:max-w-[810px] mx-auto mt-2">
```
robacarp, line 69: In general specifying widths like these makes me a little suspicious. Tailwind has a weird falling-over point where they don't specify sizes above a certain threshold, but I think they do it on purpose.

Regardless, does it make sense to add any of the sizes you've hard coded here into our tailwind config so it'll just generate classes for us?

app/views/conversations/show.html.erb
```diff
@@ -1,15 +1,100 @@
...
+          <turbo-frame id="messages" src="<%= conversation_messages_path(@conversation) %>">
+            Loading...
+          </turbo-frame>
+        </div>
+      </div>
+
+      <div id="right-footer" class="flex-none flex flex-col relative">
```
robacarp, line 76: I think I would also break this out into it's own partial. 

If it doesn't need an id for targeting purposes, that can be removed.

app/views/conversations/show.html.erb
```diff
@@ -1,15 +1,100 @@
...
+        <div class="text-center absolute -top-14 left-1/2 -ml-9">
```
robacarp, line 77: Even without tailwind a negative margin scares me a little. What's up?

app/views/messages/_message.html.erb
```diff
@@ -1,22 +1,27 @@
-<div id="<%= dom_id message %>">
-  <p class="my-5">
-    <strong class="block font-medium mb-1">Conversation:</strong>
-    <%= message.conversation_id %>
-  </p>
+<div id="<%= dom_id message %>" class="ml-9 md:ml-8 sm:ml-3 mb-5 group">
```
robacarp, line 1: I think I'd move your `mb-5` up to the flex container as `gap-5`. What do you think?

I think your `md:` and `sm:` classes should be reversed, so that the medium one overrides the small one. But more generally, splitting hairs over the left margin at different breakpoints has me wondering what this looks like through the different sizes.

robacarp (commented):
A few comments, nothing major. Sorry for the emailspam -- I forgot to start a review until half way through.

Stuff like this is often easier to review over a call and screenshare, because there's a lot of context for how these classes get applied. I took a stab, keeping an eye out for things I usually end up refactoring later, but maybe we can do a huddle and take a look?

app/views/conversations/_gpt_sidebar.html.erb
krschacht: Yes, but when settings is equal to false you have: `if ! true || false`

This logic reads pretty poorly. I should fix. Basically, there is one case where I want to suppress the settings ellipsis icon. In that case, I pass in: `locals: { settings: false }`

I know the right fix. Just yesterday I learned that rails 7 now has an elegant way to define default values for locals! I’ll clean this up. https://blog.kiprosh.com/allow-template-to-set-strict-locals/

app/views/conversations/_gpt_sidebar.html.erb
krschacht: Agreed, I would like a more elegant syntax for conditionally including a single value. I tried branching on the overall div, meaning:

```
if ….
   div
else
   div
```

But it’s too much duplication when I have 10 classes and I’m only varying a single one. It’s too easy for them to drift over time. Actually, maybe a better syntax is:

1. Using strict locals so that `selected` always exists and is initialized with a default value if it’s undefined.
2. Rewriting this line to a single override: `<div class=“flex <%=` !selected && `hidden’` %> …`

This works because `hidden` will override `flex`, but there are still going to be cases where I’ll need a tertiary operator. I don’t have a solution for that.

krschacht: Or, maybe I just write a custom tertiary helper to make it a bit simpler to inline these:

<%= if_true(selected, “flex”, “hidden”) %>

app/views/conversations/show.html.erb
krschacht: Good catch. I’m so used to react native, which flips the default to flex-col so I’m still in the habit of being explicit on flex-row. :)

app/views/conversations/show.html.erb
```diff
@@ -1,15 +1,100 @@
...
-    <% end %>
+<div class="flex flex-row h-screen">
 
-    <%= render @conversation %>
+  <div id="left-column" class="flex-col h-screen w-[260px] bg-black overflow-auto hidden md:flex pl-3 scrollbar-hide">
 
-    <%= link_to "Edit this conversation", edit_conversation_path(@conversation), class: "mt-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
-    <div class="inline-block ml-2">
-      <%= button_to "Destroy this conversation", conversation_path(@conversation), method: :delete, class: "mt-2 rounded-lg py-3 px-5 bg-gray-100 font-medium" %>
+    <div id="left-content-container" class="flex flex-col flex-grow overflow-auto scrollbar-hide">
```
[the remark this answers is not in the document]
krschacht: Yes, I’ll do that soon. When I’m doing big iteration it’s nice to spread the HTML across fewer files, but once I wrap up the last few features then I’ll do a pass to better organize the code and maybe improve readability a bit. I’ve been starting to use “id” on elements almost like comments.

app/views/conversations/_gpt_sidebar.html.erb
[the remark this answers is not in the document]
krschacht: Agreed, that’s much better

app/views/conversations/show.html.erb
krschacht: One thing that’s nice about breaking it into a partial is that it gets a name. Then we don’t need an id. But I’ve generally not created partials unless there was some bit of HTML that was being repeated just because it’s harder to refactor the HTML structure of a page when it’s split across more files. Have you had the bias of creating partials for logical units, even if they aren’t being reused?

app/views/conversations/show.html.erb
krschacht: Response design isn’t my super strength. Is there a way around it explicitly specifying widths? The CSS issue I’m tying to solve is that ChatGPT varies the column size at different breakpoints. It looks pretty nice, and I’m in the mode of “perfectly copy, don’t try to improve (yet)” so I am just blindly making everything line up perfectly. So I think I need to explicitly specify the width.

I didn’t realize you could put sizes in the tailwind config, or maybe I did see that in the docs but I haven’t done it yet for anything. Since this was the only place I was using these, it didn’t feel worth creating a custom utility class for it.

app/views/messages/_message.html.erb
krschacht: Good catch

app/views/messages/_message.html.erb
```diff
@@ -1,22 +1,27 @@
...
+  <div class="flex">
+    <div class="w-6 flex flex-col items-center ml-1">
+      <div class="w-6 h-6 bg-black dark:bg-white rounded-full mt-1"></div>
+    </div>
```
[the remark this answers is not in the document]
krschacht: That inner div is going to be replaced with the user’s and assistant’s profile pic. Right now the “bg-black rounded-full” is just a hack way to get me a circle, as a placeholder, until we have a rails attribute to drop in there.

app/views/conversations/_gpt_sidebar.html.erb
robacarp: I considered both of these suggestions but anything I could think of felt out of scope. I don't know if the value of a helper is really worth the effort yet.

app/views/conversations/show.html.erb
robacarp: Every part of tailwind I know of [has the ability to customize your theme](https://tailwindcss.com/docs/width#customizing-your-theme). There's a lot of design value to sticking to a strict theme because it creates a lot of consistency with the design. Tailwind overrides then stick out as a place that something is going on that could look different than the rest of the site.

Totally fine rolling forward.

app/views/conversations/show.html.erb
krschacht: This was tricky, but I couldn’t figure out another way. Are negative margins inherently bad? Is there a “gotcha” with it that I may not be considering?

Basically, when the center column of chat messages is not scrolled to the very bottom, a down arrow appears and it floats above the chat input.

So this down arrow isn’t really part of the chat input but it’s position is determined relative to the chat input. It was really hard to figure out how to make this part of the overall scrolling area but be anchored at the bottom. But now that I write this out, I just thought of another way I could try.

app/views/conversations/show.html.erb
robacarp: I don't think I have a hard rule. When something is "layout-ish" like a two column thing like you have here, I'm way more likely to break it out.

That makes me think, perhaps this should be a dedicated layout file which gets loaded inside of the application frame for the Conversation controller?

app/views/messages/_message.html.erb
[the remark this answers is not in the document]
robacarp: haha gotcha.

app/controllers/messages_controller.rb
```diff
@@ -1,51 +1,56 @@
 class MessagesController < ApplicationController
   skip_before_action :authenticate_user!
-  before_action :set_message, only: [:show, :edit, :update, :destroy]
+  before_action :set_conversation
+  before_action :set_message, only: %i[show edit update destroy]
```
robacarp, line 4: The cop is complaining about this, but I actually like it.

app/views/conversations/_gpt_sidebar.html.erb
robacarp: Thaaaaaaaaaaaaaaaaat is interesting, thanks for the link. I'm not usually a fan of magic comments, but I can get behind the vision there. This is a weakness of the rails View == Template pattern, I guess. It's interesting that it says "define the local variables a template can accept."

Historically I've done [something like this](https://github.com/hostedgpt/hostedgpt/pull/58/files#diff-3013a9e38dc22cf8c4ec490c15a17d1519d0fed4935531f8605bc7309a5cfdb5R1):

```erb
<% scroll_into_view ||= false %>
<div
  data-controller="conversation-message"
  data-conversation-message-scroll-into-view-value="<%= scroll_into_view %>"
>
  <%= message.content_text %>
</div>
```

app/views/layouts/application.html.erb
```diff
@@ -6,14 +6,13 @@
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
     <%= stylesheet_link_tag "tailwind", "inter-font", "data-turbo-track": "reload" %>
-    <%= stylesheet_link_tag "application", 'https://fonts.googleapis.com/css?family=Inter', "data-turbo-track": "reload" %>
+    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
+    <%= stylesheet_link_tag "https://cdn.jsdelivr.net/npm/daisyui@4.6.0/dist/full.min.css" %>
```
robacarp, line 10: It looks like this is a tailwind widgets library, but I don't think you're using it?

PR #21, opened 2024-01-08
progress commit

al3rez:


app/services/conversation_ai/stream_handler.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      new_content = format_content(@reply, new_content)
+
+      if new_content.present?
+        @reply.content += new_content
+        @reply.broadcast_updated(new_content)
+      end
+
+      @reply.save! if finish_reason.present?
+    end
```
krschacht, line 19: I haven't fully wrapped my mind around this separate of concerns. The fact that you're passing in an active record model and then saving it here smells a bit off.

It does sometimes make sense to create models that are not backed by active record. However, that's typically when a model has a really isolated concern. But when this model is so tightly coupled to an activerecord model means that these should probably be methods on the Reply model itself, or maybe there is indeed a separate of concerns but you haven't gotten the boundary layer right yet. A clean boundary layer would mean not passing in an activerecord model and saving it here.

I'll keep reading and see if I can have a better proposal for you. I don't quite get what a "StreamHandle" is and what a "TokenHandle" is

app/services/conversation_ai.rb
```diff
@@ -10,10 +7,11 @@ def initialize(user, note)
     @note = note
     @chat = note.chat
     @client = OpenAI::Client.new(access_token: @user.openai_key)
+    @token_handler = TokenHandler.new(@chat)
   end
```
krschacht, line 11: The original class I had for OpenAI stuff was not great, but at least it was nicely encapsulated and had no coupling to any specific model. For example, we could use it for a different purpose other than creating a reply.

It's not a good practice to be passing in models like this. The thing to optimize for is: (1) add all methods that pertain to the model, into the actual model itself. There's no reason to shy away from too many methods. You can just put them in a concern if there are a bunch of related methods, but don't create a new class for them. Or (2) stand-alone models that aren't backed by activerecord but have some clear mental container where we definitely need to instantiate something, manipulate it, and hide away a bunch of complexity — but it has nothing to do with the database.

PR #44, opened 2024-01-09
Fixes Render.com integration

### LINKED ISSUE #43: Render.com issues
From render.com

> A render.yaml file was found, but there was an issue.
> 
> services[0].envVars[1]
>     must be from group or have a key and value
> services[0].envVars[2]
>     cannot simultaneously specify fields value and sync

robacarp:
fixes #43 

Also fixes the need to copy/paste a master key in by asking render.com to generate a value when the app is cloned.

### DIFF
--- a/render.yaml
+++ b/render.yaml
@@ -4,6 +4,12 @@ databases:
     user: hostedgpt
     plan: free
 
+envVarGroups:
+- name: generated-secrets
+  envVars:
+    - key: rails_master_key
+      generateValue: true
+
 services:
   - type: worker
     name: soliqueue-worker
@@ -16,10 +22,9 @@ services:
         fromDatabase:
           name: hostedgpt
           property: connectionString
-      - key: RAILS_MASTER_KEY
+      - fromGroup: generated-secrets
       - key: WEB_CONCURRENCY
         value: 2 # sensible default
-        sync: false
   - type: web
     name: hostedgpt
     runtime: ruby
@@ -32,7 +37,6 @@ services:
         fromDatabase:
           name: hostedgpt
           property: connectionString
-      - key: RAILS_MASTER_KEY
-        sync: false
+      - fromGroup: generated-secrets
       - key: WEB_CONCURRENCY
         value: 2 # sensible default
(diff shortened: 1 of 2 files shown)


render.yaml
```diff
@@ -16,10 +22,9 @@ services:
         fromDatabase:
           name: hostedgpt
           property: connectionString
-      - key: RAILS_MASTER_KEY
+      - fromGroup: generated-secrets
```
anandvc, line 25: This is cool! Can you please also update the README.md file to remove that step of copy/pasting the key?

robacarp: Good call. I updated the instructions there a little. We should consider whether forking the repository is necessary for a render deploy -- if not, we can make the process even easier.

render.yaml
```diff
@@ -4,6 +4,12 @@ databases:
     user: hostedgpt
     plan: free
 
+envVarGroups:
+- name: generated-secrets
+  envVars:
+    - key: rails_master_key
+      generateValue: true
+
```
krschacht, line 12: Oh, interesting! @robacarp how does this make it so we don’t have to copy that master.key value?

My assumption was: the master.key value is important and the fix would involve telling the rails app to find the key within the master.key file, not within the ENV. But it seems like you’ve fixed it somehow by telling render to generate a new one? Is there a discrepancy between this newly generated value and the value already stored in master.key?

krschacht:
Lint is failing. :( We really need to remove this ruby linter. I have a GH Issue to go back to the super basic rubocop setup that I initially had. It only enforces a very small number (I think three) rules that I’ve found to actually be universally good. I hate enforcing optional stylistic differences or insignificant white space differences.

al3rez:
> Lint is failing. :( We really need to remove this ruby linter. I have a GH Issue to go back to the super basic rubocop setup that I initially had. It only enforces a very small number (I think three) rules that I’ve found to actually be universally good. I hate enforcing optional stylistic differences or insignificant white space differences.

I already created a PR that reverting it back to previous setup.

al3rez:
@robacarp please update your pr with the latest changes in main and lint issue should be resolved

PR #45, opened 2024-01-09
adds minimally functional page for assistant instructions

### LINKED ISSUE #37: Support custom instructions
Here is the PR with old context:
https://github.com/hostedgpt/hostedgpt/pull/10

### LINKED ISSUE #48: Configure SolidQueue to run in worker thread not web thread

robacarp:
fixes #37 
fixes #48 

- [x] feature is implemented
- [x] controller tests
- [x] system tests

### DIFF
--- a/app/controllers/assistants/instructions_controller.rb
+++ b/app/controllers/assistants/instructions_controller.rb
@@ -1,0 +1,21 @@
+class Assistants::InstructionsController < ApplicationController
+  before_action :set_assistant
+
+  def edit
+  end
+
+  def update
+    @assistant.update!(assistant_params)
+    redirect_to @assistant, notice: "Instructions have been saved."
+  end
+
+  private
+
+  def set_assistant
+    @assistant = Assistant.find(params[:assistant_id])
+  end
+
+  def assistant_params
+    params.require(:assistant).permit(:instructions)
+  end
+end
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -19,6 +19,9 @@
     <%= content_for :head %>
   </head>
   <body>
+    <% if notice.present? %>
+      <p class="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-lg inline-block" id="notice"><%= notice %></p>
+    <% end %>
     <%= yield %>
   </body>
 </html>
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -3,7 +3,7 @@
   <h2 class='text-3xl font-semibold text-black/90'>Welcome back</h2>
     <%= form_with(url: login_path, method: :post, class: "flex flex-col space-y-4 w-80") do |f| %>
     <%= f.email_field :email, id: "email", autofocus: true, placeholder: "Email address", class: "border border-gray-400 rounded text-black p-3"%>
-    <%= f.password_field :password, id: "email", autofocus: true, placeholder: "Password", class: "border border-gray-400 rounded text-black p-3"%>
+    <%= f.password_field :password, id: "password", autofocus: true, placeholder: "Password", class: "border border-gray-400 rounded text-black p-3"%>
     <%= f.submit "Continue", class: 'bg-teal-600 hover:bg-teal-700 rounded-sm font-light text-white p-4 px-4 text-center' %>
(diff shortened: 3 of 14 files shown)


app/controllers/assistants/instructions_controller.rb
```diff
@@ -0,0 +1,21 @@
+class Assistants::InstructionsController < ApplicationController
+  before_action :set_assistant
+
```
krschacht, line 3: @robacarp what’s the rationale behind a dedicated controller just for the instructions? I would have assumed we could reuse the assistants_controller#update.

For example, the “create GPT” page in ChatGPT lets you configure many aspects of an Assistant all at once (name, image, instructions, etc). We’d surely want that to just be a single post to a single endpoint to save all those attributes. So even though this new view is only letting you update this single aspect of a GPT, we can still use the same general #update controller.

robacarp: @krschacht I considered that. I broke out the #edit action because it has a dedicated view associated, with it's own form. That could also be handled within the Assistants controller (#edit_instructions and edit_instructions.html.erb), but my experience has been that staying pedantically true to one-controller-one-idea is better in the long run. [This blog post from 2016](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/) comes to mind.

So I settled onto this approach because, as we have it in the product, the idea of updating an assistant's instructions is different from the idea of updating the assistant's other fields.

krschacht: @robacarp Interesting, let's run with that. I remember that blog post, actually. And this plus other experience led me to half of this conclusion. Specifically, I've had a bias towards: stick with CRUD, any time you want to add a new action like "pending" then it's a sign you need a new controller. But I had not taken the further step of: even if you are within the CRUD paradigm, if there are multiple views that all #update a model, don't use the same controller for those. Create a new controller for each view.

That's a new consideration, but it's one that feels intuitive. For example, I can imagine the redirect after update being different for these two views. If they're hitting the same endpoint, then you'll end up adding a conditional branch to decide what to do after the update. If they're in separate controllers, then there is a natural separation of concerns.

To state that more broadly: if you found a reason to create two separate views, then it's safe to assume the needs of these views are going to diverge further over time so prepare for that by creating separate controllers.

app/views/layouts/application.html.erb
```diff
@@ -19,6 +19,9 @@
     <%= content_for :head %>
   </head>
   <body>
+    <% if notice.present? %>
+      <p class="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-lg inline-block" id="notice"><%= notice %></p>
+    <% end %>
```
robacarp, line 24: I moved this flash notice up to the layout because messages were getting swallowed. It looks terrible, but I'm vaguely aware that @krschacht is working on styles? I'm happy to spruce it up if needed.

Anyway, with the notices getting swallowed, I wasn't able to write a helper method #login_user for the system integration tests and know that the login succeeded.

I'm happy to leave this as-is (minimally working) or to finish the refactor and make the notifications work system wide.

app/views/sessions/new.html.erb
```diff
@@ -3,7 +3,7 @@
   <h2 class='text-3xl font-semibold text-black/90'>Welcome back</h2>
     <%= form_with(url: login_path, method: :post, class: "flex flex-col space-y-4 w-80") do |f| %>
     <%= f.email_field :email, id: "email", autofocus: true, placeholder: "Email address", class: "border border-gray-400 rounded text-black p-3"%>
-    <%= f.password_field :password, id: "email", autofocus: true, placeholder: "Password", class: "border border-gray-400 rounded text-black p-3"%>
+    <%= f.password_field :password, id: "password", autofocus: true, placeholder: "Password", class: "border border-gray-400 rounded text-black p-3"%>
```
robacarp, line 6: Lil bug prevented me from writing the system login_user method.

config/puma.rb
```diff
@@ -33,5 +33,3 @@
 
 # Allow puma to be restarted by `bin/rails restart` command.
 plugin :tmp_restart
-
-plugin :solid_queue
```
robacarp, line 37: Fixes #48

app/views/layouts/application.html.erb
krschacht: This seems like a fine place for now

krschacht (approved):
Looks good

PR #47, opened 2024-01-10
Fixes Render.com secret key & worker config

### LINKED ISSUE #43: Render.com issues

robacarp:
Fixes #43 
Supplants #44 - replacing with a pull from a branch on this repo rather than a fork, because my local git refused to let me update upstream and I gave up.


render.yaml
```diff
@@ -4,6 +4,16 @@ databases:
     user: hostedgpt
     plan: free
 
+envVarGroups:
+- name: generated-secrets
+  envVars:
+    - key: SECRET_KEY_BASE
+      generateValue: true
```
robacarp, line 11: From @krschacht's [original comment](https://github.com/hostedgpt/hostedgpt/pull/44#discussion_r1446820695):

> Oh, interesting! @robacarp how does this make it so we don’t have to copy that master.key value?

The rails credential file is new to me. The gist is, if deployment secrets are a management problem for your team, you can use an encrypted credential file to make it easier. In the case where you use the credentials file, rails warns:

image

I've deleted the master.key file from our repo. I believe the intent with that file is that it would be copied post-deploy from some other file store, or otherwise hosted on long-living servers somewhere. With docker, we don't really have either of those anyway.

In this repo, our intent is to build a forkable project. Each incarnation of our project will have it's own credentials and they should be unique -- even declaring the credentials on our own from the start is a risk because people _don't change the defaults_. Render.com has a nice little override for this which generates a value for you when you stand up a server from a blueprint. A neat idea I've never come across, but I like it.

So I've also deleted the credentials file and replaced the rails master key with the only value which was in the credentials file, the secret key base.

README.md
```diff
@@ -14,13 +14,13 @@ HostedGPT is an open source project that provides all the same baseline function
...
 [![Deploy to Render](https://render.com/images/deploy-to-render-button.svg)](https://render.com/deploy)
 
-3. When you are taken to Render.com confirm that the title of the page says `You are deploying from a render.yaml file for _____` with your forked repo listed in the blank.
-4. Provide a unique Blueprint name such as "hostedgpt-<yourname>".
-5. Copy value from the file `config/master.key` and paste it as the value for the RAILS_MASTER_KEY on the page. Then click "Apply".
-6. Wait for the hostedgpt database and web service to be deployed. After they are, click "Dashboard" at the top of the Render screen.
-7. You should see two "Service Names" called **hostedgpt**, click the one that is of type "Web Service"
-8. On the details screen, click the URL that looks something like _hostedgpt-XXX.onrender.com_
+3. Find hostedgpt in the list of repositories and click **Connect**
+5. Provide a unique Blueprint name such as "hostedgpt-<yourname>".
```
anandvc, line 22: Minor numbering typo: 4 vs 5.

anandvc (approved):
👍 Looks good! Feel free to update the numbering in the README before merging.

PR #52, opened 2024-01-11
Updates heroicon / replaces it with rails-heroicon

robacarp:
The gem "heroicon" hasn't been updated in >18mo, and the heroicon pack seems to have changed a lot since then.

Another gem "rails-heroicon" seems to share a lot of the code but is actually bundled with a railtie and has carried updates as recently as 2 weeks ago. The interface is identical from the views, but it comes with all the names for icons which match the current heroicon website.

I've removed the heroicon helper and initializer. Neither are required by rails-heroicon.

Fixes this warning message spamming the console:

image


app/views/shared/_settings_modal.html.erb
```diff
@@ -11,7 +11,7 @@
     <p>Settings</p>
   </button>
   <a href="/logout" class='flex space-x-2 p-2 hover:bg-black/80 mx-2 rounded text-white text-sm items-center'>
-    <%= heroicon "logout" , options: { class: 'h-5 w-5 text-gray-300' } %>
+    <%= heroicon "arrow-right-start-on-rectangle" , options: { class: 'h-5 w-5 text-gray-300' } %>
```
robacarp, line 14: There now no icon called logout and there was some [discussion](https://github.com/tailwindlabs/heroicons/discussions/457) about it all. Apparently people disagree with what looks like a logout arrow and what doesn't 😂

krschacht (approved):

app/views/shared/_settings_modal.html.erb
al3rez: i intentionally added the old version since it had icons that looked similar to old chatgpt and yeah that icon is different. :D

krschacht: Yea, I was noting that as well. But the odd thing is that it does look like the community is backing the v2 a little stronger than v1 (keeping libraries more updated, etc). The creator of v2 also talks about it as if it's his new baby — it's the new "modern" icon library. So I did a pass in my branch to try and move everything to the v2 icons and I got it looking pretty nice so we can just stick with it now.

PR #54, opened 2024-01-12
Implements conversation/messages shallow routing

robacarp:
Rails provides a concept called [Shallow Nesting](https://guides.rubyonrails.org/routing.html#shallow-nesting) which I really love. 

This implements shallow nesting with conversations/messages. I've updated the views and controllers -- I think I got all of them. 

-----------------------------------------------

# Previously, as an RFC

This proposal is something I worked with years ago and found it really helpful, although it was an API and not a user-facing website. The idea is simple: when you have an implicit "current user" scope, it should be explicit. 

In an app like this, with traditional shallow routing, we'd have these urls:

- conversations (conversations#index)
- conversations/:id (conversations#show)
- conversations/:id/messages (messages#index, scoped to conversation_id)
- messages/:id (messages#show)

Generally speaking, you don't have routes like these:
- messages (messages#index, without the conversation_id scope)

And if you do, you get yourself in a rats nets of scoping conditionals really quickly. 

But most apps have an implicit scope which is ever-present, the current user. In order to avoid URL guessing attacks, it's important to create and access resources via the `Current.user` -- `Current.user.conversations.new` and `Current.user.conversations.find(id: params[:id])` etc.

So, again years ago, we found a collision where we sometimes wanted that namespace and sometimes we didn't. It got messy. So our solution was: the "my" namespace. It produced routes like this:

- my/conversations
- conversation/:id
- conversation/:id/messages
- etc

The idea is, whenever you are implicitly scoping a collection to the current user, put it in the "my" namespace.

Some of this is premature optimization for this project. I'm not sure, maybe all of it is. 

Comments welcome 🌴


app/views/conversations/_form.html.erb
```diff
@@ -1,4 +1,11 @@
-<%= form_with(model: conversation, class: "contents") do |form| %>
+<% 
+submit_path = if conversation.new_record?
+  my_conversations_path
+else
+  edit_my_conversation_path(conversation)
+end
```
robacarp, line 6: This got weird. Rails routing magic failed here and maybe this alone is a sign that the idea is too far from happy-path to go forward.

Other options include:
- different ways to organize the partials, eg put the fields in a partial and not the form header
- not using partials -- my experience is the create and edit views rapidly diverge anyway

config/routes.rb
```diff
@@ -4,8 +4,16 @@
...
-  resources :messages
+  namespace "my" do
+    resources :conversations, only: [:new, :create, :index]
+  end
+
+  shallow do
+    resources :conversations, except: [:index, :new, :create] do
+      resources :messages
+    end
+  end
```
robacarp, line 15: This produces these routes:

```
                        my_conversations GET    /my/conversations(.:format)                                                                       my/conversations#index
                                         POST   /my/conversations(.:format)                                                                       my/conversations#create
                     new_my_conversation GET    /my/conversations/new(.:format)                                                                   my/conversations#new
                   conversation_messages GET    /conversations/:conversation_id/messages(.:format)                                                messages#index
                                         POST   /conversations/:conversation_id/messages(.:format)                                                messages#create
                new_conversation_message GET    /conversations/:conversation_id/messages/new(.:format)                                            messages#new
                            edit_message GET    /messages/:id/edit(.:format)                                                                      messages#edit
                                 message GET    /messages/:id(.:format)                                                                           messages#show
                                         PATCH  /messages/:id(.:format)                                                                           messages#update
                                         PUT    /messages/:id(.:format)                                                                           messages#update
                                         DELETE /messages/:id(.:format)                                                                           messages#destroy
                       edit_conversation GET    /conversations/:id/edit(.:format)                                                                 conversations#edit
                            conversation GET    /conversations/:id(.:format)                                                                      conversations#show
                                         PATCH  /conversations/:id(.:format)                                                                      conversations#update
                                         PUT    /conversations/:id(.:format)                                                                      conversations#update
                                         DELETE /conversations/:id(.:format)                                                                      conversations#destroy
```

al3rez:
I get the idea of structuring controllers like in your example but, this setup can get a bit tricky in more complex scenarios. Have you considered using a <>Scoped approach instead, similar to what Basecamp does with 'UserScoped', 'AdminScoped', etc.? It might offer a cleaner and more scalable solution, especially when dealing with different user roles and permissions. Happy to see how this progresses and always up for a chat.

krschacht:
I think that the “my/“ nesting is premature, but nesting messages under conversations is probably right.

In general, I shy away from nesting resources unless it really brings a lot of benefits. Almost invariably when I nest routes it starts to feel like it complicates things unnecessarily. But I’ve always suspected that I just haven’t fully wrapped my mind around all the rails route helpers (e.g. form construction, named paths, etc).

I was hesitating even to nest messages within conversations, but that one is probably “right”. I actually really like the “shallow” compromise! I just had to ask chatgpt what that was because that was new to me. But it’s great to minimally nest things so now we only deal with nesting in the cases we absolutely need the conversation_id. That’s perfect.

But I think the main reason for not doing the “my” nesting is just because we don’t actually have the need to request `conversations` independent of a user. Maybe this will come up, and I can see how within the context of an API design, this would likely come up sooner. But within this app, conversations (and pretty much all other data) will be sensitive so it’ll be tied to some “owner” and all requests will need to be made within the context of that person.

robacarp:
@al3rez @krschacht I think you're right. It's premature to put in a my/ namespace. I think I could envision that if we had multiple "owners" of conversations, but that's not the case now and might not ever be.

If we head that direction eventually we can use the Scoped approach, I like that.

For now, I've just moved the routes around so that Conversations and Messages are shallow-nested.

robacarp:
I'm going to take the work here and roll it into my pull request for #24

PR #58, opened 2024-01-12
Automatically scroll to the bottom of the conversation when a message arrives

### LINKED ISSUE #24: Automatically scroll down as new conversations come in

robacarp:
I got a bit into the weeds cleaning things up here in order to really arrive at what is a simple solution. Some of that work is redundant with work @krschacht did on #13 (eg, `Current.user.conversations` stuff). I tried to stay out of the CSS weeds but there were times it was difficult to read without making some changes. I'll take care of the merge when his goes in first.

I learned a lot about turbo and stimulus in this process, but I don't think it's all gelled yet. 

The commits include my routing changes, but honestly I think most of those routes will eventually go away in favor of a slimmer interaction model. For example, I don't think a Messages#update action needs to be here -- what does it mean to update a message when you have an ongoing exchange with a third party? Should that be allowed? Does the new api interaction model even support that? Etc.

fixes #24 

@krschacht I took care of merging your #61 into this, so it should go in first.


app/javascript/controllers/conversation_message_controller.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default class extends Controller {
+  static values = {
+    scrollIntoView: { type: Boolean, default: false }
+  }
+
+  connect() {
+    if (this.scrollIntoViewValue) {
+      this.element.scrollIntoView(true, { behavior: "smooth" })
+    }
```
robacarp, line 13: The way I've implemented this, the page doesn't scroll to the bottom when it first loads. That's probably the wrong way to go about it, but I don't think I really like it when I open a webpage and it jumps straight to the bottom. I left it as-is because it feels like there's more to think about.

krschacht:
I haven't reviewed the PR but @robacarp, one quick comment: In chat GPT you can edit a message way back in your conversation history. I do it quite often when I'm talking through a complex problem. Sometimes we'll end up in a dead-end as I'm trying to solve a problem, so I scroll back to the earlier spot in the conversation where things were good and then edit my message to take the brainstorming in a different direction.

krschacht:
@robacarp my PR is now merged in. Sorry that was so long lived! I'll keep them shorter.

app/javascript/controllers/conversation_message_controller.js
krschacht: @robacarp Hmm, that’s notable. Now that you mention it, i’m browsing around chatgpt and as I click to different conversations in the sidebar, they don’t load the full thing and then scroll me all the way down. Frankly, some of them are really long! Somehow, as the page loads you’re already at the bottom of the page and can scroll up.

I tried a bunch of long conversations where I scrolled up really quickly to see if they were doing progressive loading as I scroll back in time and I was never able to catch it doing that.

But maybe we’re thinking about this wrong? Maybe we need to reverse the flex stack direction, reverse the scroll direction, and just implement an old-school infinite scroll which loads progressively, but just in reverse?

robacarp: @krschacht I was trivially able to get this to work as expected. It's snappy too, at least in localhost.

config/routes.rb
```diff
@@ -4,9 +4,12 @@
     patch :instructions, to: "assistants/instructions#update"
   end
 
-  resources :conversations do
-    resources :messages, shallow: true
+  shallow do
+    resources :conversations do
+      resources :messages
+    end
```
krschacht, line 10: The way I saw this explained in the rails guide, you either put shallow: true on the inner route or the outer route. I'm not sure what the `shallow do` block does:
https://guides.rubyonrails.org/routing.html#shallow-nesting

app/views/messages/_message.html.erb
```diff
@@ -1,4 +1,10 @@
-<div id="<%= dom_id message %>" class="ml-9 sm:ml-3 md:ml-8 mb-5 group">
+<%# locals: (message:, conversation:, scroll_into_view: false, request_id: nil) -%>
+<div 
+  id="<%= dom_id message %>"
```
krschacht, line 3: I never knew about that helper; that's nice

app/models/user.rb
```diff
@@ -7,6 +7,7 @@ class User < ApplicationRecord
   validates :password_confirmation, presence: true, on: :create
 
   has_many :chats, dependent: :destroy
+  has_many :conversations, dependent: :destroy
```
krschacht, line 10: it's worth us adding a unit test for destroying a user. it looks like I left that one off. If you look at the last test in assistant_test.rb you'll see how I typically do them. Nothing fancy, and I don't have strong opinions about how we test them, but cascading destroys has bitten me in the past and lead to data corruption when someone inadvertently changes one in the future so it's nice to have coverage on those.

app/models/message.rb
```diff
@@ -10,4 +10,8 @@ class Message < ApplicationRecord
 
   validates :run, presence: true, if: -> { assistant? }
   validates :role, presence: true
+
+  after_create_commit -> {
+    broadcast_append_to conversation, partial: "messages/message", locals: {scroll_into_view: true, conversation: conversation}
```
krschacht, line 15: I love the ergonomics of this, especially the simple `scroll_into_view` to drive the interaction!

app/javascript/controllers/new_message_controller.js
```diff
@@ -0,0 +1,26 @@
...
+
+  disableSubmitButton() {
+    console.log(this.inputTarget.value.length)
+    if (this.inputTarget.value.length < 1) {
+      this.submitTarget.disabled = true
+    } else {
+      this.submitTarget.disabled = false
+    }
+  }
+}
```
krschacht, line 26: It seems a little surprising to have a stimulus controller solely for this form. My instinct would have been to either expand the form controller we already have to make it able to handle some of these extra features like initial focus and submit disable. But it's totally fine to leave this. We can either have separate form conrollers b/c maybe they're specific enough or we can generalize later after we have multiple forms.

app/controllers/messages_controller.rb
```diff
@@ -44,14 +45,14 @@ def destroy
...
+    @conversation = Current.user.conversations.find(params[:conversation_id])
   end
 
   def set_message
     @message = Message.find(params[:id])
   end
 
   def message_params
-    params.require(:message).permit(:role, :content_text)
+    params.require(:message).permit(:content_text)
```
krschacht, line 56: But message does need the role attribute set, right? I thought maybe you modified message.rb to have a default value but I didn't see anything like that. Won't role end up being nil for all newly created messages?

krschacht (approved):
@robacarp this looks great! I left a few minor comments, mostly as questions, but this looks good to go. The only thing I didn't do it fire it up localy.

Oh, and I merged my PR into main after addressing your comments, and then merged main back into this PR so it was fully up to date. I initially messed up a couple aspects of the merge but I believe I fixed them all.

config/routes.rb
robacarp: The docs on shallow nesting provide a lot of options for expression. Near the bottom is the `shallow` method, which they say does the same thing as putting the shallow flag on individual routes:

image

I selected this one because it felt like it would be easier to add more resources under the pattern than adding `shallow: true` to each. Perhaps that's a premature move, since our data model is likely to stay so lean.

app/javascript/controllers/new_message_controller.js
robacarp: I hemmed and hawed about this. Stimulus seems like it'll be at it's best when controllers are tiny and single purposed, but it also seems to encourage reuse. The two conflict in my mind...

I opted for a dedicated controller here because this is the primary interaction widget on the screen and it's likely to be different from other forms we might have in many ways, and it's important to be robust.

app/controllers/messages_controller.rb
robacarp: The message gets the role set [in the controller action](https://github.com/hostedgpt/hostedgpt/pull/58/files/2b2c4ec716b6452c20bc69862347996ea6c8cccc#diff-9b58332ea76756301b1aafeacfb0659d1244ba014947602b0e0bf2d0f2921d26R22), not in the message params. Allowing it to be set in the message params means someone could "impersonate" the assistant by putting `role=assistant` into the form data manually.

In this case it's probably not a security concern, but the notion of setting the role in this controller is implicitly to set it to "user". I favor making that explicit in code rather than submitting a parameter in the form.

app/models/user.rb
robacarp: Done. I also had to add the dependent-destroy for user->assistants, and I added a test for that, too.

PR #59, opened 2024-01-13
Update render.yaml to use smallest paid starter plan

anandvc:
Worker type instances are not supported on the free plan, so use the smallest starter plan that works

Relevant docs: https://docs.render.com/blueprint-spec#instance-types


krschacht (approved):
👍

krschacht:
Actually @anandvc, I just saw the discussion about "Render minimum plan requirement" on basecamp. I just chimed in. Maybe we solve this a different way rather than bumping to the paid plan.

anandvc:
(Convert to Issue)

PR #61, opened 2024-01-13
Further refine design of home experience

krschacht:
- [x] Add a vertical grab handle to explicitly close the left sidebar
- [x] Wire this up and make it actually work
- [x] The text area to type your chat message needs to auto-expand as someone types
- [x] Improve the fonts
- [x] Get the GPT name drop-down menu working
- [x] Make buttons depress when clicked
- [x] Refine the tooltips to disappear when item is clicked

Next steps, which I did not do within this PR:

- [ ] A close look at Safari and Firefox to make sure there are no small pixel differences
- [ ] Make the background of the tooltips white and tweak the glow effect in dark mode, it's garish
- [ ] Try this web UI on tablet & phone and lift some design details from ChatGPT iPad app
- [ ] Clicking a button to do a drop-down menu, clicking again it should close
- [ ] Add transition/animations such as the sidebar opening & closing


app/javascript/controllers/textarea_autogrow_controller.js
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  connect () {
+    this.element.style.overflow = 'hidden'
+    this.onResize = this.resizeDebounceDelayValue > 0 ? debounce(this.autogrow, this.resizeDebounceDelayValue) : this.autogrow
+
+    this.autogrow()
+
+    this.element.addEventListener('input', this.autogrow)
+    window.addEventListener('resize', this.onResize)
```
robacarp, line 18: I think you need to add `.bind(this)` to these event listeners, or `this` will be undefined when the function is called.

app/javascript/controllers/textarea_autogrow_controller.js
```diff
@@ -0,0 +1,39 @@
...
+    this.autogrow = this.autogrow.bind(this)
+  }
+
+  connect () {
+    this.element.style.overflow = 'hidden'
+    this.onResize = this.resizeDebounceDelayValue > 0 ? debounce(this.autogrow, this.resizeDebounceDelayValue) : this.autogrow
+
+    this.autogrow()
+
+    this.element.addEventListener('input', this.autogrow)
```
robacarp, line 17: This event is new to me. It looks like it's less than a year old. Neat!

app/views/conversations/_assistant_header.html.erb
```diff
@@ -0,0 +1,11 @@
+<div class="dropdown">
+  <div tabindex="0" role="button" class="cursor-pointer scale-down inline-flex px-3 py-2 rounded-xl items-center font-medium text-lg outline-none hover:bg-gray-100 dark:hover:bg-zinc-800">
+    <%= conversation.assistant.name %>
+    <%= icon "chevron-down", variant: :mini, class: 'text-gray-500 ml-[2px]' %>
+  </div>
+
+  <ul tabindex="0" class="dropdown-content z-10 menu p-2 shadow-xl bg-base-100 rounded-box w-52 dark:!bg-zinc-800">
```
robacarp, line 7: I confess I haven't done much with tabindex, and I had to look this pattern up. Neat

app/views/shared/_chat_messages.html.erb
```diff
@@ -9,7 +9,7 @@
             </div>
             <p><%= note.content %></p>
           </div>
-          <%= heroicon "pencil-suqare" , options: { class: 'h-6 w-6' } %>
+          <%= heroicon "pencil-square" , options: { class: 'h-6 w-6' } %>
```
robacarp, line 12: 🔲

app/javascript/controllers/nested_clickable_controller.js
robacarp, line 1: This name is totally opaque to me. For general purpose components like these, should we have a practice of adding a docblock at the top of the file that explains what it does?

app/javascript/controllers/nested_clickable_controller.js
```diff
@@ -0,0 +1,23 @@
+import { Controller } from '@hotwired/stimulus'
+
+export default class extends Controller {
+
+  connect () {
+    this.element.addEventListener('mouseover', this.removeParentClickable)
+    this.element.addEventListener('mouseout', this.addParentClickable)
```
robacarp, line 7: Are these events working without calling `.bind(this)` on these? Maybe I'm missing something about some magic stimulus does for this...

app/javascript/controllers/click_toggle_controller.js
```diff
@@ -0,0 +1,24 @@
...
+
+export default class extends Controller {
+  static classes = [ "flippable" ]
+  static targets = [ "trigger", "destination" ]
+
+  connect () {
+    this.triggerTargets.forEach(triggerElement => {
+
+      triggerElement.addEventListener('click', () => {
+        console.log(`clicked`, this.flippableClass)
```
robacarp, line 11:
```suggestion
```

app/javascript/application.js
```diff
@@ -1,7 +1,6 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
-import "@hotwired/turbo-rails"
-import "controllers"
+import "@hotwired/turbo-rails";
+import "controllers";
```
robacarp, line 3: Not that I have any _problem_ with these...but my eyebrows are raised.

app/views/conversations/show.html.erb
```diff
@@ -48,52 +48,86 @@
   </div>
 
 
-  <div id="right-column" class="flex flex-1 flex-col h-screen bg-white dark:bg-graphite">
+  <div id="right-column" class="flex flex-1 flex-col h-screen bg-white dark:bg-graphite relative">
+
+    <div id="handle-container" class="absolute left-0 top-1/2 w-8 h-20 -mt-10 z-20 hidden md:flex" data-click-toggle-target="trigger">
+      <div id="handle" class="flex-1 flex cursor-pointer group" data-controller="click-toggle" data-click-toggle-target="trigger" data-click-toggle-flippable-class="!hidden">
```
robacarp, line 54: I'm confused -- this div has both `data-controller="click-toggle"` and `data-click-toggle-target="trigger"`. Is this the trigger or a new instance of the controller? 

Why use `!hidden` as the toggled class instead of `visible`?

app/javascript/controllers/click_toggle_controller.js
```diff
@@ -0,0 +1,24 @@
...
+        this.destinationTargets.forEach(element => {
+          element.classList.toggle(this.flippableClass)
+        })
+      })
+    })
+  }
+
+  disconnect () {
+    this.triggerTargets.forEach(triggerElement => {
+      triggerElement.removeEventListener('click')
```
robacarp, line 21: This doesn't do anything, you'll [need to pass the function which is registered as the event handler](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#matching_event_listeners_for_removal).

app/views/conversations/show.html.erb
robacarp: I think I understand what's happening now, but I'm still not sure about how it works. It seems like stimulus would get confused. How does it handle nested controllers? 

This is a clever reduction of the patterns of needing to toggle both the sidebar presence and the button states, but I think I'm going to have to reimplement it in the application controller or a dedicated controller to get a keyboard shortcut hooked onto it.

krschacht: Ah, shoot, I didn't consider keyboard shortcuts. But it is possible to add multiple stimulus controllers to a single element, so we might just be able to make a really lightweight "keyboard-to-click" stimulus controller and bind it in the exact same two places this is bound.

This particular line you've tagged is a new instance of the controller, and for this new instance I'm also defining this particular div as the trigger but it doesn't have to be here. It's just a coincidence that the trigger happens to be the same container object which I'm scoping the controller. The CSS class attribute (flippable) must be declared on the same object as the controller (I found that in the reference here: https://stimulus.hotwired.dev/reference/css-classes#attributes).

The click-toggle controller needs two objects: a trigger and a target. The trigger is where it binds the click event and the target is where the class will be toggled.

The reason I toggled `!hidden` is because, unfortunately, the tooltip CSS class is a bit heavy handled and it forces the object to be a block object so I have to override that in order to get the thing to disappear, but even "hidden" isn't enough to override that because the Daisy UI library (where I'm getting the tooltip) is included in application.css *after* tailwinds so it's classes take precedence.

I found guidelines for nested stimulus here: https://stimulus.hotwired.dev/reference/controllers#nested-scopes Basically, it "just works".

app/views/conversations/_assistant_header.html.erb
krschacht: I did not understand the tabindex line, that came from the Daisy UI docs for declaring a drop-down menu. Apparently, without this the menu will not be dismissed when you click outside: https://daisyui.com/components/dropdown/#method-2-using-css-focus

The thing I like about the Daisy UI library is that it's completely CSS components. No JS inclusion is needed. It's amazing how much interactivity you can accomplish solely with CSS these days (tooltips & menus are the two items I'm using). But there are still some occasionally small hacky things like this tabindex.

app/javascript/controllers/textarea_autogrow_controller.js
krschacht: Yes, I am still a little confused by the javascript this scoping. But on this particular example it seems to be working without an error. What makes you think it needs the bind(this)?

krschacht: @robacarp Yes, I am still a little confused by the javascript this scoping. But on this particular example it seems to be working without an error. What makes you think it needs the bind(this)?

app/javascript/controllers/nested_clickable_controller.js
krschacht: Yea, they're all working. I know there is a difference between these two ways of declaring a callback. Maybe this is why? 

addEventListener('mouseover', this.removeParentClickable)
addEventListener('mouseover', () => this.removeParentClickable())

app/javascript/application.js
krschacht: Oh, I agree, I don't like semicolons. I removed these at some other point but my have reverted them back in.

app/javascript/controllers/nested_clickable_controller.js
krschacht: I agree, that's a good suggestion. Adding it now.

app/javascript/controllers/click_toggle_controller.js
krschacht: Good catch, fixing!

app/javascript/controllers/nested_clickable_controller.js
krschacht: I did some reading about this and I am now thoroughly confused. I don't understand why it works. But noably, the click_toggle_controller fails to work as a stand-alone function unless I bind the this context.

Just to be thorough, I've convered them all to arrow functions just to ensure `this` is preserved.

app/javascript/controllers/textarea_autogrow_controller.js
krschacht: I re-wrote all these just to be safe. Explained more in a different comment.

app/views/conversations/show.html.erb
robacarp: Wonderful thank you for the links. I'm still learning what Stimulus can do and what it trips on.

app/views/conversations/_assistant_header.html.erb
robacarp: Here's what I found in the mozilla docs when I went looking.

image

PR #62, opened 2024-01-13
Basic Keyboard Shortcut Support

### LINKED ISSUE #25: Basic keyboard support working
- [ ] `Cmd+Enter` to submit a chat
- [ ] Focus the chat input with the existing shortcut + `/`
- [ ] Close the sidebar with the existing shortcut + `Cmd+.`
- [ ] Something to start a new conversation

https://www.stimulus-components.com/docs
https://github.com/stimulus-use/stimulus-use

robacarp:
fixes #25 

Note that this branch is built on top of the conversation scroll logic branch, just for my own sanity and convenience.

- [x] hook in shortcuts for UI widgets which aren't yet present in this branch after rebase

### DIFF
--- a/app/javascript/controllers/click_toggle_controller.js
+++ b/app/javascript/controllers/click_toggle_controller.js
@@ -11,18 +11,17 @@ export default class extends Controller {
 
   connect () {
     this.triggerTargets.forEach(triggerElement => {
-
-      triggerElement.addEventListener('click', () => this.clicked())
+      triggerElement.addEventListener('click', this)
     })
   }
 
   disconnect () {
     this.triggerTargets.forEach(triggerElement => {
-      triggerElement.removeEventListener('click', () => this.clicked())
+      triggerElement.removeEventListener('click', this)
     })
   }
 
-  clicked () {
+  handleEvent () {
     this.destinationTargets.forEach(element => {
       element.classList.toggle(this.flippableClass)
     })
--- a/app/views/conversations/show.html.erb
+++ b/app/views/conversations/show.html.erb
@@ -1,7 +1,14 @@
 <%= turbo_stream_from @conversation %>
-<div class="flex h-screen" data-controller="click-toggle" data-click-toggle-flippable-class="md:flex">
+<div class="flex h-screen"
+     data-controller="click-toggle"
+     data-click-toggle-flippable-class="md:flex"
+     data-action="keydown.meta+shift+s@document->click-toggle#handleEvent:prevent"
+     >
 
-  <div id="left-column" class="flex-col h-screen w-[260px] bg-black overflow-y-auto overflow-x-clip hidden md:flex pl-3 scrollbar-hide" data-click-toggle-target="destination">
+  <div id="left-column"
+       class="flex-col h-screen w-[260px] bg-black overflow-y-auto overflow-x-clip hidden md:flex pl-3 scrollbar-hide"
+       data-click-toggle-target="destination"
+   >
 
     <div id="left-content-container" class="flex flex-col flex-grow overflow-y-auto overflow-x-clip scrollbar-hide">
       <div id="left-content" class="flex-grow overflow-y-auto overflow-x-clip scrollbar-hide">
@@ -124,7 +131,14 @@
               <%= form.text_area :content_text,
                 data: {
                   new_message_target: "input",
-                  controller: "textarea-autogrow"
+                  application_target: "newMessageField",
+                  controller: "textarea-autogrow",
+                  action: %|
+                    new-message#disableSubmitButton
+                    keydown.enter->new-message#submitForm
+                    keydown.shift+esc@document->new-message#focusKeydown
+                    keydown.slash@document->new-message#focusKeydown
+                  |
                 },
                 rows: 1,
                 placeholder: "Message #{@conversation.assistant.name} ...",
--- a/app/javascript/controllers/application.js
+++ b/app/javascript/controllers/application.js
@@ -1,6 +1,14 @@
-import { Application } from "@hotwired/stimulus"
+import { Application, defaultSchema } from "@hotwired/stimulus"
 
-const application = Application.start()
+const applicationSchema = {
+  ...defaultSchema,
+  keyMappings: {
+    ...defaultSchema.keyMappings,
+    "slash": "/",
+  }
+}
+
+const application = Application.start(document.documentElement, applicationSchema)
 
 // Configure Stimulus development experience
 application.debug = false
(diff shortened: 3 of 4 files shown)


app/javascript/controllers/application_controller.js
```diff
@@ -0,0 +1,63 @@
...
+
+  connect() {
+    document.addEventListener("keydown", this.keypress.bind(this), false)
+  }
+
+  disconnect() {
+    document.removeEventListener("keydown", this.keypress.bind(this))
+  }
+
+  keypress(event) {
```
robacarp, line 16: I looked at Stimulus-Use's Hotkey library for this and honestly it felt totally unnecessary. 

This is fundamental javascript and we want it to be as fast as possible -- importing several hundred lines of meta-dispatching based on string compositions representing key chords just to accomplish some basic switching feels unnecessary.

However, if we end up in a place where we allow users to reconfigure shortcuts, I'd vote we move for that package.

app/javascript/controllers/application_controller.js
```diff
@@ -0,0 +1,63 @@
...
+      case "/":
+        if ( ! this.newMessageFieldTarget.matches(":focus")) {
+          this.newMessageFieldTarget.focus()
+          matched = true
+        }
+        break
+
+      // Cmd+Shift+S toggles the sidebar
+      case "s":
+        if (event.metaKey && event.shiftKey) {
```
robacarp, line 38: This is _probably_ the wrong shortcut for windows and linux users, which would prefer the ctrl key here.

app/javascript/controllers/click_toggle_controller.js
```diff
@@ -11,18 +11,17 @@ export default class extends Controller {
...
-
-      triggerElement.addEventListener('click', () => this.clicked())
+      triggerElement.addEventListener('click', this)
     })
   }
 
   disconnect () {
     this.triggerTargets.forEach(triggerElement => {
-      triggerElement.removeEventListener('click', () => this.clicked())
+      triggerElement.removeEventListener('click', this)
```
robacarp, line 20: @krschacht I went on a dive with gpt and mozilla docs this morning to learn about this. 

The reason these were working without a `.bind(this)` is that you wrapped it an arrow function, which implicitly includes the binding from the parent scope. The anonymous function is discouraged because it's not possible to de-register -- restating the anonymous function in the detach handler call doesn't reference the same function, so it's never detached.

Apparently, `addEventListener('whatever', this.responder.bind(this))` has the same problem -- `this.responder.bind(this)` creates a function reference with binding. Running that code again creates a new function reference. It's identical, but not the same, so this doesn't work:

```javascript
this.elementTarget.addEventListener('click', this.someFunction.bind(this))
this.elementTarget.removeEventListener('click', this.someFunction.bind(this)) // not the same function reference as above
```

In order to do it properly:

```javascript
this.boundEventHandler = this.someFunction.bind(this)
this.elementTarget.addEventListener('click', this.boundEventHandler)
this.elementTarget.removeEventListener('click', this.boundEventHandler) // works, because it's the same reference
```

But then I found this:

> The method addEventListener() works by adding a function, or an **object that implements a handleEvent() function**, to the list of event listeners.

[mozilla docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#the_event_listener_callback)

So for controllers like this which only contain a single event handler, we can easily do this:

```javascript
this.elementTarget.addEventListener('click', this)
this.elementTarget.removeEventListener('click', this)

// assuming this function exists:
handleEvent(event) {
   // ...
}
```

krschacht: Oh, very interesting! That's helpful. The way you refactored this with the `handleEvent()` reads nicely. It relies on a little more confidence we're not going to eventually add additional handlers. But stimulus controllers do naturally encourage us to keep them small / single purpose.

I like the syntax of arrow functions a little better than the bind(this) so I'll probably bias towards `this.boundEventHandler = () => this.someFunction()` in those cases.

But the weird thing about this whole situation is that I was not initially using the arrow functions. For every case except click_toggle_controller, I was simply passing a handler to the function and it was working. For example, this intermediate commit was definitely working for me:
[`9728e14` (#61)](https://github.com/hostedgpt/hostedgpt/pull/61/commits/9728e1465f8f7328320d486ca20cb10904b8d0a9)

So I think the stimulus controller class is still doing something magical if you reference `this.element` and `this.parent` such that they work without binding the `this` context. But in any case, I'd rather follow the explicit pattern of ensuring that the context is properly set.

app/views/conversations/show.html.erb
```diff
@@ -1,7 +1,14 @@
 <%= turbo_stream_from @conversation %>
-<div class="flex h-screen" data-controller="click-toggle" data-click-toggle-flippable-class="md:flex">
+<div class="flex h-screen"
+     data-controller="click-toggle"
+     data-click-toggle-flippable-class="md:flex"
+     data-action="keydown.meta+shift+s@document->click-toggle#handleEvent:prevent"
+     >
```
krschacht, line 6: I'll start trying to adopt this line-breaking since these HTML tags are all getting so long, thanks to the combo of stimulus and tailwind

app/views/conversations/show.html.erb
```diff
@@ -1,7 +1,14 @@
 <%= turbo_stream_from @conversation %>
-<div class="flex h-screen" data-controller="click-toggle" data-click-toggle-flippable-class="md:flex">
+<div class="flex h-screen"
+     data-controller="click-toggle"
+     data-click-toggle-flippable-class="md:flex"
+     data-action="keydown.meta+shift+s@document->click-toggle#handleEvent:prevent"
```
krschacht, line 5: I just did a little research to see if there is a more standard one:

* Slack uses `Cmd + .` for the right sidebar and `Cmd Shift D` for the left nav bar
* Notion uses `Cmd + .` for the left nav bar -or- `Cmd \` (they seem to more prominently advertise the `\` even though I've always used period)
* Visual Studio uses `Cmb + B` to open and close the sidebar

...I really wish there was a more standard shortcut like there is for focusing within the search bar.

krschacht (approved):
👍 PR looks good. I'm liking how stimulus is playing out :) Some of the syntax is gnarly like `keydown.slash@document->new-message#focusKeydown` all within a HTML attribute, but it's readable and leads to a nice encapsulation of the javascript so I like the tradeoffs so far.

krschacht (approved):
👍 PR looks good. I'm liking how stimulus is playing out :) Some of the syntax is gnarly like `keydown.slash@document->new-message#focusKeydown` all within a HTML attribute, but it's readable and leads to a nice encapsulation of the javascript so I like the tradeoffs so far.

app/views/conversations/show.html.erb
robacarp: This line ... is like a line of Perl code. It's so obtuse to declare it all this way, but I like that I don't have to write import a bunch of node packages to make it work.

PR #64, opened 2024-01-15
Add the Mac M1 segfault fix into Procfile

krschacht:
This attempts to fix https://github.com/hostedgpt/hostedgpt/issues/42

However, I'm not able to repro this issue any longer in my dev environment so it's hard for me to confirm. But I think it's safe to merge in and since it happened to multiple people I do think it's still an issue.


robacarp:
@krschacht I haven't had this issue since moving SolidQueue out of puma.

PR #65, opened 2024-01-15
Creates RUN_SOLID_QUEUE_IN_PUMA env var

### LINKED ISSUE #63: Make render use puma for worker (but keep worker in it's own thread for dev)

robacarp:
Set RUN_SOLID_QUEUE_IN_PUMA to any value to cause solid queue to run within puma. In render.yaml, the default is to enable the in-puma queue worker.

This is helpful in settings where puma load is low, and running a dedicated worker is unnecessary.

fixes #63


krschacht (approved):
👍

krschacht:
I'm going to go ahead and merge this in so I can try a new deploy to render

PR #66, opened 2024-01-16
run system tests in CI

robacarp:


.github/workflows/rubyonrails.yml
```diff
@@ -55,6 +55,61 @@ jobs:
...
+    env:
+      RAILS_ENV: test
+      DATABASE_URL: "postgres://rails:password@postgres:5432/hostedgpt_test"
+    steps:
+      - run: |
+          set -x
+          apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 4EB27DB2A3B88B8B
+          echo "deb [arch=amd64] https://dl.google.com/linux/chrome/deb/ stable main" | tee /etc/apt/sources.list.d/google.list >/dev/null
+          apt-get update
+          apt-get install -y google-chrome-stable jq curl unzip
```
krschacht, line 80: I'm kind of surprised there isn't a docker image with this already installed, but I don't have any concerns with gettin git working this way too

.github/workflows/rubyonrails.yml
```diff
@@ -55,6 +55,61 @@ jobs:
...
+          gem install bundler
+
+      - name: Install gems
+        run: |
+          bundle config set path vendor/bundle
+          bundle install
+
+      - name: Set up database schema
+        run: bin/rails db:schema:load
+
```
krschacht, line 109: I don't think you need this; I was pretty sure tests did this automatically at the start of each test run, but I could be wrong. There is little harm in having it.

krschacht (approved):
👍 the whole idea of this looks good, some small comments, and I see the newly added system test is failing but I'm guessing that's not a result of this PR but simply that it's now revealing the breakage that we've had

.github/workflows/rubyonrails.yml
robacarp: There's a github action for it, but it's broken. That's where I pulled these lines from.

.github/workflows/rubyonrails.yml
robacarp: I was surprised to see this in here too -- I just copied most of this action from the workflow that was already present.

.github/workflows/rubyonrails.yml
robacarp: I was actually able to resolve this -- the github action for chromedriver wasn't working because we were running our tests in a container, and the action wasn't compatible with that.

PR #74, opened 2024-01-18
Improves the signup and login flows

### LINKED ISSUE #67: Improve styling of the login & signup form (and importantly, we need to remove OpenAI logo)

### LINKED ISSUE #68: Immediately after logging in (or registering, which also logs in) direct to new homepage (conversations/123/messages) and bootstrap initial data

robacarp:
fixes #68 
fixes #67 

I also added a bunch of missing logic to the signup and login flows, including validating that an email is lowercase, unique, and not wrapped with whitespace. If you're anything like me, you'll probably have to `db:reset` after this goes in to wipe out a bunch of duplicated Persron records.

The login and signup pages look essentially identical:

image

There's still some weirdness with the error messaging when the password has errors (at account creation), but I think that can wait for now.

image


app/contexts/registration_context.rb
```diff
@@ -0,0 +1,43 @@
+class RegistrationContext < ApplicationContext
```
robacarp, line 1: I broke this logic out from UserController because there's a lot happening here, and probably more eventually.

The UserController#create action was already getting a little bloated, and I think it's worth breaking out into a dedicated bit of code.

I tried to create a healthy, usable interface, something that is pretty inspired by railsisms (with ActiveModel::Errors), is testable, and doesn't bloat the controller.

Comments welcome!

test/test_helper.rb
```diff
@@ -35,23 +35,9 @@
...
-    self == OpenStruct.new || super
-  end
-end
-
 class ActionDispatch::IntegrationTest
   include Rails.application.routes.url_helpers
 
-  def response
-    JSON.parse(@response.body, object_class: OpenStruct)&.data
-  end
```
robacarp, line 53: I had to remove this because it was attempting to parse every response as JSON. I poked at it a bit to switch on response type, but decided to remove it because we don't have anything json right now. There's some stuff here which is obviously inspired by patterns we don't have, eg `.data` at the end there.

If we later need a json-parsing helper here, I'm all for it.

app/controllers/users_controller.rb
```diff
@@ -2,24 +2,27 @@ class UsersController < ApplicationController
...
-        if @person.save
-          reset_session
-          session[:current_user_id] = @user.id
-          format.html { redirect_to dashboard_path, notice: "Account was successfully created." }
-          format.json { render :show, status: :created, location: @user }
-        else
-          format.html { render :new }
-          format.json { render json: @user.errors, status: :unprocessable_entity }
-        end
+    context = RegistrationContext.new params.permit(:email), params.permit(:password)
```
krschacht, line 10: I don't know about this pattern. It feels like it obfuscates and buries some important business logic into this little pocket on the side, whereas I think it deserves to be more mainline. Rails has a really great built in solution for where to put this kind of logic: in the model concerns & model callbacks. I just found those scale a lot better over time with fewer tradeoffs. Admittedly, I don't think I did the best job of that in the early days of the lava codebase, but the more I leaned into that pattern, it really worked well.

I get that the controller was expanding in scope, and you don't want to bloat it even more with additional extra logic. I definitely would share that goal. But I'm not sure about creating an extra abstraction in order to handle this. I once did a project that involved a lot of service objects of different kinds and it ends up spreading out a lot of important business logic into a bunch of different places and I found it makes it harder to get a unified view of what's happening.

I'd push this into the model. To play this out:

This line could simply be changed to:
```
@user = User.new(password: user_params[:password], password_confirmation: user_params[:password])
```

Then within the User model create a concern such as `include Registerable` where all this RegistrationContext lives. It can do all the things that Context is doing.

A potential worry of adding all this logic into the model is that the model will collect quite a lot of logic over time, but the combination of Concerns and callbacks make that quite easy to manage LOTS of business logic all wrapped behind the ActiveRecord API for models. I think it's actually not hard to manage, provided we keep the file organized, and we keep bundling things into concerns.

I've been going through 37signals basecamp source code snippets and many of their models are huge, but they're surprisingly easy to wrangle still. Here's one example: https://github.com/krschacht/37signals-rails-code/blob/main/basecamp/app/models/recording.rb

I think the two main advantages of this are:

* You open a model like the one I linked to above and the whole thing reads almost like a table of contents! You start with a list of concerns that name large chunks of logic, followed by core associations and hooks.

* There will inevitably be other places that we create User objects and the vast majority of the time we're also going to want to fully initialize those with a Person record and other stuff. Having it in the model encourages that golden path, and the one-off times we don't want to we can use the various blocks that activerecord exposes to suppress callbacks. Whereas when we sprinkle this business logic in Contexts, it's pushing us to create a new Context for a slightly different situation and duplicate a lot of the same logic.

* There are often bits of logic that we are adding to one model which initially seem like they apply onto to that model: so we start off with a concern that is used solely by that model. But over time we see ourselves doing something similar in another model. After the 2nd or 3rd time we generalize a single concern and mix it in for all the models that do it. So starting off with concerns really sets you up for lifting that logic up and sharing it across models. And the same applies to hooks as well. You start off with one-off hooks, realize you're doing this in a few places and generalize and elevate that hook up.

What do you think about pushing the models further before creating new classes?

krschacht (changes requested):
@robacarp I have some hesitations about the new abstraction, left some comments. Let me know what you think.

krschacht (changes requested):
@robacarp I have some hesitations about the new abstraction, left some comments. Let me know what you think.

app/models/user.rb
```diff
@@ -3,8 +3,7 @@ class User < ApplicationRecord
 
   has_secure_password
 
-  validates :password, presence: true, on: :create
-  validates :password_confirmation, presence: true, on: :create
+  validates :password, presence: true, on: :create, length: { minimum: 6 }
```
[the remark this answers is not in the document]
robacarp: Oh good call, yep.

app/controllers/users_controller.rb
robacarp: Oooh interesting. I didn't consider a model concern, I've always exclusively used those for features which are shared between multiple models (Taggable, Commentable, etc) I'll rearrange this code and see what I come up with. Good suggestion!

I proposed this because it's a pattern I've seen which did better than the kind of lazy idea of just shoving everything into one file -- but there's a lot in there which isn't great.

krschacht: Nice. This is probably more info than you're interested in so don't feel obligated to consume this :) but a year ago I stumbled on a bunch of talks that DHH gave really unpacking how he thinks about organizing rails code. This is what started me studying the Basecamp code more closely. It changed a lot of my thinking. I think these two are the ones related to what we're going back and forth on, and were probably the most interesting ones, all around:

https://www.youtube.com/watch?v=M3JPTOTqsnE
https://www.youtube.com/watch?v=Yd3j3ZEx6DU

robacarp: Ah yes, I'm remembering a bit more. One of the things about the registration action is that it's not constrained to a single object. Because we have Person(:email) and User(:password), the registration form gets weird.
Rendering errors really gets weird, which is why I spent so much of this code wrangling errors into a single Errors list.

It also doesn't really play well with StrongParameters, or at least I don't know how to make them play well together.

I'm going to find some more docs on delegated types and see if there's an easy way through this it.

robacarp: @krschacht this worked out brilliantly. I need to move the tests around still, but I'm much happier with the code organization using nested attributes. Thanks for your nudge.

app/models/user.rb
```diff
@@ -1,10 +1,9 @@
 class User < ApplicationRecord
-  include Personable
+  include Personable, Registerable
 
   has_secure_password
 
-  validates :password, presence: true, on: :create
-  validates :password_confirmation, presence: true, on: :create
+  validates :password, length: { minimum: 6 }, allow_nil: true
```
krschacht, line 6: Do you want to allow nil password? It feels like we should require it to always be present on User objects. In my mind, it's the definition of a User: it's a person with credentials.

app/models/concerns/user/registerable.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  included do
+    after_create :create_initial_assistant
+  end
+
+  private
+
+  def create_initial_assistant
+    assistant = assistants.create! name: "HostedGPT"
+    conversations.create! title: "HostedGPT", assistant: assistant
```
krschacht, line 12: What about making the conversation.create! a side effect (hook) inside Assistant rather than here? I would assume we will pretty much always want an assistant to have an initial conversation. It continues the spirit of pushing considerations like this deeper into the model where they'll get the maximum reuse.

Although, another possibility—but I think this is out of scope for this PR—is that we make the PR work elegantly without needing a conversation to exist. One side effect of pre-creating this conversation is that it's going to appear in the side-bar and have this title, which may not be a title that actually makes sense.

app/controllers/users_controller.rb
```diff
@@ -2,24 +2,28 @@ class UsersController < ApplicationController
   include Accessible
 
   def new
+    @person = Person.new
+    @person.personable = User.new
```
krschacht, line 6: One small simplification is that you can just use one of these lines and then push the initialization of the association down into the model. If we definitely want `@person` as the end result, then within Person.rb

```
after_initialize :set_default_personable

private

def set_default_personable
  self.personable ||= User.new
end
```

app/controllers/users_controller.rb
```diff
@@ -2,24 +2,28 @@ class UsersController < ApplicationController
...
-          format.html { redirect_to dashboard_path, notice: "Account was successfully created." }
-          format.json { render :show, status: :created, location: @user }
-        else
-          format.html { render :new }
-          format.json { render json: @user.errors, status: :unprocessable_entity }
-        end
+    @person = Person.new
+    @person.personable_type = "User"
+    @person.update person_params
+
```
krschacht, line 13: But it actually feels backwards that the `def new` and the `def create` are treating @person as the object we're dealing with, when this is the users_controller. Can't we do the reverse and always initialize and create User objects? I'm pretty sure it would work to:

1. Change `def new` to do `@user = User.new` by adding this to User.rb.

```
after_initialize :set_default_person

private

def set_default_person
  self.person ||= Person.new
end
```

I think this is actually better than what I proposed in the comment above, because a User *always* has a Person but a Person does not always have a User.

2. Change this create to simply be `@user = User.new ...` and within User.rb you can then add:

```
  has_one :person, as: :personable, touch: true, autosave: true
  accepts_nested_attributes_for :person
```

app/contexts/application_context.rb
```diff
@@ -0,0 +1,22 @@
+class ApplicationContext
```
krschacht, line 1: you can drop this file

app/contexts/registration_context.rb
```diff
@@ -0,0 +1,43 @@
...
+  attr_reader :user, :person, :first_conversation
```
krschacht, line 2: and this file too

krschacht (changes requested):
@robacarp One more potential simplification that could be significant is to do everything driven by @user objects rather than @person objects. Added some notes to explain the proposal

app/models/user.rb
robacarp: password is a virtual attribute and is nil when the user isn't updating their password -- the digest is saved in it's own column

app/controllers/users_controller.rb
robacarp: I learned some things today, and re-learned some other things, haha. Good call -- I didn't think you could do nested attributes across `has_` associations, but they've done a good job of making the api more robust than I thought.

app/models/concerns/user/registerable.rb
robacarp: I considered putting it inside of the assistant model, but then it would be copying over a user between the associated objects which feels a bit muddy. Maybe it's correct? I think the right approach is to make the conversation created after the first message is sent.

krschacht: The user is an attribute of assistants, so it wouldn't actually need to reference `assistants.user` as an object, it's technically referencing `assistants.user_id`. And I think it feels proper for the initialization of a conversation object by referencing multiple attributes of assistants.

BUT, I agree initializing the conversation after the first message is sent feels like the right ultimate answer. Both because conversations have titles, which can't really be set properly until the conversation starts, and because conversations have a created_at, which conceptually should refer to the moment the conversation started.

I'd vote to just leave it as you've done it for now and we refactor this as part of a subsequent PR. Better to get this PR landed.

app/controllers/users_controller.rb
krschacht: I wasn't sure either! ChatGPT for the win. This is the kind of "architecture conversation" I have with ChatGPT so often these days. It turns out that I have a friend named Samantha who knows the rails API *really* well. :) This might be interesting for you to see or it might be old hat, by now:

image

PR #77, opened 2024-01-21
Further home refinements

krschacht:
- [x]  Add a couple more keyboard shortcuts (Esc to unfocus and Cmd+. to toggle sidebar)
- [x]  A close look at Safari and Firefox to make sure there are no small pixel differences
- [x]  Make the background of the tooltips white and tweak the glow effect in dark mode, it's garish
- [x]  Make menu items be white
- [x]  Make the sidebar keyboard shortcut bind to two separate elements
- [x] Ensure the textbox size is recalculated when sidebar is opened/closed
- [x] Ensure all event listeners are added/removed properly by wrapping those functions

I did not finish:

- [ ]  Try this web UI on tablet & phone and lift some design details from ChatGPT iPad app
- [ ]  Add transition/animations such as the sidebar opening & closing


app/javascript/controllers/new_message_controller.js
```diff
@@ -47,4 +47,11 @@ export default class extends Controller {
     this.focusInput()
     event.preventDefault()
   }
+
+  unfocusKeydown(event) {
+    if (event.key === "Escape" && event.target.tagName === "TEXTAREA") {
```
robacarp, line 52: Since you're attaching this handler with `keydown.esc->new-message#unfocusKeydown` you shouldn't need to check that it's an Escape key.

app/views/conversations/show.html.erb
```diff
@@ -138,18 +144,20 @@
                     keydown.enter->new-message#submitForm
                     keydown.shift+esc@document->new-message#focusKeydown
                     keydown.slash@document->new-message#focusKeydown
+                    keydown.esc->new-message#unfocusKeydown
```
robacarp, line 147: If you put this on your textarea instead of up here, do you still need to check the event.target and conditionally prevent default?

robacarp (commented):
Just one little refinement on your focus handler.

app/javascript/controllers/new_message_controller.js
krschacht: Ohhh, I didn't understand why you did that with focusKeydown so I was just blindly following the pattern, but now that I read the comment you made it clear: it's because "/" is a legit character that can be typed int he input so it needs to be checked.

app/views/conversations/show.html.erb
krschacht: Good call, I think you're right. I'll move it.

app/views/layouts/application.html.erb
```diff
@@ -5,10 +5,10 @@
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
+    <%= stylesheet_link_tag "https://cdn.jsdelivr.net/npm/daisyui@4.6.0/dist/full.min.css" %>
```
krschacht, line 8: Within application.css, I want to override some of the classes that we get from this library. In order to do that, application.css needs to get included *after* this library is imported.

robacarp: I wondered if that was the case and started to type out a comment earlier today about it. That might be worth having in a source comment.

app/javascript/application.js
```diff
@@ -2,5 +2,6 @@
 import "@hotwired/turbo-rails"
 import "controllers"
 import Rails from "@rails/ujs"
+import './utils/debounce'
```
krschacht, line 5: @robacarp I'm not sure if this is the best pattern for including global utilities or not

app/javascript/utils/debounce.js
```diff
@@ -0,0 +1,18 @@
...
+        timeout = null
+        if (!immediate) func.apply(context, args)
+      }
+      var callNow = immediate && !timeout
+      clearTimeout(timeout)
+      timeout = setTimeout(later, wait)
+      if (callNow) func.apply(context, args)
+    }
+  }
+}
```
krschacht, line 18: @robacarp If I don't hang the function on window (like lodash does) then it's not visible inside the stimulus controller. Probably it's because I need to do an import? I figured: it'd be nice not to have to an import a collection of utility functions we build up over time so maybe I put them in the global namespace by copying this hack? Thoughts welcome.

app/views/conversations/show.html.erb
```diff
@@ -59,7 +60,14 @@
...
     <div id="handle-container" class="absolute left-0 top-1/2 w-8 h-20 -mt-10 z-20 hidden md:flex" data-click-toggle-target="trigger">
-      <div id="handle" class="flex-1 flex cursor-pointer group" data-controller="click-toggle" data-click-toggle-target="trigger" data-click-toggle-flippable-class="!hidden">
+      <div  id="handle"
+            class="flex-1 flex cursor-pointer group"
+            data-controller="click-toggle"
+            data-click-toggle-target="trigger"
+            data-click-toggle-flippable-class="!hidden"
+            data-action="keydown.meta+shift+s@document->click-toggle#handleEvent:prevent
+                         keydown.meta+period@document->click-toggle#handleEvent:prevent"
+            >
```
krschacht, line 70: @robacarp I realized that when the keydown event was only bound to the outer click-toggle then the sidebar opens & closes with the keyboard shortcut but the visual of the handle doesn't change properly. So I added a data-action here as well.

It kind of worked, but two issues:

1) I don't know why `:prevent` is on here but I saw you did it on the outer one and I figured you may have a good reason. I tried it without :prevent and it seemed to work fine so I'm inclined to remove it, but do we need it?

2) When I added it here then the binding events work properly when I click to focus on the document, however when the focus is inside the textarea (which is the default on load) then one of these two bindings fires twice! It's so odd. I load the page and press `Cmd+.` and I see 3 fires: 2x for one of them and 1x for another. I spent awhile trying to track it down and could not. It may have something to do with my nested controllers that I don't quite understand. In the end I solved it by simply adding a debouncer, which feels a little dirty but "good enough" for now.

app/javascript/controllers/textarea_autogrow_controller.js
```diff
@@ -2,41 +2,28 @@ import { Controller } from '@hotwired/stimulus'
...
-function debounce (callback, delay) {
-  return (...args) => {
-    const context = this
-    if (typeof timeout !== 'undefined')
-      clearTimeout(timeout)
-    else
-      var timeout = undefined
-
-    timeout = setTimeout(() => callback.apply(context, args), delay)
-  }
```
krschacht, line 41: Since a second time has arisen for a debouncer I factored it out into the global utility.

app/javascript/application.js
robacarp: The "Rails 7 Way" as far as I can tell is this:

- Use ES7 module syntax and import it when you need it and where you need it; the browser will take care of loading it only once when the javascript files all come down from the server.
- Register the utility in the importmap so that Rails renders a module preload script tag in the layout, so you don't have cascade module loading (A loads B loads C loads D).

app/javascript/controllers/textarea_autogrow_controller.js
```diff
@@ -2,41 +2,28 @@ import { Controller } from '@hotwired/stimulus'
...
     this.element.addEventListener('input', () => this.autogrow())
-    window.addEventListener('resize', () => this.onResize())
+    window.addEventListener('resize', () => this.debouncedAutogrow())
   }
 
-  disconnect () {
+  disconnect() {
     this.element.removeEventListener('input', () => this.autogrow())
-    window.removeEventListener('resize', () => this.onResize())
+    window.removeEventListener('resize', () => this.debouncedAutogrow())
```
robacarp, line 20: This disconnect won't work because the anonymous function is different from the one on line 15. From the way you have it implemented, I believe you can just do this:

```suggestion
    window.removeEventListener('resize', this.debouncedAutogrow)
```

(and match above)

app/javascript/controllers/textarea_autogrow_controller.js
```diff
@@ -2,41 +2,28 @@ import { Controller } from '@hotwired/stimulus'
...
-  connect () {
+  connect() {
     this.element.style.overflow = 'hidden'
-    this.onResize = this.resizeDebounceDelayValue > 0 ? debounce(this.autogrow, this.resizeDebounceDelayValue) : this.autogrow
-
     this.autogrow()
 
     this.element.addEventListener('input', () => this.autogrow())
-    window.addEventListener('resize', () => this.onResize())
+    window.addEventListener('resize', () => this.debouncedAutogrow())
```
robacarp, line 15:
```suggestion
    window.addEventListener('resize', his.debouncedAutogrow)
```

app/javascript/utils/debounce.js
robacarp, line 18: Here's what I would do:

```suggestion
export default function(func, wait, immediate) {
    var timeout

    return function() {
      var context = this, args = arguments

      var later = function() {
        timeout = null
        if (!immediate) func.apply(context, args)
      }
      var callNow = immediate && !timeout
      clearTimeout(timeout)
      timeout = setTimeout(later, wait)
      if (callNow) func.apply(context, args)
    }
  }
}
```

And then when you go to use it:

```javascript
# javascript/controllers/some_example.js
import debounce from "../utils/debounce.js"
import { Controller } from '@hotwired/stimulus'

export default Class extends Controller {
 // blah
}
```

-------

If you're set on it being a "collection" of utilities, I'd probably do this instead. The only material difference is exporting more than one thing from the module.

```javascript
export const debounce = function (...) {...}
```
and
```javascript
import { debounce } from "../utils"
```

robacarp: The es6/7 module js and import statements make it so this isn't necessary, but more importantly it provides a benefit too.

If for some reason you need to use two different debounce functions, you can do so and name them however you want in the module that needs them and not affect the rest of the application. When you declare it globally like this, if someone later loads lodash itself into the global namespace, it'll overwrite this and you end up with a strange bug that's difficult to find and hard to fix.

If you use it in your module you can do this and still use our code and some other code:

```javascript
import { debouce as ourDebouncer } from "../utils/debounce"
import { debounce as someOtherDebouncer } from "google_secret_sauce_debouncer"
```

robacarp (commented):
I'll have to come back later for my thoughts about the keyboard shortcut, but here's my comments about the debouncer.

app/views/conversations/show.html.erb
robacarp: I know stimulus is all about generic controllers, but it sounds to me like it would be simpler to just have a sidebar controller and give it a few dedicated targets and actions. It feels like forcing the pattern to do this much work. 

```javascript
import { Controller } from "stimulus"

  static targets = [ "sidebar", "handle" ]

  toggle() {
    this.sidebarTarget.classList.toggle("md:flex")
    this.handleTargets.forEach( e => e.classList.toggle("hidden") )
  }
}
```

PR #80, opened 2024-01-24
Docker container

robacarp:
This started out as an attempt to make deploying to Fly.io easy, but I don't think that's going to be the case:

- Fly.io [doesn't support or have plans to support](https://community.fly.io/t/deploy-to-fly-button-config-like-heroku/4811/8?u=stale-snow-white-sco) a "Deploy to Fly.io" button.
- Deploying to fly.io has side effects on the local repository -- it adds a gem, a config file, and it can modify the dockerfile
- There is no easy way to deploy to fly.io with a cheap or free database, it requires manually hitting several fly.io commands or API endpoints.

I think building something like this is possible, but it'll be a contraption of Lambda functions hosted somewhere. It's not a bad idea down the road, but I don't think it's worth the investment right now.

So here's what's left:

- A trimmed down dockerfile, without some of the extra fluff that comes in a stock Rails7 dockerfile
- A template fly.toml file, which seems like a good practice to support fly.io deployments even though actually launching the app will mutate the toml.
- A CI action to build the docker container, to make sure we don't break it accidentally.


.ruby-version
robacarp, line 1: I deleted this file because it's redundant to .tool-versions, and I didn't want to have to deal with keeping the ruby version up to date in a bunch of places.

Gemfile
```diff
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-ruby "3.2.2"
+ruby File.readlines(File.join(__dir__ , '.tool-versions')).select{|l| l =~ /ruby/}.first.strip.split.last
```
robacarp, line 5: To make it so there's one source of truth for the ruby version.

krschacht: Clever!

.tool-versions
```diff
@@ -1 +1 @@
-ruby 3.2.2
+ruby 3.2.3
```
krschacht, line 1: It might not matter, but I had started off using the latest stable ruby (3.3.0) and when I did a deploy to Render I ran into an issue. I can't recall what it was, but something blew up when it was trying to auto-install that version of ruby. I found some documentation ([I think this](https://docs.render.com/ruby-version) that explained their default ruby pre-installed is 3.2.2 so I switched us to this version in order to avoid the Render install issue.

If you need to bump this minor version then we should just be sure that Render still works (and you may need to merge in to determine that)

krschacht: That doc also makes reference to a ruby-version so does Render need us to have that? I'd expect it can read from .tool-versions but flagging to keep an eye out for any indication that Render freaks out w/o that file

krschacht (approved):
👍 One Q about ruby version, but looks good, assuming Render is happy. Do you want to write up the steps to deploy to Fly now that we have a toml file, just adding to the README while it's fresh?

Gemfile
robacarp: Apparently in an upcoming Bundler version we can just say `ruby file: '.tool-versions'` and it knows what to do...but not just yet!

.tool-versions
robacarp: Oh good call, I'll look into that. If nothing else maybe we can deploy the container to Render instead of relying on their notion of buildpacks...that way there's less diversity to account for in the hosting systems.

PR #81, opened 2024-01-24
Formats toast messages so they don't break the page layouts

### LINKED ISSUE #75: Toast / Flash messages styling updates
Currently whenever we show a toast or flash message, it breaks the whole page layout. They should float and be dismissable somehow.

image

robacarp:
fixes #75 

image

image


krschacht (approved):
👍

PR #83, opened 2024-01-25
Conversations Layout User menu

### LINKED ISSUE #76: Port the user menu over from chats/notes area to conversations
Currently when you land in the app at conversations/show, the user menu is a placeholder. There's no way to click to logout.

robacarp:
fixes #76 

I simplified the tailwind styles a lot, and I think there were some half baked solutions to the modal problems in there too. I removed them in favor of just using the click toggle controller.

I duplicate a bunch of things because I tried not to break the Chats view, but that means that the partials for rendering the user menu and modal for Chats can be deleted when the chats features are finished getting ported.

image

image


app/views/shared/_user_menu.html.erb
```diff
@@ -0,0 +1,64 @@
+<div
+    class="w-full border-t border-gray-200/50 pt-6 pb-2 pr-2 bg-black flex flex-col gap-2"
+    data-controller="click-toggle"
+    data-click-toggle-flippable-class="hidden"
+    >
+
+    <a href="#"
+       data-action="click->click-toggle#toggleClasses:prevent"
```
robacarp, line 8: @krschacht notably, I didn't bind this as a target for the controller but instead attached the event handler here instead. That didn't occur to me before, but it's probably unnecessary to have the manual attach/detach event handler logic in clickToggle usage.

krschacht: Interesting, I never fully acknowledged those were two different ways to do things. It makes me wonder if the other places we have data-action could instead be converted to data-target? :) I know that's the opposite of what you did here, it seems like you're biasing towards just being explicit with data-action even though it has this perl expressiveness. :)

I'm not opposed to it though! It is kind of nice to have to write less javascript. I'd say let's keep what you did here, knowing that we now have two different approaches, and live with this a bit longer and then soon pick one approach. I'm not sure which is better.

One small suggestion: maybe remove :prevent here and put a preventDefault inside the toggleClasses function. That at least makes this data-action line slightly more readable.

app/views/shared/_user_menu.html.erb
```diff
@@ -0,0 +1,64 @@
...
+       class="flex justfiy-left gap-2 rounded-lg text-white text-sm hover:bg-[#28282b] p-2"
+    >
+      <%= icon "cog" , size: 20 %>
+      Settings
+    </a>
+
+    <a href="/logout" class="flex justify-left gap-2 hover:bg-[#28282b] rounded-lg text-white text-sm p-2">
+    <%= heroicon "arrow-right-start-on-rectangle" , options: { class: 'h-5 w-5 text-gray-300' } %>
```
krschacht, line 16: I think we should always use the wrapper method, <%= icon %>, rather than heroicon. It'll make future refactors easier. And it supports "class:" so you don't have to prefix with options:

app/views/shared/_user_menu.html.erb
```diff
@@ -0,0 +1,64 @@
...
+    <p>Logout</p>
+  </a>
+
+  <div data-click-toggle-target="destination"
+       class="fixed inset-0 z-50 hidden"
+       >
+    <div
+      class="absolute inset-0 bg-[rgba(0,0,0,0.5)] flex justify-center items-center"
+      data-action="click->click-toggle#toggleClasses:self"
+    >
```
krschacht, line 26: stylistically I'm not sure I love the floating `>`, does it make it too hard to read the indentation of you include it at the end of the previous line? That's probably why you are floating it.

If we do float it, you probably intended to align with opening `<div`. The one right above this is aligned with `class`

app/views/shared/_user_menu.html.erb
```diff
@@ -0,0 +1,64 @@
...
+        <h2 class="text-2xl font-bold">Settings</h2>
+        <button class="text-gray-300 hover:text-gray-400 text-3xl" data-action="click->click-toggle#toggleClasses">
+          &times;
+        </button>
+      </div>
+
+        <%= form_with(
+          model: Current.user,
+          method: :put,
+          url: user_path(current_user),
```
krschacht, line 38: `user_path(Current.user)`? I didn't know Ali had added `def current_user`, but I think we should remove that and lean into Current.user everywhere.

krschacht (approved):
👍 couple small style suggestions

app/views/shared/_user_menu.html.erb
robacarp: Interesting point -- since this is a link, the implicit action is a click, so I can remove `click->` from this binding. Putting preventDefault in the handler makes sense only if we want to always prevent default...setting the syntax complaints I have aside, it's nice to be able to make that choice on a per-binding basis. In this case I think it'll always be the case though.

I started moving in this direction while researching the `removeEventListener` and keyboard events. It felt too hard the way we've been doing it. Keyboard listeners get messy quick -- do I want Shift and Meta or hang the listener on the Body for a global handler? Now my controller has to be attached to the body?

But reading through the examples in the manual again -- they are incredibly interactive but never explicitly bind to targets in javascript, but rather bind actions in the html. So I think the Stimulus paradigm is to push as much as possible into the html side and the controllers end up being tiny, even when they're generic. 

Theres [this section in the reference on naming conventions](https://stimulus.hotwired.dev/reference/actions#naming-conventions) for actions and they explicitly recommend against handlers like "click" because it doesn't "help you reason about the behavior of a block of HTML without having to look at the controller source". Which is an interesting concept and I think I get where they're going.

It's clearly not my instinct, because I've been doing it the other way around...but I'm starting to suspect that the "stimulus way" is in this direction.

app/views/shared/_user_menu.html.erb
robacarp: I don't like it either. I don't like breaking up the tags like this but I don't see a way around it. I think moving it up to the end of the data-action is fine. I put it here because it gets weird when you look at context with something nested inside the div, and then what level does that tag indent to? 

From my angle it does look to be aligned with the div though?

image

app/views/shared/_user_menu.html.erb
robacarp: Agreed, I probably just missed this one when I moved this code over.

app/views/shared/_user_menu.html.erb
robacarp: Agreed -- I didn't touch this code but I'll give it a once-over and bring it up to spec.

app/views/shared/_user_menu.html.erb
```diff
@@ -0,0 +1,64 @@
...
+    </a>
+
+    <a href="/logout" class="flex justify-left gap-2 hover:bg-[#28282b] rounded-lg text-white text-sm p-2">
+    <%= heroicon "arrow-right-start-on-rectangle" , options: { class: 'h-5 w-5 text-gray-300' } %>
+    <p>Logout</p>
+  </a>
+
+  <div data-click-toggle-target="destination"
+       class="fixed inset-0 z-50 hidden"
+       >
```
krschacht, line 22: sorry it was this one that was misaligned with the div

PR #84, opened 2024-01-26
Restructure routes for cleaner new conversations

krschacht:
For #87 

Move messages#new and messages#create to be nested under /assistants rather than conversations. This is so that we can present users with a new message form without having to pre-create the conversation.

This also does work deep in the model so that when a new message is being created, if a conversation_id wasn't provided then it pre-creates the conversation. In order for this to be clean, this PR also **adds assistant_id as an attribute of the Message model**.

I'm preparing for a user experience where you can start a new conversation at this route:
http://localhost:3000/assistants/630504688/messages/new

After you submit the first chat message you'll be redirected to this route. A Message and Conversation object were created together:
http://localhost:3000/conversations/115582124

Rails Turbo will make this nice and fast so the user won't even notice. This will support cleaner models and controllers.

(I've already started the other front-end refactor of views to make this all work but I'm breaking the route refactor into it's own PR for easier review.)

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -1,7 +1,7 @@
 class MessagesController < ApplicationController
   skip_before_action :authenticate_user!
-  before_action :set_assistant, only: [:new, :create]
   before_action :set_conversation, only: [:index]
+  before_action :set_assistant, only: [:index, :new, :create]
   before_action :set_message, only: [:show, :edit, :update, :destroy]
 
   def index
@@ -45,12 +45,13 @@ class MessagesController < ApplicationController
 
   private
 
-  def set_assistant
-    @assistant = Current.user.assistants.find(params[:assistant_id])
-  end
-
   def set_conversation
     @conversation = Current.user.conversations.find(params[:conversation_id])
+  end
+
+  def set_assistant
+    @assistant = Current.user.assistants.find_by(id: params[:assistant_id])
+    @assistant ||= @conversation.assistant
   end
 
   def set_message
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -1,6 +1,6 @@
 class MessagesController < ApplicationController
   skip_before_action :authenticate_user!
-  before_action :set_current_assistant, only: [:new, :create]
+  before_action :set_assistant, only: [:new, :create]
   before_action :set_conversation, only: [:index]
   before_action :set_message, only: [:show, :edit, :update, :destroy]
 
@@ -12,14 +12,14 @@ class MessagesController < ApplicationController
   end
 
   def new
-    @message = Current.assistant.messages.new
+    @message = @assistant.messages.new
   end
 
   def edit
   end
 
   def create
-    @message = Message.new(message_params)
+    @message = @assistant.messages.new(message_params)
 
     if @message.save
       redirect_to @message, notice: "Message was successfully created."
@@ -45,8 +45,8 @@ class MessagesController < ApplicationController
 
   private
 
-  def set_current_assistant
-    Current.assistant = Current.user.assistants.find(params[:assistant_id])
+  def set_assistant
+    @assistant = Current.user.assistants.find(params[:assistant_id])
   end
 
   def set_conversation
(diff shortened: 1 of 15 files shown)


app/controllers/messages_controller.rb
```diff
@@ -1,6 +1,7 @@
 class MessagesController < ApplicationController
   skip_before_action :authenticate_user!
-  before_action :set_conversation, only: [:index, :new, :create]
+  before_action :set_current_assistant, only: [:new, :create]
```
krschacht, line 3: I made an interesting design decision that I'm now questioning: I save the assistant onto `Current.assistant`. The reason is because assistant_id exists in the URL so it gets passed in as params. But when I get to `def create`, the new message takes `message_params` and assistant_id is not an attribute of message. However, creating a message has the side effect of creating a new conversation (this is within model.rb) and the creation of this new conversation needs assistant_id.

So how to get assistant_id from the controller down into message? I did it through Current.assistant.

The code ends up reasonably clean, but it feels a little odd to put something into Current which will not be available upon *every* request. Current.user and Current.ip_address and Current.user_agent, etc, those feel like "universals." But I've already found myself trying to remember: "for this particular request, does Current.assistant exist or not?" so that makes it feel wrong.

As I'm reflecting on it more, the root of the messiness was that messages to not relate *directly* to assistants — they relate through conversations. But I think it might be a better tradeoff to simply denormalize the tables further, add assistant_id as an attribute of the message model, even though it isn't needed.

That way, new messages are created through `@assistant.messages.create!` and then the side-effect which creates conversations simply plucks it off of `message.assistant_id`

Another small benefit of this is that when you see the route as `/assistants/12345/messages` it seems to imply that messages directly relate to assistants so it's further surprising when you discover they don't have a direct relationship. I'll reflect on all this a bit further before merging this PR in...

robacarp: There might be benefits to attaching messages to assistants directly and detaching conversations. For example, I wonder if there's a situation where I might want to have a conversation with multiple assistants? What if I want to switch between gpt3.5 and 4 mid conversation, or even openai and azure llm models? That's something we can't render right now, as conversations are coupled to assistants. 

Current.assistant feels like it's going to open up a series of counterintuitive situations. I don't think the Current class should even be available outside of controllers. Controller actions reasonably operate within the scope of a request at all times, but Models aren't so clean. We routinely access and trigger model code from background workers, rails console, etc. I can't think of a time when I ran a controller action from within the console or a worker but models are created and destroyed all the time outside of request contexts. What does it mean to bind Current.user or Current.assistant in a background job?

app/views/messages/_new_form.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_with(model: [message.conversation, message], class: "contents") do |form| %>
+<%= form_with(model: [Current.assistant, message], class: "contents") do |form| %>
```
robacarp, line 1: Maybe this `model: [...]` syntax is evidence that we're pushing hard on something because we're off the standard route? Shouldn't rails know how to render this route without this?

app/models/message.rb
```diff
@@ -1,17 +1,33 @@
 class Message < ApplicationRecord
+  belongs_to :assistant
   belongs_to :conversation
-  delegate :assistant, to: :conversation
```
krschacht, line 3: Message is now directly associated to Assistant

app/models/message.rb
```diff
@@ -1,17 +1,33 @@
...
-    broadcast_append_to conversation, partial: "messages/message", locals: {scroll_into_view: true, conversation: conversation}
-  }
+  after_create_commit :broadcast_message
+
+
+  private
+
+  def create_conversation
+    self.conversation = Conversation.create!(user: Current.user, assistant: assistant)
+  end
```
krschacht, line 24: Note: Current is still used even when creating Messages. I think this a good example of the intended use of Current. It might seem a bit odd to have Current being access from a model, but sometimes the Model creation does indeed depend on context.

I don't anticipate we will ever create Messages outside of a controller when a Conversation will need to be initiated like this then Current.user can be set (for example within a worker or a test). In fact, I do it within the message_test file.

robacarp: Won't we be creating messages from the worker when receiving from the backend api?

app/views/messages/_new_form.html.erb
krschacht: I think that's okay, I consider that pretty standard rails. It's just the syntax when you have nested routes.

app/models/message.rb
krschacht: Those messages will specify the conversation so we'll avoid the need for Current with them.

robacarp (approved):

PR #86, opened 2024-01-29
implements a sidebar powered by the database

### LINKED ISSUE #85: Update conversations sidebar to be powered by real data

robacarp:
fixes #85 

Since I know #84 is in play, I didn't bother getting the pencil icons to actually link to the new message for assistant path. 

I simplified some of the sidebar css -- primarily there were several scrollable containers on top of each other, which I was able to simplify. As far as I can tell it looks and works the same.


test/models/conversation_test.rb
```diff
@@ -47,4 +47,34 @@ class ConversationTest < ActiveSupport::TestCase
       conversation.reload
     end
   end
+
+  test "#grouped_by_increasing_time_interval_for_user" do
+    Timecop.freeze do
```
robacarp, line 52: There's a really tiny probability that this test happing when the clock rolls over a second, which causes a roll over of a week, month, or year.

Doing that would cause this code to break, which is perhaps a remote enough possibility that Timecop isn't warranted. On the other hand, I've used timecop for a _long time_ and it has show itself to be stable and well maintained for many years.

Open to suggestions!

krschacht: Agreed, timecop is very nice and flappy tests suck. :)

app/models/conversation.rb
```diff
@@ -5,4 +5,44 @@ class Conversation < ApplicationRecord
...
+        ..range_start
+      else
+        range_end..range_start
+      end
+
+      sidebar_conversations.where(updated_at: range)
+    end
+
+    Hash[keys.zip(values)].delete_if { |_, v| v.empty? }
+  end
```
krschacht, line 47: that’s a nice bit of ruby there!

app/controllers/conversations_controller.rb
```diff
@@ -42,6 +45,14 @@ def destroy
 
   private
 
+  def set_sidebar_conversations
+    @sidebar_conversations = Conversation.grouped_by_increasing_time_interval_for_user(Current.user)
+  end
+
+  def set_sidebar_assistants
+    @sidebar_assistants = Current.user.assistants.to_a - [@conversation&.assistant]
```
krschacht, line 53: are you removing the currently active assistant? I think we should keep that one visible in the list and make that the selected one (similar to how you did for conversation). That’s how ChatGPT does it.

Notably, when conversations are selected they are visibly highlighted. When the currently active assistant is selected it’s much more subtle: they just keep the new-pencil icon ever present rather than only appearing on hover.

krschacht (approved):
👍 this looks great. nice job simplifying that css too. I just flagged one question: it looks like you may be hiding the currently selected assistant. but regardless of that, looks good to merge

app/views/conversations/show.html.erb
```diff
@@ -7,50 +7,35 @@
...
-        <div class="sticky top-0 pt-3 mb-1 bg-black z-10">
-          <%= render 'assistant_sidebar_link', title: "ChatGPT", settings: false %>
-        </div>
-        <%= render 'assistant_sidebar_link', title: "Samantha" %>
-        <%= render 'assistant_sidebar_link', title: "Samantha but really long so it wraps" %>
-        <%= render 'assistant_sidebar_link', title: "Samantha", selected: true %>
-        <%= render 'assistant_sidebar_link', title: "Samantha" %>
+    <div id="left-content-container" class="flex flex-col flex-grow overflow-y-auto overflow-x-clip scrollbar-hide pr-5">
+      <div class="sticky top-0 pt-3 bg-black">
+        <%= render 'assistant_sidebar_link', assistant: @conversation.assistant, settings: false, selected: true %>
```
robacarp, line 16: @krschacht right here

app/controllers/conversations_controller.rb
robacarp: The current assistant is manually displayed at the top. In order to always have it at the top, I didn't want to fiddle with searching the list. Maybe it shouldn't always be at the top? I didn't consider that. Anyway, the logic is in the view and shows the current assistant at the top of the sidebar, in a floating/sticky panel. The rest of the list of assistants is attached to the list of conversations, which is how I remember the OpenAI interface being. I pinged you on the location.

krschacht: Ohh, that seems reasonable enough. We'll try that.

This all looks good, it's just the test failure.

app/views/conversations/show.html.erb
krschacht: If you include the currently selected assistant within the collection, then separating out this HTML becomes a little tricky. But you could move this stick-wrappy-div into the assistant_sidebar_link. When you render a collection, rails gives you a free index local variable. It'll probably be called `assistant_counter` so you can just check if it's 0 and if so render that extra div around it.

app/controllers/conversations_controller.rb
robacarp: I agree with your suggestion. The only real drawback to that approach is that Chat functionally uses the top GPT as their logo on the page. I don't really think that matters as much to this project.

app/views/conversations/show.html.erb
robacarp: This actually worked out nicely, but it did require updating the rails version slightly because I tripped on this bug: https://github.com/rails/rails/issues/49761

Before that merge, the `_counter` variables weren't populated when a partial had strict locals enabled.

PR #96, opened 2024-01-30
Automatically title new Conversations

### LINKED ISSUE #93: Get new conversations to title themselves automatically so everything is not "untitled"
I need to pay more attention to what ChatGPT does, although I don't think it does a particularly good job because the titles of my conversations are almost unusable. But the experience I've been considering is:

1. After you submit your first message, which creates a conversation, we hit a GPT-3.5 endpoint using our API and a custom prompt to come up with a title based on that. The conversation.title gets updated with that.
2. Right after that the first reply from the API will come back. We don't do anything then, we wait for the user to send a second message and we wait for the second reply from the API to come back, then we send all 4 messages in the conversation so far to our endpoint again to extract a new conversation.title and update it again.

Hopefully it can be a decent title by this point. Although, if the user were to edit it in between step 1 and step 2 we wouldn't want to overwrite it. We'd either need to track a bit to know if the user edited the title -or- we would need to not do step 2. Probably we should just do step 1 as a v1 to keep things simple.

krschacht:
Fixes #93 

This PR does:

- [x] Adds a ChatCompletionAPI singleton class for one-off calls to GPT
- [x] Adds an async job that takes a conversation_id and titles it, using this GPT API
- [x] Adds an after_create hook within the Conversation to call this job

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,6 +50,8 @@ gem "bootsnap", require: false
 gem "image_processing", "~> 1.2"
 gem "rails_heroicon", "~> 2.2.0"
 
+gem "rest-client"
+
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem "debug", platforms: %i[mri windows]
@@ -83,7 +85,6 @@ group :test do
   gem "selenium-webdriver"
   gem "simplecov", require: false
   gem "rails-controller-testing"
-  gem "webmock"
   gem "mocha"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -83,7 +83,6 @@ group :test do
   gem "selenium-webdriver"
   gem "simplecov", require: false
   gem "rails-controller-testing"
-  gem "mocha"
 end
 
 gem "ruby-openai", "~> 6.3"
--- a/Gemfile
+++ b/Gemfile
@@ -84,6 +84,7 @@ group :test do
   gem "simplecov", require: false
   gem "rails-controller-testing"
   gem "webmock"
+  gem "mocha"
 end
 
 gem "ruby-openai", "~> 6.3"
--- a/Gemfile
+++ b/Gemfile
@@ -50,8 +50,6 @@ gem "bootsnap", require: false
 gem "image_processing", "~> 1.2"
 gem "rails_heroicon", "~> 2.2.0"
 
-gem "rest-client"
-
 group :development, :test do
   # See https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-debug-gem
   gem "debug", platforms: %i[mri windows]
--- a/Gemfile
+++ b/Gemfile
@@ -83,6 +83,7 @@ group :test do
   gem "selenium-webdriver"
   gem "simplecov", require: false
   gem "rails-controller-testing"
+  gem "webmock"
 end
 
 gem "ruby-openai", "~> 6.3"
--- a/app/jobs/autotitle_conversation_job.rb
+++ b/app/jobs/autotitle_conversation_job.rb
@@ -1,0 +1,38 @@
+class AutotitleConversationJob < ApplicationJob
+  queue_as :default
+
+  def perform(conversation_id)
+    conversation = Conversation.find(conversation_id)
+    Current.user = conversation.user
+
+    message = conversation.messages.order(:id).first
+    new_title = generate_title_for(message.content_text)
+    conversation.update!(title: new_title)
+  end
+
+
+  private
+
+  def generate_title_for(text)
+    json_response = ChatCompletionAPI.get_next_response(system_message, [text], response_format: {type: 'json_object'})
+    json_response['topic']
+  end
+
+  def system_message
+    <<~END
+      You extract a 2-4 word topic from text. I will give the text of a chat message that someone sent. You reply with the
+      topic of this message, but summarize the topic in 3 words.
+
+      Example:
+      ```
+      when a rails view is rendering a collection, within that collection I want to know if I'm rendering the first item
+      of the collection so I can have a conditional to render it differently
+      ```
+
+      Your reply (always do JSON):
+      ```
+      { topic: "Rails render collection counter" }
+      ```
+    END
+  end
+end
--- a/app/jobs/autotitle_conversation_job.rb
+++ b/app/jobs/autotitle_conversation_job.rb
@@ -21,7 +21,7 @@ class AutotitleConversationJob < ApplicationJob
   def system_message
     <<~END
       You extract a 2-4 word topic from text. I will give the text of a chat message that someone sent. You reply with the
-      topic of this message, but summarize the topic in 3 words.
+      topic of this message, but summarize the topic in 2-4 words.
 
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -83,6 +85,7 @@ group :test do
   gem "selenium-webdriver"
   gem "simplecov", require: false
   gem "rails-controller-testing"
+  gem "mocha"
```
krschacht, line 88: I don't think there is a built-in way within mini test to mock things. Since the code is now making API calls I want to exercise most of the code but short circuit right where the API call is made. Mocha is the way I've done this in the past and it's always felt pretty clean, but I'm open to other suggetsions.

app/jobs/autotitle_conversation_job.rb
```diff
@@ -0,0 +1,38 @@
+class AutotitleConversationJob < ApplicationJob
+  queue_as :default
+
+  def perform(conversation_id)
+    conversation = Conversation.find(conversation_id)
+    Current.user = conversation.user
```
krschacht, line 6: ChatCompletionAPI grabs the openai_key from Current.user

app/services/chat_completion_a_p_i.rb
```diff
@@ -0,0 +1,154 @@
...
+
+        params[:messages][-1][:content] = content[0...50000] + "\n...\n" + content[-50000...]
+        retry
+      else
+        puts "Error: retried #{retries} times. Error: #{e}"
+        #binding.pry  unless e.message == 'length'
+        sleep 2
+        retry  if keep_retrying
+      end
+    end
```
robacarp: This rescue block makes me uncomfortable, but I know some of it is forged with hard won lessons. Here are some things that make me skeptical:

- This rescues every exception, and will blindly retry even when exceptions occur that have nothing to do with the http api code above. If someone writes a bug in lines 72-79 it'll retry even a successful API request, burning API credits in the process. Having a dedicated `rescue RestClient::Exceptions::ReadTimeout =>` clause would draw the scope in a bit.
- The retry counting logic is muddled with stuff which checks the exception message for an arbitrary "length" string and checks the length of some parameter...and then changes the parameter? There's some kind of heuristic going on here which implicitly represented, but I think if it were explicit it would be easier to understand. Would it be possible to separate the "retry the request when it fails and needs to be restructured" logic out into a different place on the stack?
- The raise on line 78 is also opaque, and I think it's triggering some of this logic somehow. Is it possible to represent that logic without raising a string and catching it? It feels like throwing an exception here doesn't represent an exceptional circumstance, but an ordinary one.

app/services/chat_completion_a_p_i.rb
```diff
@@ -0,0 +1,154 @@
...
+    keep_retrying = true
+
+    begin
+      verify_token_count!(params)
+      response = ""
+      finished_reason = nil
+
+      if Rails.env.test?
+        raw_response = client.chat(parameters: params)
+        response = JSON.parse(raw_response)&.dig("choices", 0, "message", "content")
```
robacarp: Since we're using a test mock, why do we need to special case the rails environment?

Gemfile
robacarp: Minitest has both [mocks](https://www.rubydoc.info/gems/minitest/Minitest/Mock#expect-instance_method) and [stubs](https://www.rubydoc.info/gems/minitest/Object#stub-instance_method). It's a minimal implementation, like the rest of minitest, but it works. I think minitest's vernacular for what a mock is and does differs a bit from the rspec world, and I always have to look it up.

I tend to think of Minitest's mocks/stubs as "syntactic salt" -- they're muddy to use, and I think they do it that way on purpose because "clean test code rarely uses mocks". 

I'm not opposed to including mocha, but maybe you won't feel the need after seeing what minitest has to offer. I made a single suggestion below with what I believe the stub would look like without mocha.

test/jobs/autotitle_conversation_job_test.rb
```diff
@@ -0,0 +1,11 @@
+require "test_helper"
+
+class AutotitleConversationJobTest < ActiveJob::TestCase
+  test "sets conversation title to 'Hear me'" do
+    ChatCompletionAPI.stubs(:get_next_response).returns({"topic" => "Hear me"})
+
+    conversation = conversations(:greeting)
+    AutotitleConversationJob.perform_now(conversation.id)
+    assert_equal "Hear me", conversation.reload.title
```
robacarp, line 9: The minitest syntax for this is:

```suggestion
    ChatCompletionAPI.stub :get_next_response, {"topic" => "Hear me"} do
      conversation = conversations(:greeting)
      AutotitleConversationJob.perform_now(conversation.id)
      assert_equal "Hear me", conversation.reload.title
    end
```

robacarp (commented):
I'm stoked to see this.

Gemfile
robacarp: I think I might amend what I said about salt to clarify that the mocks are salty, but I don't think the stubs are any more difficult to use than anything else out there.

krschacht: @robacarp Oh sheesh, chatgpt had implied I needed mocha and that my brain said "oh that's right, there is no stubbing in minitest" and I didn't actually check. :) Their syntax is good enough, I'm switching.

app/services/chat_completion_a_p_i.rb
```diff
@@ -0,0 +1,154 @@
...
+      'gpt-4-vision-preview': 128000,
+      'gpt-4': 8192,
+      'gpt-4-32k': 32768,
+      'gpt-4-0613': 8192,
+      'gpt-4-32k-0613': 32768,
+      'gpt-3.5-turbo-1106': 16385,
+      'gpt-3.5-turbo': 4096,
+      'gpt-3.5-turbo-16k': 16385,
+      'gpt-3.5-turbo-instruct': 4096
+    }[name.to_sym]
```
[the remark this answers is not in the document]
krschacht: Interesting. I'm really surprised that converting a string to a symbol is expensive. That seems trivial.

But it's equally trivial to remove this casting so I'll do that. :)

app/services/chat_completion_a_p_i.rb
krschacht: @robacarp yea, good call.I copied this library from another project and the rescue block was actually twice as complex. :) I stripped out a lot of project-specific stuff but you're right that even what I left is pretty heavy handed.

I'll just remove the rescue for now and we'll re-build up checks as we actually need them, for this project.

app/services/chat_completion_a_p_i.rb
krschacht: I wanted to create a nice golden path for stubbing. Without this conditional, you either stub out the full method (`call_api`) and skip all the code execution, or you try to stub out `client.chat` which is complex.

I simplified this a bit more since you left this comment. It does still have the conceptual complexity of a branch, solely for the test environment. Maybe that's bad? But it has the added benefit of raising a really nice exception in the test environment because it's easy to not realize an API call is happening deep in a test.

Does it seem like a good solution, given all that?

krschacht: Actually, I know how to make this better! I'll clean it up a bit more.

PR #99, opened 2024-01-31
Get views working for creating new conversation & viewing

krschacht:
This is ready for review. I have a larger PR that I'm breaking into pieces for easier review. After I merge this PR in, the sidebar is in a little bit of a messed up state, but I'm going to do a separate PR that makes a few more changes to the sidebar. I restricted this PR mostly to moving files around, renaming them, and changing routes. What's included is:

- [x] The root_url is now assistants#index which is a smart route that grabs your first assistant and starts a new conversation with it (later we can tweak the logic for which is your "primary" assistant)
- [x] After login, you area redirected to this root_url
- [x] /register is phased out in favor of /users/new (reducing the non-restful routes)
- [x] Starting a new conversation for an assistant lives here: /assistants/123/messages/new. This view now works and submits to messages#create which creates a new message & and new conversation
- [x] After the conversation is started, continuing the conversation lives here: /conversations/123/messages. The core functionality of this view works.
- [x] These two views share a layout for messages_controller which lives in layouts/messages and inherits from the application layout


krschacht:
I see I have some broken tests I need to fix still.

app/views/layouts/application.html.erb
```diff
@@ -11,11 +11,11 @@
     <%= stylesheet_link_tag "application", 'https://fonts.googleapis.com/css?family=Inter:wght@100;200;300;400;500;600;700;800;900&display=swap', "data-turbo-track": "reload" %>
     <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
     <%= javascript_importmap_tags %>
-    <%= turbo_refreshes_with method: :morph, scroll: :preserve  %>
+    <%= turbo_refreshes_with method: :morph, scroll: :preserve %>
     <%= content_for :head %>
   </head>
   <body class="font-sans">
-    <%= yield %>
+    <%= content_for?(:messages_layout_yield) ? yield(:messages_layout_yield) : yield %>
```
robacarp, line 18: What about calling this just body?

```suggestion
    <%= content_for?(:body) ? yield(:body) : yield %>
```

robacarp (approved):
👍 one small comment but this is mostly reorg code.

I have in the past done some magic somehow to make rendering and inheriting layouts easier than this, but I don't remember what it was anymore. Less magic is better anyway I think.

PR #100, opened 2024-01-31
Get scroll-to-bottom arrow working

### LINKED ISSUE #91: Make the scroll-to-bottom arrow work

krschacht:
Fixes #91 

This should be fully working and ready. However, this doesn't have any front-end tests.

Maybe I should figure out how automated testing works w/ stimulus before merging in, or maybe I just merge in and then in a separate PR we add automated tests for the front-end interactions. Either way, I don't think we should go much further without getting good test coverage for the subtle front-end interactions since I'll start dropping context soon about all the interactions that we've built.


robacarp (approved):
💯

PR #101, opened 2024-01-31
Fix styling and interactions of the sidebar

krschacht:
- [x] A bunch of small tweaks to the sidebar styling
- [x] Get collapsing/expanding working again (and simplify the stimulus controller for this)
- [x] Get the hover/click effect working properly again with regards to cursor-pointer
- [x] Link all the sidebar elements to the correct routes

I still need to get the sidebar to properly refresh only the right frame upon clicking a new conversation, but this will require some deeper changes so I decided to wrap up this PR.

This PR is ready for review and merging in.


app/javascript/controllers/click_toggle_controller.js
```diff
@@ -8,29 +8,15 @@ import debounce from "../utils/debounce.js"
 
 export default class extends Controller {
```
krschacht, line 9: I implemented your insight about us not needing a "trigger target" since we can just bind to the stimulus action data attribute. In the process, I also managed to eliminate the nested controllers, greatly simplifying things. It turned out quite nice!

app/views/layouts/messages.html.erb
```diff
@@ -32,15 +31,13 @@
...
-              data-click-toggle-flippable-class="!hidden"
-              data-action="keydown.meta+shift+s@document->click-toggle#handleEvent
-                          keydown.meta+period@document->click-toggle#handleEvent"
-              >
+      <div  id="handle-container"
+            class="absolute left-0 top-1/2 w-8 h-20 -mt-10 z-20 hidden md:flex"
+            data-action= "keydown.meta+shift+s@document->click-toggle#toggleClass
+                          keydown.meta+period@document->click-toggle#toggleClass
+                          click->click-toggle#toggleClass"
+      >
```
krschacht, line 39: I think we could put these keydown actions anywhere, but I figured we might as well group them with the click one.

Notably, because it's a div I had to explicitly specify `click->` but if it was a link or button then that's assumed.

app/views/shared/_assistant_sidebar_link.html.erb
```diff
@@ -1,19 +1,19 @@
...
             flex justify-between items-center
-            mb-1 p-1 pr-2
+            mb-1 p-1 pl-2 pr-2 mr-5
             bg-black hover:bg-zinc-800
-            group cursor-pointer 
-            text-sm rounded-lg 
+            group cursor-pointer
+            text-sm rounded-lg
 ">
-  <%= link_to assistant, class: "ml-2 flex-1 flex items-center text-white truncate" do %>
```
krschacht, line 11: A bunch of little tweaks to get the spacing just right with the sticky header

app/views/shared/_assistant_sidebar_link.html.erb
```diff
@@ -34,12 +35,11 @@
...
     <% end %>
 
-    <%= link_to new_assistant_message_path(assistant) do %>
+    <%= link_to new_assistant_message_path(assistant), class: "inline-flex" do %>
       <%= icon "pencil-square",
         variant: :micro,
         size: 18,
-        class: "inline text-gray-400 cursor-pointer group-hover:visible hover:text-white #{selected ? 'text-white visible' : 'invisible'}",
-        data: { selected: selected },
+        class: "text-gray-400 cursor-pointer group-hover:visible hover:text-white #{selected ? 'text-white visible' : 'invisible'}",
```
krschacht, line 42: In the ~10 lines above I tweaked the placement of "inline" because it was preventing the icon from slightly depressing and shrinking in size upon click.

app/assets/stylesheets/application.css
```diff
@@ -10,6 +10,6 @@
  * files in this directory. Styles in this file should be added after the last require_* statement.
  * It is generally better to create a new file per style scope.
  *
- *= require_tree .
+ *= require_tree ./includes
```
krschacht, line 13: As I was inspecting our generated CSS in chrome inspector, I noticed something really odd. `application.tailwind.css` which lives in the /assets/stylesheets/ directory was being compiled into tailwind.css (good) but it was *also* being included raw in it's uncompiled state. I fixed it by updating this line.

The implication of this /includes change is that any additional CSS files we want to create, over time, need to be put within the /includes directory. They cannot be put directly in the stylesheets/ directory.

What's odd is that all of this is plain vanilla generated rails config, so it may be a bug with rails 7.

I considered:

1. Moving application.tailwind.css to another directory, but tailwindcss is configured to look for it where it is
2. Removing require_tree altogether and just manually requiring any future CSS files we put in stylesheets/ but that feels like it would be more annoying.

Best solution I could come up with was to kill the `require_tree .` in favor of `require_tree ./includes'

app/views/shared/_conversation_sidebar_link.html.erb
```diff
@@ -1,20 +1,21 @@
 <%# locals: (conversation:, settings: true) %>
 <% selected = conversation == @conversation %>
 
-<div class="flex rounded-lg hover:bg-zinc-800 <%= selected && 'bg-zinc-800' %> justify-between items-center group">
-  <%= link_to conversation.title, conversation, class: "flex-1 pl-2 py-2 text-white truncate text-sm text-opacity-80 w-5" %>
+<div class="flex rounded-lg hover:bg-zinc-800 <%= selected && 'bg-zinc-800' %> justify-between items-center group mr-5 cursor-pointer">
```
krschacht, line 4: Adding cursor-pointer to this line is what fixed the javascript error. We want it on this overall element because we want it to look like it's depressing upon click (that's a styled behavior I added to the cursor-pointer class).

app/assets/stylesheets/application.css
robacarp: I think that's fine. With tailwind I rarely ever need to declare styles manually. You tend to do them once or twice and then you're done. 

Actually, given how infrequent this is likely to be, it might be worth putting either a comment here in code explaining this, or moving the few declarations we have for tooltips into the includes directory to set an example to follow.

app/views/layouts/messages.html.erb
```diff
@@ -3,18 +3,17 @@
...
       data-controller="click-toggle"
-      data-click-toggle-flippable-class="md:flex"
-      data-action="keydown.meta+shift+s@document->click-toggle#handleEvent
-                   keydown.meta+period@document->click-toggle#handleEvent"
+      data-click-toggle-flippable-class="!hidden"
   >
 
     <div id="left-column"
-        class="flex-col h-screen w-[260px] bg-black hidden md:block overflow-y-auto overflow-x-clip scrollbar-hide"
+        class="flex-col h-screen w-[260px] bg-black hidden md:flex overflow-y-auto overflow-x-clip scrollbar-hide"
```
robacarp, line 10: Oh this is probably what broke the sidebar. The reason I switched it to block is, there aren't > 1 children of this div (or there weren't -- I can't tell from the diff if that's still true) so a flex is useless. Maybe it doesn't hurt? I don't really know if there's a performance hit... 👍 anyway

app/views/layouts/messages.html.erb
robacarp: agreed, with the ability to dangle these things anywhere in the source it's worth putting them near their friends so they don't get lost.

app/views/layouts/messages.html.erb
```diff
@@ -3,18 +3,17 @@
...
         data-click-toggle-target="destination"
     >
 
       <div id="left-content-container" class="flex flex-col flex-grow">
-        <div class="pl-3 pr-5 flex flex-col mb-28">
+        <div class="pl-3 flex flex-col mb-28">
+          <div id="spacer-above-sticky-assistant" class="sticky top-0 h-4 z-10 bg-black"></div>
```
robacarp, line 16: I'm generally surprised any time a z-index is needed, but in this case I'm a little more surprised.

app/views/shared/_conversation_sidebar_link.html.erb
robacarp: I'm not sure I understand all of what the nested-pointer stuff is doing yet, but that's ok. 👍

robacarp (approved):
Might be worth moving some of the CSS declarations around a bit to make the change you made to the application.css more obvious? Otherwise👍

app/assets/stylesheets/application.css
krschacht: Good idea! I'll pull them out to start the pattern.

app/views/layouts/messages.html.erb
krschacht: I believe "sticky" ensures that the subsequent items scroll underneath it, however all those items have tooltips and the tooltips have a z-index so they were floating above this sticky header as the items were scrolling underneath it. This ensures the header is above any tooltips.

app/views/layouts/messages.html.erb
krschacht: Actually, you're right this can be block. I tried changing this back to flex in my various attempts to debug what was wrong and didn't revert it.

The issue was that the data-click-toggle-class was referencing "md:flex" so changing that to "md:block" required changing the stimulus configuration. That's now been changed as part of the refactor I did so this should be block again.

PR #102, opened 2024-02-01
Preserve the sidebar scroll position

krschacht:
Rob — Take this as a RFC. Is this a good restructuring? It enabled me to solve the sidebar scroll issue and feels like a small step in the right direction towards better organizing views, but it still doesn't feel like the ultimate "best" organization. This might impact any work you're doing to stream messages in, but not in any significant way as I'm just moving files around. Hopefully it's easy for you to move a single chunk of your code if I merge this in.

---

The diff looks big but there are really just two key things I'm doing in here. It was a pretty quick so I can scrap it if it's not a good.

First, I took the `layouts/messages` file and turned it into a partial `shared/two_columns`. Then the `messages#index` and `messages#new` view use this partial and set a `content_for :message_content` in order to populate the right hand column.

After this was done, it was easy for me to make the conversations in the sidebar target the `<turbo-frame "conversation">` to update the right side.

This was enough to preserve the scroll position, but the selected state of the conversation was not changing. I accomplished this with a stimulus controller that simulates radio button behavior. (It turned out to be hard to use radio buttons because a lot of the Turbo magic relies on links being clicked, not radio buttons.)

Quick video demo so you can see the intended behavior in action: https://share.zight.com/jku8DXLL

If this is all on the right track, let me know and I'll fix up some tests that broke because of the changes to view structure.

---

Crap, I just noticed that the browser's back state isn't working. Notice in my video the browser's URL is not changing as I click through conversations. I just found something in the docs about this: https://turbo.hotwired.dev/handbook/frames#promoting-a-frame-navigation-to-a-page-visit

So I simply tried adding `data-turbo-action="advance"` to the `<turbo-frame id="conversation" ...` but doing so breaks all the behavior. Now the full page is reloading. I'm so confused... Maybe there is another way to simply push something into the back state.


app/assets/stylesheets/application.tailwind.css
```diff
@@ -21,3 +21,10 @@
 .icon.cursor-pointer:active {
   @apply scale-90 ease-in-out duration-300
 }
+
+
+/* Parent utility classes which activate children */
+
+.relationship-flex .relationship\:flex {
+  @apply !flex
+}
```
krschacht, line 30: This let's me apply "relationship-flex" to a parent element and any children that have "relationship:flex" get the flex activated. I'm copying the tailwind notation from "group-hover:..."

app/models/message.rb
```diff
@@ -28,6 +28,6 @@ def set_default_role
   end
 
   def broadcast_message
-    broadcast_append_to conversation, partial: "messages/message", locals: { scroll_into_view: true, conversation: conversation }
+    broadcast_append_to conversation, partial: "messages/message", locals: { scroll_into_view: true }
   end
```
krschacht, line 32: conversation can be inferred from message.conversation

app/views/conversations/_conversation.html.erb
```diff
@@ -1,16 +1,125 @@
...
-  <p class="flex gap-2">
-    <strong class="block font-medium mb-1">Assistant:</strong>
-    <%= conversation.assistant.name %>
-  </p>
-
-  <p class="flex gap-2">
-    <strong class="block font-medium mb-1">Title:</strong>
-    <%= conversation.title %>
-  </p>
-</div>
```
krschacht, line 16: This whole chunk is lifted straight from the previous layout/message — it's the whole contents of the right column

app/views/messages/_message.html.erb
```diff
@@ -1,9 +1,12 @@
-<%# locals: (message:, conversation:, scroll_into_view: false, request_id: nil) -%>
-<div 
+<%# locals: (message:, scroll_into_view: nil, message_counter: -1) -%>
```
krschacht, line 1: I don't get why `message_counter: -1` is needed but I saw you did that elsewhere

robacarp: This is a workaround for the fact that this variable doesn't exist when the partial is rendered without a collection. When you render with a collection this will be overridden with the actual counter index, but without this magic value means the counter == 0 logic won't be fired.

app/javascript/controllers/radio_behavior_controller.js
robacarp, line 1: Nice, I like it.

app/javascript/controllers/radio_behavior_controller.js
```diff
@@ -0,0 +1,25 @@
...
+export default class extends Controller {
+  static classes = [ "selected" ]
+  static targets = [ "radio" ]
+
+  select(event) {
+    this.radioTargets.forEach(element => {
+      element.classList.remove(...this.selectedClasses)
+    })
+
+    event.target.parentNode.classList.add(...this.selectedClasses)
```
robacarp, line 23: I'm surprised this needs to navigate up to the parent. Why is that? Does it need to match to make sure that the parentNode is one of the targets?

robacarp (commented):
I think this is a good direction. I'm a little unclear about the flex inheritance thing, but maybe that's because I don't yet understand why you'd need to toggle the flex state of the parent with the child. I think it's worth finishing up.

app/views/shared/_two_columns.html.erb
```diff
@@ -1,36 +1,37 @@
...
       </div>
-    </div>
 
-    <div id="right-column" class="flex flex-1 flex-col h-screen bg-white dark:bg-graphite relative">
+      <%= render partial: "shared/user_menu" %>
+    </div>
+  </div>
 
+  <div id="right-column" class="flex flex-1 h-screen bg-white dark:bg-graphite">
+    <turbo-frame id="conversation" class="flex flex-1 flex-col relative">
```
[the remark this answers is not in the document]
krschacht: Hmm... yea, conceptually that does make more sense for this to live in index.html. However, I intentionally am wrapping the turbo-frame around messages + composer. I'm considering that whole right column to be the "conversation".

And the new.html includes the composer too. In place of the messages we'll eventually add the placeholder "tips for new conversation".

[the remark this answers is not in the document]
robacarp: If this were an application wide layout, it'd be clear to have the layout contain a left and right side, and the name of  the content boxes would be vague: "sidebar" and "content-area" or something.  If one page has the composer and message list in it, that's just one use of it. 

The way you have it here, the column layout is tied to the right hand side being for "messages," which means that any use of that which isn't messages is going to require a refactor or be a wedge. I'm fuzzy on what the "rest of" the site is going to look like, or if anything even needs to exist outside of this two column layout. 

Since it's all within messages, I don't think it matters that much.

app/javascript/controllers/radio_behavior_controller.js
krschacht: It's good to point that out. That's brittle. It was intentional becuase I wand the radio class to be applied to a div which wraps the link href. I cannot put the data-action on the wrapping div because the href steals the click (I think because of stimulus).

I've updated the code to not just navigate to parent but search upward for the closest element. That will make this less brittle. And I updated the example to explain this.

PR #103, opened 2024-02-01
Fix back-button history for turbo-frames

krschacht:
There is an issue with turbo-frame elements where you can target them to be updated while preserving the overall page state, but the URL does not change to reflect this and the browse back button does not work. This may be intended behavior or it may be a bug with Turbo, I can't tell. I have an open issue with them here: https://github.com/hotwired/turbo/issues/1156

But regardless, we want it to work because clicking a conversation in the left sidebar refreshes the turbo-frame id="conversation" in the right side. We do the frame refreshing because we want the scroll position of the left side to remain preserved.

This works by manually updating the browser's pushState and manually handling the back button for these cases. Demo here: https://share.zight.com/jku8YrZLj

### DIFF
--- a/app/javascript/controllers/turbo_frame_history_controller.js
+++ b/app/javascript/controllers/turbo_frame_history_controller.js
@@ -1,0 +1,39 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+
+  connect() {
+    document.addEventListener('turbo:before-fetch-request', this.beforeFetchRequest)
+    window.addEventListener("popstate", this.popState)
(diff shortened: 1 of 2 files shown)


robacarp (approved):
this is awesome. it would be good to write a rails system test for this, I don't think it would be hard at all to represent this in there.

krschacht:
> this is awesome. it would be good to write a rails system test for this, I don't think it would be hard at all to represent this in there.

Good call, I'll start a new PR with tests. I want to add tests for a few of these stimulus things I've done recently.

PR #105, opened 2024-02-02
Ensure an authenticated user only creates messages for their conversations

keithschacht:
I recently updated the way Messages are created: Messages are linked to Assistant and the conversation_id is not an optional param. When it's provided, the message is linked to that conversation. When it's not, a new conversation is created on-the-fly when a Message is created.

But this introduced a security issue where not an authenticated user could specify any arbitrary conversation_id for the #create endpoint. This PR fixes that security issue by validating the conversation_id.

I debated whether the check should be in the controller or on the model, and I opted for putting it in the model. However, the model does need to support creating Messages for any arbitrary conversation_id since workers will be creating messages, so I use Current.user as the key.

Meaning, if a Message is being created created and a conversation is explicitly provided — we check whether Current.user is set. If it is, then it's clear we're executing within the context of a specific user and we confirm that the conversation in play is owned by that user.

I added tests for all these cases.

### DIFF
--- a/test/controllers/assistants_controller_test.rb
+++ b/test/controllers/assistants_controller_test.rb
@@ -2,8 +2,8 @@ require "test_helper"
 
 class AssistantsControllerTest < ActionDispatch::IntegrationTest
   setup do
-    @assistant = assistants(:samantha)
-    @user = @assistant.user
+    @user = users(:keith)
+    @assistant = @user.assistants.order(:id).first
     login_as @user
   end
 


db/migrate/20240202162322_alter_column_instructions_on_runs_to_allow_nil.rb
```diff
@@ -0,0 +1,5 @@
+class AlterColumnInstructionsOnRunsToAllowNil < ActiveRecord::Migration[7.1]
+  def change
+    change_column :runs, :instructions, :string, null: true
+  end
```
krschacht, line 4: In order to test for this case, I needed to flesh out the fixtures so there is another user (named Rob :) ) and this user has a conversation. In the process I discovered a small bug int he domain model: instructions should be optional. I went ahead and fixed that bug within this PR.

test/fixtures/runs.yml
```diff
@@ -12,7 +12,7 @@ hear_me_response:
   model: gpt-4
   instructions: You are a helpful assistant
   additional_instructions:
-  tools: [{"type": "retrieval"}, {"type": "code_interpreter"}]
+  tools: [{"type": "code_interpreter"}, {"type": "retrieval"}, {"type": "function"}]
```
krschacht, line 15: A small bug in the fixtures: these tools should have matched the tools of the assistant. Correcting it.

PR #106, opened 2024-02-02
Add system tests for key UI interactions

krschacht:
- [x] Test navigation when clicking conversations in the sidebar
- [x] Test sidebar open/close interaction
- [x] Test keyboard shortcuts to focus & unfocus the composer
- [x] Submit button disabled in composer until you type text then it enables
- [x] Ensuring Enter submits the composer and Shift+Enter inserts a newline
- [x] Ensuring clicks to the "new" icons, in sidebar and top-right, start a new conversation
- [x] The scroll down arrow properly appearing, disappearing, and scrolling down when clicked
- [x] New message added to a conversation stream in — this situation is actually broken but we have a test for it

The ones I did not get done but may do soon:

* Breakpoints for resizing browser width

Things to defer until we finish implementing them:

* Ensuring clicking Logout in sidebar works
* Drop-down menus in sidebar and in assistant's name at the top of the screen
* Settings link in sidebar (since we're going to move that out of a modal)

### DIFF
--- a/app/views/shared/_two_columns.html.erb
+++ b/app/views/shared/_two_columns.html.erb
@@ -1,7 +1,7 @@
 <%= turbo_stream_from @conversation %>
 
 <div class="flex h-screen"
-    data-controller="click-toggle"
+    data-controller="click-toggle turbo-frame-history"
     data-click-toggle-flippable-class="!hidden"
 >
 
--- a/app/views/shared/_two_columns.html.erb
+++ b/app/views/shared/_two_columns.html.erb
@@ -13,10 +13,12 @@
     <div id="left-content-container" class="flex flex-col flex-grow">
       <div class="pl-3 flex flex-col mb-28">
         <div id="spacer-above-sticky-assistant" class="sticky top-0 h-4 z-10 bg-black"></div>
-        <div  data-controller="radio-behavior"
-              data-radio-behavior-selected-class="relationship"
-        >
-          <%= render partial: "shared/assistant_sidebar_link", collection: @sidebar_assistants, as: :assistant %>
+        <div id="assistants">
+          <div  data-controller="radio-behavior"
+                data-radio-behavior-selected-class="relationship"
+          >
+            <%= render partial: "shared/assistant_sidebar_link", collection: @sidebar_assistants, as: :assistant %>
+          </div>
         </div>
 
         <turbo-frame  id="conversations"
--- a/app/views/shared/_two_columns.html.erb
+++ b/app/views/shared/_two_columns.html.erb
@@ -76,7 +76,7 @@
       <div id="wide-header" class="flex-none justify-between ml-2 mt-2 mb-3 h-10 text-black dark:text-white text-lg hidden md:flex">
         <%= render 'shared/assistant_header' %>
 
-        <%= link_to new_assistant_message_path(@assistant), class: "pencil block cursor-pointer inline-flex mr-3 px-2 py-2 group" do %>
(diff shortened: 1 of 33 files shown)


test/fixtures/messages.yml
```diff
@@ -23,6 +25,7 @@ identify_photo:
   content_text: What do you see in this photo?
   content_document: # the cat photo is not attached in this field, currently it's attached at the document level. info: https://platform.openai.com/docs/assistants/tools/knowledge-retrieval and https://platform.openai.com/docs/guides/vision
   run:
+  created_at: 2023-12-30 1:02:00
```
robacarp, line 28: Instead of hardcoding these, which introduces a maintenance burden on these files, what do you think of declaring a T value at the top of the file and then incrementing it by one minute for each of these? Like this:

```yml
<% t = DateTime.now %>

identify_photo:
  created_at: <% t += 1.minute %>
```

Or, if it's easy to do, declaring a helper function somewhere that's available in the yaml files?

test/application_system_test_case.rb
```diff
@@ -1,17 +1,81 @@
...
+      find(selector_or_element, wait: wait)
+    end
+
+    unless element.matches_css?(".tooltip") # sometimes we're checking the tooltip on a link but within the link is an icon, check that instead
+      element = element.find(:xpath, './*', match: :first, wait: wait)
+    end
+
+    assert element.matches_css?(".tooltip")
+    assert_equal text, element[:'data-tip'], "Expected element to have tooltip #{text}. #{error_msg}"
+  end
```
robacarp, line 57: These are nice helpers, they clean up the tests well.

robacarp (approved):
One little idea about the timestamps in the fixtures, take it or leave it.

👍

test/fixtures/messages.yml
krschacht: That's a good idea when we need the timestamp to be something specific. In this case, having it be in the past at some fixed point totally works. This is just populating "old" / existing conversations for a user so we shouldn't ever need to update these timestamps.

In this PR my update was because I had left the time off. I believe fixtures do not guarantee the order that it will insert records, so I was getting a flaky test when I was expecting the `.messages.first` to be consistent and it wasn't. Now we order messages by created_at so this ensures they'll be consistent in our tests.

app/views/shared/_two_columns.html.erb
```diff
@@ -150,7 +152,7 @@
               %>
 
               <div id="send-button" class="absolute right-8 bottom-4 w-[39px] h-[39px]">
-              <%= form.button data: { new_message_target: "submit" }, disabled: true, class: %w|
+              <%= form.button data: { new_message_target: "submit" }, disabled: true, id: "submit", class: %w|
```
krschacht, line 155: do I need this?

krschacht (commented):
@robacarp As I played with this whole interaction more, there is a weird issue with jumping of the page. As I submit a message in the composer, it appears and then the page jumps and it appears again.

I started commenting things out and I'm pretty sure morphing has stopped working. The best clue, before you comment anything out, is to scroll the left sidebar so it's offset. Then submit a message in the composer and you'll see the left sidebar jump back to the top as the full page refreshes itself. I started commenting things out one by one and I think the core morph isn't working. The mere act of posting to "create" and having that redirect back to it's same place; that's not morphing.

But I went back to main and it's not morphing there either so let's get your PR in after you checkout my comments and we can track this down separately.

PR #107, opened 2024-02-05
implements streaming messages into conversations

robacarp:


app/javascript/controllers/turbo_frame_history_controller.js
```diff
@@ -2,15 +2,15 @@ import { Controller } from "@hotwired/stimulus"
...
+  // }
 
-  disconnect() {
-    document.removeEventListener('turbo:before-fetch-request', this.beforeFetchRequest)
-    window.removeEventListener("popstate", this.popState)
-  }
+  // disconnect() {
+  //   document.removeEventListener('turbo:before-fetch-request', this.beforeFetchRequest)
+  //   window.removeEventListener("popstate", this.popState)
+  // }
```
robacarp, line 13: This got weird, and I don't think I have context enough about when and where Turbo is refreshing to correct it.

The behavior is, when you send a message it works, and the reply starts streaming back in. Then, when the reply is done streaming, there's a turbo broadcast refresh. That fails because it somehow tries to run a GET to a nonexistent url.

app/jobs/get_next_ai_message_job.rb
```diff
@@ -0,0 +1,45 @@
+class GetNextAiMessageJob < ApplicationJob
+  def perform(conversation_id)
+    puts "GetNextAiMessageJob.perform(#{conversation_id})"
```
robacarp, line 3: I'd normally leave this kind of thing up to the queue to do, but it's really terse in it's output right now. I found it helpful to have this in the logs so that I could see that _something_ was happening.

app/jobs/get_next_ai_message_job.rb
```diff
@@ -0,0 +1,45 @@
...
+      parameters: {
+        model: "gpt-3.5-turbo",
+        messages: messages,
+        temperature: 0.8,
+        stream: build_response_handler,
+        n: 1
+      }
+    )
+
+    @conversation.broadcast_refresh
```
robacarp, line 22: After the message is finished streaming, the #chat call above returns and we know it's done streaming. This manually broadcasts a refresh request to the browser. It might not be necessary, but there were times where something rendered weirdly and this fixed the glitch.

app/jobs/get_next_ai_message_job.rb
```diff
@@ -0,0 +1,45 @@
...
+    puts "Finished GetNextAiMessageJob.perform(#{conversation_id})"
+
+  rescue => e
+    puts "Error in GetNextAiMessageJob: #{e.inspect}"
+    puts e.backtrace
+  end
```
robacarp, line 28: SolidQueue has some error handling stuff, but I haven't spent any time digging into it yet.

app/models/message.rb
```diff
@@ -11,14 +11,20 @@ class Message < ApplicationRecord
   before_validation :set_default_role, on: :create
   before_validation :create_conversation, on: :create, if: -> { conversation.blank? }
 
-  validates :content_text, :role, presence: true
-  validates :run, presence: true, if: -> { assistant? }
+  validates :role, presence: true
+  validates :content_text, presence: true, unless: :assistant?
```
robacarp, line 15: This change is required because when a reply is streaming in, the initial message is blank.

app/views/conversations/show.html.erb
```diff
@@ -106,47 +106,45 @@
             <%= icon "paper-clip", variant: :mini, size: 23, class: 'cursor-pointer text-zinc-600 hover:text-zinc-900 stroke-2 dark:text-gray-300 dark:hover:text-white', title: 'Attach', tooltip: :top %>
           </div>
 
-          <turbo-frame id="new-message-form">
```
robacarp, line 109: The turbo frame here didn't seem to be required. But maybe that's what broke the javascript history thing above?

app/views/messages/_message.html.erb
```diff
@@ -22,9 +22,7 @@
         <%= message.user? ? 'You' : message.conversation.assistant.name %>
       </div>
 
-      <div class="text-zinc-700 dark:text-white" data-role="content_text">
-        <%= message.content_text %>
-      </div>
+      <div class="text-zinc-700 dark:text-white whitespace-pre-wrap" data-role="content_text"><%= message.content_text %></div>
```
robacarp, line 25: In order to get paragraph breaks in Chat's response to format correctly, I added whitespace-pre-wrap...but that means this all needs to be on one line or it renders weirdly.

app/views/messages/index.html.erb
```diff
@@ -2,10 +2,10 @@
 
   <div class="min-w-full">
     <turbo-frame id="messages">
-      <%= render @conversation.messages %>
+      <%= render @messages %>
```
robacarp, line 5: This was rendering messages out of order, because the #sorted method wasn't being called on the collection. I was also querying for messages twice.

app/javascript/controllers/turbo_frame_history_controller.js
krschacht: It's okay to leave it commented out. I have an open github issue with Turbo about a better solution to this.

app/views/messages/_message.html.erb
krschacht: It's probably worth adding an HTML comment so someone doesn't reformat it in the future

app/javascript/controllers/turbo_frame_history_controller.js
```diff
@@ -22,9 +22,10 @@ export default class extends Controller {
     // This "fixes" history state for turbo frames, but it doesn't use the Turbo page cache.
     // This simply means that going back will cuase a new server request, but that's okay for our
     // purposes.
-    let fetchTargetingFrame = (event.detail.fetchOptions.headers['Turbo-Frame'] != undefined)
+    let getFetchTargetingFrame =  (event.detail.fetchOptions.headers['Turbo-Frame'] != undefined) &&
+                                  (event.detail.fetchOptions.method == "get")
```
krschacht, line 26: I figured out what the issue was. The POST to the form was getting added to the back state history. This extra check prevents that. I tested locally and it should be working now.

app/views/messages/index.html.erb
```diff
@@ -1,8 +1,10 @@
 <%= content_for :messages_content do %>
 
+  <%= turbo_stream_from @conversation %>
```
krschacht, line 3: I moved the stream_from listener out of the `two_columns` and into this in order to fix a bug. In the old place, when you load the page and click to a new conversation on the left side, it would not reconnect the stream_from listener. When I moved it inside here, now it continues to stream in new responses even when you change conversations.

krschacht (approved):
It's working! I was actively testing in the browser and fixed a couple things as I was testing. Comments inline.

A couple small suggestions too, but this is basically ready to merge in. It's exciting to see this thing come to life.

app/views/conversations/show.html.erb
krschacht: We actually need to delete this file. We aren't using this view any longer.

krschacht:
@robacarp As I played with this whole interaction more, there is a weird issue with jumping of the page. As I submit a message in the composer, it appears and then the page jumps and it appears again.

I started commenting things out and I'm pretty sure morphing has stopped working. The best clue, before you comment anything out, is to scroll the left sidebar so it's offset. Then submit a message in the composer and you'll see the left sidebar jump back to the top as the full page refreshes itself. I started commenting things out one by one and I think the core morph isn't working. The mere act of posting to "create" and having that redirect back to it's same place; that's not morphing.

But I went back to main and it's not morphing there either so let's get your PR in after you checkout my comments and we can track this down separately.

app/javascript/application.js
```diff
@@ -1,6 +1,3 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "@hotwired/turbo-rails"
 import "controllers"
-import Rails from "@rails/ujs"
-
-Rails.start()
```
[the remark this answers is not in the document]
robacarp: No, it wasn't interfering. I was reading through the network traffic and saw it pull in ujs.

app/views/conversations/show.html.erb
robacarp: Ok, done.  I had to delete a bunch of other tests and related content to match, but that probably means we have less dead ends hanging around. 👍

app/jobs/get_next_ai_message_job.rb
```diff
@@ -0,0 +1,23 @@
...
+    @conversation = Conversation.find conversation_id
+    messages = @conversation.messages.sorted.map(&:for_openai)
+    @new_message = @conversation.messages.create! role: :assistant, content_text: "", assistant: @conversation.assistant
+    @new_message.broadcast_append_to @conversation
+
+    response = AiBackends::OpenAi.new(@conversation.user.openai_key)
+      .get_next_chat_message(messages) do |content_chunk|
+        @new_message.content_text += content_chunk
+        @new_message.broadcast_replace_to @new_message.conversation, locals: { scroll_into_view: true }
+      end
```
robacarp, line 14: @krschacht I'm actually pretty happy with this cleaner API. It's nice to be able to hide the mess of building the proc behind a veil.

app/services/ai_backends/open_ai.rb
```diff
@@ -0,0 +1,51 @@
...
+  #
+  # In regular tests, mock this method or the TestClients::OpenAi class to do
+  # what you want instead.
+  def self.client
+    if Rails.env.test?
+      TestClients::OpenAi
+    else
+      OpenAI::Client
+    end
+  end
```
robacarp, line 16: @krschacht this isn't my favorite -- special casing the test environment feels like playing with fire to me. But this abstraction layer did clean up the interface nicely anyway so I'll call it even.

app/services/test_clients/open_ai.rb
robacarp, line 1: This feels like the wrong place to put this file, but Zeitwork doesn't seem to want me to put it in the testing folder without a bunch of hijinks.

test/system/messages/composer_test.rb
```diff
@@ -39,7 +39,7 @@ class MessagesComposerTest < ApplicationSystemTestCase
     send_keys "Entered text so we can now submit"
     refute @submit.disabled?
     @submit.click
-    assert_not_equal path, current_path
+    # assert_not_equal path, current_path
```
robacarp, line 42: @krschacht I think you fixed this somewhere else?

krschacht: @robacarp Yes, in my PR which branches off this one, I went ahead and fixed this and added a bunch more tests. So you can keep commented out in this PR just so tests don't break (assuming it's still broken)

krschacht:
@robacarp nice to see you got a wedge! You might be doing more on this, but I'm good with you merging in whenever it's ready. It looks like tests are passing.

Then I'll land my PRs #113 and #120 which branch off this

app/javascript/application.js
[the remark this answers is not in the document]
krschacht: @robacarp I just discovered that elsewhere in our app, I added a `data: { confirm: "Are you sure?" },` to a form button and it didn't work. I tracked it down to this removal of UJS. ChatGPT is telling me that there is some baseline rails functionality such as this delete confirmation that are still handled by UJS and that it can coexist with stimulus. So I think we should add it back. Any thoughts?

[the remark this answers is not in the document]
robacarp: I'm not opposed, but I think my preference would be to just make it a stimulus controller.

PR #113, opened 2024-02-07
Get the conversation to auto-title

### LINKED ISSUE #104: Get the conversation title to update in the sidebar after it's automatically titled by the worker

krschacht:
Fixes #104 

When a conversation is first created, it has no title and the sidebar renders that as "New chat". After the second message is created on a conversation, we use GPT-3.5 in order to analyze the contents of the conversation and automatically give it a reasonable title.


app/models/conversation.rb
```diff
@@ -6,7 +6,7 @@ class Conversation < ApplicationRecord
   has_many :runs, dependent: :destroy
   has_many :steps, dependent: :destroy
 
-  after_create_commit :set_title_async, if: -> { title.blank? }
+  after_create_commit :set_title_async, if: -> { title.blank? && messages.count >= 1 }
```
krschacht, line 9: I struggled a bit with the trigger. Initially, after_create_commit + title.blank? seemed right. But `Message.create` creates a conversation a split second before a Message so if the job runs too fast it could run before the message has been created.

Initially, I tried changing this to `after_update_commit` since the creation of a message updates the conversation timestamp and then I added `messages.count >= 1`. But we update conversation A TON because every single update to a message fragment while it's streaming in. I actually wonder if we need to make that more efficient.

So I changed this back to after_create_commit and I added a 2 second delay (down in set_title_async).

But... there's a possibility it runs once and still fails to find any messages. Maybe I need to add a delayed retry.

app/models/conversation.rb
```diff
@@ -58,6 +58,6 @@ def self.grouped_by_increasing_time_interval_for_user(user)
   private
 
   def set_title_async
-    AutotitleConversationJob.perform_later(id)
+    AutotitleConversationJob.set(wait: 2.seconds).perform_later(id) # it's more important to let the queue prioritize the message response so we delay this a bit
   end
```
krschacht, line 62: Further up in this class, the trigger is `after_create_commit` if `title.blank?` . I struggled with the proper trigger because Message.create creates a conversation a split second before a Message is created, so if the job runs too fast it could run before the message has been created.

Initially, I tried changing this to after_update_commit since the creation of a message updates the conversation timestamp and then I added messages.count >= 1. But we update conversation A TON because every single update to a message fragment while it's streaming in and so it was creating a ton of duplicate set_title jobs. (I actually wonder if we need to make the quantity of messages updates more efficient?)

So I changed this back to after_create_commit and I added a 2 second delay (down in set_title_async).

But... there's a possibility it runs once and still fails to find any messages. I just added a retry mechanism to help with that.

app/jobs/autotitle_conversation_job.rb
```diff
@@ -1,11 +1,16 @@
+class ConverstionNotReady < StandardError; end
+
 class AutotitleConversationJob < ApplicationJob
   queue_as :default
+  retry_on ConverstionNotReady
```
krschacht, line 5: Apparently the retry_on defaults to 3.second delay and up to 5 retries. Those felt like reasonable defaults.

app/jobs/autotitle_conversation_job.rb
```diff
@@ -1,11 +1,16 @@
+class ConverstionNotReady < StandardError; end
```
robacarp, line 1: I think I might try to move this down into the class rather than at the top level so it doesn't exist in the global namespace. Does that work?

robacarp (approved):
While testing the streaming stuff I asked Chat a bunch of terse conversation starters, eg "Write a long story".

The behavior was actually that Chat titled the conversation based off of the _response_ it gave rather than my question. [Here is a conversation where it did exactly that](https://chat.openai.com/share/4a7b5647-be5e-496f-b1ec-c0db224cfb5a). At some point we may want to delay sending this until some amount of "meaningful text" is present in the messages.

Regardless, this is good, 👍

app/jobs/autotitle_conversation_job.rb
krschacht: Ah, I thought it had to be top-level. But I just tried it within the class and it worked. Good suggestion

krschacht:
@robacarp I added tests and cleaned this up a bit. It's now ready to merge in, but I branched off your branch so we'll wait until yours lands first. No rush

PR #120, opened 2024-02-08
Quick pass and cleaning up code cruft

### LINKED ISSUE #118: Remove unused code (e.g. chats, replies, conversation#show)

keithschacht:
Fixes #118 

Removes the old `replies` and `notes` models.
Removes some scaffolded views we aren't using (and I don't think we will use)
Cleans up some irrelevant tests

There is more cleanup we can do, but it's progress.


krschacht:
@robacarp Actually, one problem with merging this in is that it may be helpful to keep Ali's old replies & notes models in place while we're trying to get message uploading working. Maybe I'll wait to merge in until after we get image attachments working.

PR #124, opened 2024-02-09
Stop using href="#" to fix page scroll jumpiness

krschacht:
There is one place I used href="#" to call a stimulus action, but the data-action was missing a :prevent so it was causing some odd behavior because the link anchor was being activated. It took me awhile to track down this bug. I fixed it by converting it to a button tag. I realized it's probably always preferable for us to use button_tags over href="#" links so I converted the other instances of them as well. This way we don't have to remember to do :prevent.


app/views/shared/_two_columns.html.erb
```diff
@@ -103,7 +103,7 @@
 
         <div id="right-footer" class="flex-none flex flex-col relative">
           <div id="scroll-button" class="text-center absolute -top-14 left-1/2 transform -translate-x-1/2 hidden" data-scrollable-target="widget">
-            <%= link_to "#",
+            <%= button_tag type: "button",
```
krschacht, line 106: This is the definite but that I fixed. Since the action did not have a :prevent, clicking the link was triggering the anchor which was causing the message list to flicker. I'm not sure if it's because it was jumping to the top of the screen and then auto-scrolling back down, or if it's because it was causing the frame to reload. Regardless, I think :prevent is a smell we should avoid.

PR #126, opened 2024-02-12
Debug page morphing & add tests

krschacht:
* Upgrade to turbo-rails 2 (a stable version)
* Remove the turbo frame history hack and switch to `data-turbo-action="advance"` since it's now working, mostly
* Fix a bug where sometimes the composer input does not clear (and add a test)
* Fix some bugs with morphing not happening occurring when submitting a message & add tests to catch regressions
* Add some additional test helpers: `click_element`, `click_test`, `watch_page_for_morphing`, `assert_page_morphed`


app/views/shared/_two_columns.html.erb
```diff
@@ -119,7 +120,7 @@
               <%= icon "paper-clip", variant: :mini, size: 23, class: 'cursor-pointer text-zinc-600 hover:text-zinc-900 stroke-2 dark:text-gray-300 dark:hover:text-white', title: 'Attach', tooltip: :top %>
             </div>
 
-            <%= form_with(model: [@assistant, @new_message], data: { controller: 'new-message' }) do |form| %>
+            <%= form_with(model: [@assistant, @new_message], data: { controller: "new-message", turbo_frame: "_top" }) do |form| %>
```
krschacht, line 123: It's counter-intuitive but without targeting the `_top` frame then morphing won't happen. Instead, the form replaces the `turbo-frame id="conversation"` and it does not morph this replacement.

test/system/messages/sidebar_test.rb
```diff
@@ -39,9 +39,11 @@ class MessagesSidebarTest < ApplicationSystemTestCase
...
     assert_first_message conversations(:greeting).messages.sorted.first
 
-    page.go_back
-    assert_current_path new_assistant_message_path(assistant)
-    assert_selected_assistant assistant
+    # TODO: There is a bug with the latest turbo where the final back doesn't properly load from cache.
+    #
+    # page.go_back
+    # assert_current_path new_assistant_message_path(assistant)
+    # assert_selected_assistant assistant
```
krschacht, line 46: I ran into a bug with turbo rails 2 and the `data-turbo-action="advance"`. As soon as you navigate with this "advance" action, it starts handling history in a different way. The history works, except until you go back so far that you hit the first page in history which was *not* an `advance` page.

For now, we'll just accept this bug so I've commented out the test for it. But at some point we can circle back and confirm the library fixed it or implement a work-around.

app/views/shared/_two_columns.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="flex h-screen"
-    data-controller="click-toggle turbo-frame-history"
```
krschacht, line 2: This history hack stopped working with turbo-rails 2. But fortunately the `data-turbo-action="advance"` mostly works now in turbo-rails 2 so rather than continuing to support this hack, I've moved us to use `advance`.

app/javascript/controllers/new_message_controller.js
```diff
@@ -27,8 +27,10 @@ export default class extends Controller {
   }
 
   submitForm() {
-    if (this.cleanInputValue.length > 0)
-      this.element.submit()
+    if (this.cleanInputValue.length > 0) {
+      this.element.requestSubmit()
```
krschacht, line 31: We definitely need to be using requestSubmit, without this morphing doesn't work.

app/javascript/controllers/new_message_controller.js
```diff
@@ -27,8 +27,10 @@ export default class extends Controller {
...
+      this.inputTarget.value = ''
```
krschacht, line 32: Every so often I'm able to submit a message and the form input doesn't clear itself. I don't know why but to fix it I just always cleared it. I added a test for this as well.

robacarp (approved):
easy win, nice polish

PR #130, opened 2024-02-13
Add support for messages to include image attachments

krschacht:
Just like with ChatGPT, now you can attach images to the composer to ask questions about them:

image

It does not work for non-images yet. Most of the underlying data model work is ready to support non-images but I didn't bother to make it completely work until we move from the ChatCompletions API to the new AssistantsAPI since that's what will be required to ask questions about PDFs and other docs.


app/controllers/messages_controller.rb
```diff
@@ -24,7 +24,7 @@ def create
     @message = @assistant.messages.new(message_params)
 
     if @message.save
-      GetNextAiMessageJob.perform_later(@message.conversation.id)
+      GetNextAIMessageJob.perform_later(@message.conversation.id, @assistant.id)
```
krschacht, line 27: Deeper in this job, I need to know whether the assistant being used supports images or not (3.5 does not but GPT 4 does). I could have hacked it a different way but I decided to pass in the assistant from here to prepare for the day we let multiple assistants participate in the same conversation

app/javascript/controllers/image_upload_controller.js
```diff
@@ -1,56 +1,80 @@
 import { Controller } from "@hotwired/stimulus"
```
krschacht, line 1: I did a bunch of rework of this whole stimulus controller. It's probably easier to view it as the finished file rather than the diff

app/javascript/controllers/new_message_controller.js
```diff
@@ -29,7 +29,7 @@ export default class extends Controller {
   submitForm() {
     if (this.cleanInputValue.length > 0) {
       this.element.requestSubmit()
-      this.inputTarget.value = ''
+      this.inputTarget.reset()
```
krschacht, line 32: we want to make sure the file attachment field is also cleared

app/javascript/controllers/scrollable_controller.js
```diff
@@ -10,7 +10,7 @@ export default class extends Controller {
   checkScroll() {
     const target = this.scrollableTarget
     const isAtTop = target.scrollTop === 0
-    const isAtBottom = target.scrollHeight - target.scrollTop === target.clientHeight
+    const isAtBottom = Math.round(target.scrollHeight - target.scrollTop) === target.clientHeight
```
krschacht, line 13: Oddly, I have one conversation where the scrollHeight is `xxx.5` ! I've never run into fractional pixel heights but rounding solved it

app/jobs/get_next_a_i_message_job.rb
```diff
@@ -0,0 +1,29 @@
+class GetNextAIMessageJob < ApplicationJob
+  def perform(conversation_id, assistant_id)
+    puts "GetNextAIMessageJob.perform(#{conversation_id}, #{assistant_id})"
+
+    @conversation = Conversation.find conversation_id
+    @assistant = Assistant.find assistant_id
+
+    @new_message = @conversation.messages.create! role: :assistant, content_text: "", assistant: @conversation.assistant
+    @new_message.broadcast_append_to @conversation
```
krschacht, line 9: Sometimes the job is slow to be picked up. If we moved these two lines to be in the controller then we'd get an instant stub of the assistant response even if it takes awhile to process. But that was out of scope here so I didn't do yet.

app/models/document.rb
```diff
@@ -3,7 +3,51 @@ class Document < ApplicationRecord
...
+
+  private
+
+  def file_present
+    errors.add(:file, "must be attached") unless file.attached?
+  end
+
+  def set_default_user
+    self.user ||= message.conversation.user # this won't work when multiple people are in a conversation
+  end
```
krschacht, line 40: I did not want to have `user_id` be a hidden form field because then it could be a vulnerability where someone could create a message for another arbitrary user. I wasn't sure where it was best to set the user_id so I just pushed all the way down into here.

app/models/message.rb
```diff
@@ -19,13 +21,6 @@ class Message < ApplicationRecord
 
   after_create_commit :broadcast_message
 
-  def for_openai
-    {
-      role: role,
-      content: content_text
-    }
-  end
-
```
krschacht, line 28: Now that we support images, we need to consider the assistant being used to know how to properly render this so I moved it into the `open_a_i.rb` client

test/fixtures/documents.yml
```diff
@@ -8,8 +8,32 @@ background:
 
 cat_photo:
   user: keith
-  assistant: samantha
+  assistant:
```
krschacht, line 11: I am pretty sure that documents should be connected to the user -or- the assistant. I didn't do testing of the OpenAI API to be sure and the docs weren't totally clear so I just fixed it here in the fixtures. But if that is true then we should probably make this association polymoprhic.

app/javascript/controllers/image_upload_controller.js
```diff
@@ -1,56 +1,80 @@
...
 export default class extends Controller {
-  static targets = ["file", "content"]
+  static targets = [ "file", "content", "preview" ]
 
   connect() {
-    this.dropHandler()
-    this.pasteHandler()
+    this.fileTarget.addEventListener("change", this.boundPreviewUpdate)
+    this.contentTarget.addEventListener("dragover", this.boundIgnore)
+    this.contentTarget.addEventListener("dragenter", this.boundIgnore)
```
robacarp, line 9: I'm surprised that we even need to listen for these events if they're just going to be "prevented" -- but I really have very little experience with the drag and drop API. What happens when these aren't blocked?

app/javascript/controllers/image_upload_controller.js
```diff
@@ -1,56 +1,80 @@
...
+    const input = this.fileTarget
+    if (input.files && input.files[0]) {
+      const reader = new FileReader()
+      reader.onload = (e) => {
+        this.previewTarget.querySelector("img").src = e.target.result
+        this.element.classList.add("show-previews")
+        this.contentTarget.focus()
+        window.dispatchEvent(new Event('resize')) // Throw this event will cause textarea_autogrow to reprocess
+      }
+      reader.readAsDataURL(input.files[0])
```
robacarp, line 33: This is cool. I had no idea FileReader existed.

app/javascript/controllers/image_upload_controller.js
```diff
@@ -1,56 +1,80 @@
...
-            const dataURL = await this.readBlobAsDataURL(blob)
-            this.addImageToFileInput(dataURL, blob.type)
-          } catch (error) {
-            console.error("Error reading pasted image:", error)
-          }
+        try {
+          const dataURL = await this.readPastedBlobAsDataURL(blob)
+          this.addImageToFileInput(dataURL, blob.type)
+        } catch (error) {
+          console.error("Error reading pasted image:", error)
```
robacarp, line 71: Why not just let this error throw out to the console on it's own? Does catching here allow you to continue doing other things?

app/jobs/get_next_a_i_message_job.rb
robacarp: I had that thought at one point too, and I think it might be nice to do. We could also add a placeholder animation of some sort to "spin" while the job is sitting idle.

app/jobs/get_next_a_i_message_job.rb
robacarp, line 1: There is some way to teach the zeitwork inflector about acronymns, which I came across at one point. I didn't really feel like it was worth bothering with in that PR, but I didn't try to put underscores between the letters for some reason to get the acronym to be capitalized. 👍

app/models/document.rb
robacarp: Yeah this is fine for the current model.

app/services/a_i_backends/open_a_i.rb
```diff
@@ -0,0 +1,65 @@
...
+          content: content
+        }
+      else
+        {
+          role: message.role,
+          content: message.content_text
+        }
+      end
+    end
+  end
```
robacarp, line 64: This was true before, and is a little more true now. If this logic gets much more complicated, it might be worth putting in a presentation jbuilder or something to render messages apropriately without so many fiddly conditionals.

test/models/document_test.rb
```diff
@@ -6,15 +6,15 @@ class DocumentTest < ActiveSupport::TestCase
...
   end
 
   test "has an associated message" do
     assert_instance_of Message, documents(:cat_photo).message
   end
 
-  test "simple create works" do
-    assert_nothing_raised do
+  test "create fails without a file" do
+    assert_raises do
```
robacarp, line 17: I think assert_raises can take the expected Error class as a parameter -- adding that here will prevent false-passes on this test in the future if it starts throwing StandardError for some reason.

robacarp (approved):
nice work! I left a couple small comments

app/javascript/controllers/image_upload_controller.js
krschacht: Good call, this was old code from Ali that I didn't correct. I'll update this.

app/javascript/controllers/image_upload_controller.js
krschacht: ChatGPT taught me that one. :)

app/services/a_i_backends/open_a_i.rb
krschacht: Agreed, I'll skip for now but it's just about worth it.

test/models/document_test.rb
krschacht: Good call, fixing

app/javascript/controllers/image_upload_controller.js
krschacht: Good call. I didn't actually try it without them. I just did and only one is needed for one case so I've removed it and refactored the code with the single suppression that's needed.

PR #143, opened 2024-02-20
Simplify the view structure

krschacht:
I haven't been super happy with the way I did the include for the `two_columns` view file. And @robacarp you made a remark in the PR about: why not just have a left-column and right-column content_for.

Well, while cleaning up some other view files I kept pulling the thread of the sweater and ended up restructuring the overall messages view to follow this advice. It worked out nicely.

There are not any logic changes in here. The summary of what I did:

* Created a layout/messages file which all messages views use automatically
* This has a left_column and right_column yield
* Because messages/new and messages/index have identical left_columns, there is a separate left_column partial. Likewise, there is a separate right_columns partial.
* I went ahead and moved the assistant_sidebar_link and conversation_sidebar_link into _assistant and _conversation, so we get the nice syntax of `<%= render @conversations %>`. I've been thinking through how to implement editing of conversation titles and assistant titles and I think this small change will facilitate that.

While doing this refactor, I was very happy for our thorough test coverage because it helped me catch a couple subtle bugs I introduced. :)


krschacht:
@robacarp I'm going to go ahead and merge in, but just flagging you on this as an FYI so that you can at least read the PR description and know what happened.

app/views/layouts/messages.html.erb
robacarp, line 1: this is nice!

app/views/messages/_right_column.html.erb
```diff
@@ -0,0 +1,187 @@
+<turbo-frame  id="conversation"
```
robacarp, line 1: should this be `dom_id(@conversation)`?

krschacht: No, this is right. The sidebar links frame target="conversation".  Conceptually, the "conversation" frame is the title + messages + composer, all of which are tied to the conversation in some way. It's basically the full right-hand side.

Issue #108, opened 2024-02-07
When conversation is scrolled to last message, it doesn't react to resizing browser


krschacht:
When you resize the browser such that the last message is no longer visible in the viewport, it should either auto-scroll down or the scroll-down arrow should appear. I'm not sure which would be the ideal behavior. The risk of auto-scrolling down is that we'd only want to do that if you were definitely scrolled to the bottom of the page because you could have scrolled a specific message into the viewport for reference — you resize your browser — then we jump you down.

Ideal behavior would be: know if you're at the bottom of the page w/ a hidden scroll down arrow, if so then when the browser resizes we re-scroll you down.

Next best behavior: when the browser resizes, we just re-run our logic for determining whether to show the scroll-down arrow.

PR #144, opened 2024-02-20
Improve the logic for autoscrolling down

krschacht:
The main problem I was solving with this PR was that when ChatGPT is responding with a really long message, the browser keeps auto-scrolling to the bottom, but for long messages I find myself scrolling back up so I can start reading the beginning while the rest is streaming in.

I fixed this by adding some nuance to the autoscroll down logic. When you press Enter to submit a new question, it always scrolls down. And when the very first stub response comes back, it always scrolls down. But from then on, it only scrolls down if it was previously at the bottom. If you have explicitly scrolled further up into the conversation it won't keep scrolling you down as the conversation streams in.

I also had to throttle the number of Turbo Stream messages we are sending to the browser. It only sends one every 500 ms. This makes the streaming in a little less smooth, but this way when you scroll up the browser scroll event which fires has enough time to acknowledge you're no longer on the bottom so that when the next message piece streams in it doesn't try to scroll you down again.

While I was at it I fixed issue #108 — resizing the browser and expanding the composer both re-run the autoscroll down logic. I added some tests for these too.


app/javascript/controllers/image_upload_controller.js
```diff
@@ -34,7 +34,7 @@ export default class extends Controller {
     this.previewTarget.querySelector("img").src = ''
     this.element.classList.remove("show-previews")
     this.contentTarget.focus()
-    window.dispatchEvent(new Event('resize')) // Throw this event will cause textarea_autogrow to reprocess
+    window.dispatchEvent(new Event('resize'))
```
krschacht, line 37: I removed the comment because I added another listener for resize so it now does multiple things. It's not worth keeping the comment updated.

app/javascript/controllers/message_scroller_controller.js
```diff
@@ -0,0 +1,40 @@
...
+  static values = {
+    scrollDown: { type: Boolean, default: false },
+    onlyScrollDownIfNearBottom: { type: Boolean, default: false }
+  }
+
+  connect() {
+    if (window.lastMessageControllerInstance) {
+      window.lastMessageControllerInstance.disconnect()
+      window.lastMessageControllerInstance = this
+    }
```
krschacht, line 16: There will potentially be A LOT of these stimulus controller instances created. When it was only firing once, that wasn't a big deal. But now that we have an event listener I don't want them all going off so I added this global tracker to disconnect previous ones as a new one is added.

app/javascript/controllers/new_message_controller.js
```diff
@@ -1,5 +1,3 @@
-"use strict";
-
```
krschacht, line 2: Based on my reading, ES6 modules are always in strict mode so we shouldn't need this

app/javascript/controllers/textarea_autogrow_controller.js
```diff
@@ -23,7 +23,11 @@ export default class extends Controller {
...
   debouncedAutogrow = debounce(() => this.autogrow(), 50)
   autogrow() {
-    this.element.style.height = 'auto'
-    this.element.style.height = `${this.element.scrollHeight + 2}px` // the +2 is a hack to make the size not jump on load. The scrollHeight differs from than actual height for the empty state
+    const newHeight = `${this.element.scrollHeight + 2}px` // the +2 is a hack to make the size not jump on load. The scrollHeight differs from than actual height for the empty state
+    if (this.element.style.height && this.element.style.height != newHeight) {
+      this.element.style.height = 'auto'
+      this.element.style.height = newHeight
+      window.dispatchEvent(new Event('resize'))
+    }
```
krschacht, line 31: When I added this dispatchEvent I discovered this style.height was being set a lot more frequently than it needed to be so this extra logic only sets it when it changed

app/javascript/controllers/new_message_controller.js
krschacht, line 2: Based on my reading, this isn't needed in ES6 modules. Apparently they use strict by default

app/javascript/controllers/message_scroller_controller.js
robacarp: Interesting idea! I think you need to move line 19 out of the conditional?

app/javascript/controllers/message_scroller_controller.js
```diff
@@ -0,0 +1,55 @@
...
+  debouncedScrollDownIfWasBottom = debounce(() => this.scrollDownIfWasBottom(), 50, true)
+  scrollDownIfWasBottom() {
+    if (window.wasScrolledToBottom) this.debouncedScrollDown(false)
+  }
+
+  debouncedScrollDown = debounce((instant) => this.scrollDown(instant), 50, true)
+  scrollDown(instant) {
+    let instantScroll = instant ?? this.instantValue
+    setTimeout(() => {
+      window.wasScrolledToBottom = true // even if we don't get the full way, it was the intention
```
robacarp, line 47: should this flag be set outside of the timeout?

krschacht: This was a good catch. You had a couple good flags on logic related to scroll down, and notably, I wasn't totally happy with the scroll down logic. It still was jumpy at times and I still did not have an easy enough time scrolling **up** the page while a conversation was streaming in. I was really hoping that one of these things you pointed out was the root cause.

I fixed them, but unfortunately the issues didn't go away. :) But it led me down the right rabbit hole and I think I now have a root cause fix. I'm about to tee up a new PR.

PR #145, opened 2024-02-21
Stub out a new /settings controllers & views

krschacht:
My thinking is that we'll move all settings to a full screen experience rather than trying to make it all fit in a modal. Not only do I think this will be easier to code, but we'll probably end up with lots of settings in the long run so it'll be a better foundation to build up from.

This PR adds `settings/assistants/123` and `settings/user`. It just stubs out the actions & views that we'll need. It also adds contollere & system tests.

The two immediate things I want to get working (in a follow-up PR):

1. Moving the current Settings modal (first name, last name, and openAI Key) over to `/settings/user`
2. Make the `Settings` drop-down work for any Assistant in order to modify the custom instructions


app/controllers/settings/assistants_controller.rb
```diff
@@ -0,0 +1,43 @@
+class Settings::AssistantsController < ApplicationController
+  before_action :set_assistant, only: [:edit, :update, :destroy]
+
```
krschacht, line 3: We'll need all the CRUD actions in order to allow adding, editing, and removing assistants (i.e. GPTs)

app/controllers/settings/people_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      redirect_to edit_settings_person_path, notice: "Person was successfully updated.", status: :see_other
+    else
+      render :edit, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def person_params
+    params.require(:person).permit(:email, personable_attributes: [:id, :first_name, :last_name, :password, :openai_key])
```
krschacht, line 17: It took me awhile to make it work so that Person.update() would update the associated User. It turns out that passing in the :id field on User was the key to make it work. I think this opens up the possibility of submitting bad data to this endpoint to try and modify a user who is not you, so I added checks deep in the Person model to ensure that updates always use the proper id

app/services/a_i_backends/open_a_i.rb
```diff
@@ -34,14 +34,23 @@ def get_next_chat_message(&chunk_received_handler)
...
   private
 
+  def system_message
+    return [] if @assistant.instructions.blank?
+
+    [{
+      role: 'system',
+      content: @assistant.instructions
+    }]
+  end
```
robacarp, line 52: 💯

app/models/person.rb
```diff
@@ -1,11 +1,31 @@
...
+      errors.add(:personable_id, "cannot be changed after creation")
+    end
+  end
+
+  def proper_personable_id
+    if personable.id.blank?
+      errors.add(:personable_id, 'must be provided on update')
+    elsif personable.id != Current.person.personable_id && Current.person.personable_id.present?
+      errors.add(:personable_id, 'must match the current person')
+    end
```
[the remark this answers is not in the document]
krschacht: This is a good flag.

My other check above (personable_id_unchanged) ensures that the two exploits you mentioned don't happen, so really my extra check with Current.person is unnecessary. But I agree that it would be better to handle this at the controller. I'll update that in a quick PR.

PR #148, opened 2024-02-22
Update color scheme to latest version

florianfelsing:


florianfelsing:
Some more polishing to do here @krschacht, but feel free to take a glimpse whenever you feel like it.

krschacht:
@florianfelsing I just pulled it and toggled back-and-forth between chatgpt and hostedgpt, and it looks great! The only thing I notice is that our left-column is too wide. To be honest, I think I inadvertently changed this in a recent PR. Could you change our left-column to match the width of ChatGPT?

I'll take a look at your diff and see if I have any feedback on implementation.

app/views/messages/_right_column.html.erb
```diff
@@ -31,46 +31,42 @@
...
         <%= button_tag type: "button",
-              class: "block rounded-full bg-white dark:bg-graphite border border-gray-300 dark:border-zinc-600 inline-flex px-2 py-2 tooltip tooltip-top hover:tooltip-open cursor-pointer hover:bg-zinc-100 dark:hover:bg-zinc-600 group",
+              class: "block rounded-full bg-white dark:bg-gray-800 border border-gray-100 dark:border-gray-600 inline-flex px-2 py-2 tooltip tooltip-top hover:tooltip-open cursor-pointer group",
               data: {
                 tip: "Scroll down",
                 action: "scrollable#scrollToBottom"
               } do
         %>
-          <%= icon "chevron-down", variant: :outline, size: 19, class: 'text-gray-800 dark:text-zinc-400 cursor-pointer group-hover:text-black dark:group-hover:text-white stroke-2' %>
+          <%= icon "arrow-down", variant: :outline, size: 16, class: 'text-gray-800 dark:text-white cursor-pointer stroke-3' %>
```
krschacht, line 69: good catch

app/views/messages/_right_column.html.erb
```diff
@@ -31,46 +31,42 @@
...
 
   <!-- Only one of these headers is visible at a time, depending on width -->
-  <div id="wide-header" class="flex-none justify-between ml-2 mt-2 mb-3 h-10 text-black dark:text-white text-lg hidden md:flex">
+  <div id="wide-header" class="justify-between flex-none hidden h-10 mt-2 mb-3 ml-2 text-lg text-gray-950 dark:text-gray-100 md:flex">
     <%= render 'messages/right_column/assistant_header' %>
-
-    <%= link_to new_assistant_message_path(@assistant), class: "block cursor-pointer inline-flex mr-3 px-2 py-2 group", data: { role: "pencil" } do %>
-      <%= icon "pencil-square", variant: :outline, size: 22, class: 'stroke-2 text-zinc-600 group-hover:text-zinc-900 dark:text-zinc-400 dark:group-hover:text-white', title: "New chat", tooltip: :left %>
-    <% end %>
   </div>
```
krschacht, line 36: Let's keep the pencil in the top-right of the wide header. You were probably just comparing our design to chatgpt on the web. But annoyingly, the design of ChatGPT on mobile (phone & tablet) is slightly different than the web. On mobile, they have a persistent "new" button in the top-right.

For ChatGPT, they clearly have a separate codebase (and maybe separate team) doing the mobile stuff, so they've diverged. But when we get mobile working it'll be the exact same CSS. So over time we'll incorporate more of the mobile-specific stuff into our overall design so that we don't have to target those two environments too much.

app/views/assistants/_assistant.html.erb
```diff
@@ -44,13 +44,14 @@
...
       <%= link_to new_assistant_message_path(assistant), class: "inline-flex", data: { role: "pencil" } do %>
         <%= icon "pencil-square",
-          variant: :micro,
+          variant: :outline,
           size: 18,
-          class: "text-gray-400 cursor-pointer group-hover:visible hover:text-white invisible relationship:visible relationship:text-white",
+          class: "text-gray-950 dark:text-gray-100 cursor-pointer group-hover:visible invisible 
+                  relationship:visible relationship:text-gray-950 dark:relationship:text-gray-100",
           'data-controller': "nested-pointer",
           title: "New"
```
krschacht, line 52: Did you think the outline version looked better? At this small size, I thought the micro looked a bit nicer. I thought detail of the "outline" version got a bit lost at this size. But maybe you did it for consistency since, now that I'm looking at things, it looks like we might have outline icons everywhere else.

app/views/messages/_left_column.html.erb
```diff
@@ -1,45 +1,42 @@
...
-                        "
+  <div id="left-footer" class="flex-none pt-2 pb-4 pl-3 pr-5"
   >
-    <div class="dropdown dropdown-top dropdown-end w-full">
-      <div tabindex="0" role="button" class="cursor-pointer scale-down h-10 flex pr-2 rounded-lg hover:bg-zinc-400 hover:bg-opacity-25 justify-between items-center w-full group mb-1">
-        <div id="profile" class="ml-2 mr-2 w-7 h-7 bg-white rounded-full"></div>
-        <div class="flex-1 text-white truncate">
+    <div class="w-full dropdown dropdown-top dropdown-end">
+      <div tabindex="0" role="button" class="flex items-center justify-between w-full h-10 pr-2 mb-1 rounded-lg cursor-pointer scale-down hover:bg-gray-100 dark:hover:bg-gray-700 group">
+        <div id="profile" class="ml-2 mr-2 bg-green-600 rounded-full w-7 h-7"></div>
```
krschacht, line 29: Let's also make this circle a neutral color that doesn't draw the eye

krschacht (changes requested):
Great work @florianfelsing! I'm excited to have your help.

I left some small comments; the most notable being adding the top-right "pencil" icon back and those circle colors. Take a look.

app/views/layouts/messages.html.erb
```diff
@@ -11,10 +11,10 @@
       <%= yield :left_column %>
```
krschacht, line 11: Let's tweak the width of this column to match chatgpt (it's a few lines above)

florianfelsing:
Three things that come to mind that would still be open:

- The actual ChatGPT theme also makes use of gradients in different places, which gives their interface some special touch. Not sure if we also wanna handle these within the scope of this task, though
- The "Send message" should not be rounded, but have cornered edges. This was a little bit more tricky since I thought at first, though, because it seems like the tooltip is overwriting the icon's stroke width somewhere. Got a couple of hacky solutions in mind there, but nothing that I really liked so far
- ChatGPT applies blur to their sidebar when you hover over the "Close sidebar" button - would probably move that into a separate issue

app/views/assistants/_assistant.html.erb
krschacht: Let's keep the change you made!

krschacht: Let's keep the change you made! To be honest, I don't love the heroicon pack. The other pencil icon (the one for edit) is particularly weak at this small size. But it's a bigger project to try and switch that out and I think it's good enough to make a solid product. At some point we'll introduce a few alternate themes, and I think that's the time to revisit icons in a bigger way.

PR #154, opened 2024-02-23
Improve toast & turn click-toggle stimulus into generalized transition

krschacht:
I wanted to improve the styling of the toast, notably to make it disappear. I got a quick crash course of CSS transitions but it seems like I can't make it disappear solely with CSS. I tried to avoid javascript but I think we need it.

We use the click-toggle stimulus controller in a few places. It's been a useful one. However, this is a new case where I don't want a "click" to be what triggers a class toggle, instead I want to set a timeout which triggers the timeout. With this change, then I can use the same stimulus controller for the toast.

This PR:

* Restyles the toast
* Refactors click-toggle-controller into transition-controller

Screenshot of new toast (error one is a darker color to stand out more):

image


app/views/layouts/_toast.html.erb
```diff
@@ -1,31 +1,25 @@
...
+          "alert-info"
+        when "alert"
+          "alert-error"
+        else
+          "alert-info"
+        end
+      %>
+      <div class="alert <%= alert_class %>">
+        <span><%= text %></span>
+      </div>
```
krschacht, line 22: This toast component comes from the DaisyUI library. It slides in from the top-right of the screen and slides back out when it auto-dismisses.

app/javascript/controllers/transition_controller.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default class extends Controller {
+  static classes = [ "toggle" ]
+  static targets = [ "me" ]
+  static values = {
+    afterTimeout: Number
+  }
+
+  connect() {
+    console.log(`connect`)
```
robacarp, line 20: ❌ 

The other day I turned on the stimulus debug mode in application.js and it emits this message automatically. It would probably cover your logs on line 25/26 too. It's _verbose_ though because it turns on debug statements for every controller and action.

app/javascript/controllers/transition_controller.js
```diff
@@ -0,0 +1,35 @@
...
+// <div data-controller="transition" data-transition-toggle-class="!hidden">
+//   <div id="element-that-shows-and-hides" class="block" data-transition-target="me"></div>
+//   <a href="#" data-action="transition#toggleClass">click to toggle</a>
+// </div>
+//
+// Every element that is of target = me will have the class toggled.
+
+export default class extends Controller {
+  static classes = [ "toggle" ]
+  static targets = [ "me" ]
```
robacarp, line 14: Since this is the Transition controller, could this be "transitionable" or "transitioned"?

app/javascript/controllers/transition_controller.js
```diff
@@ -0,0 +1,35 @@
...
+  toggleClass() {
+    console.log('toggling', this.meTargets[0])
+    console.log('class', this.toggleClasses[0])
+    this.meTargets.forEach(element => {
+      element.classList.toggle(...this.toggleClasses)
+    })
+
+    // Showing and hiding elements can cause the page to flow differently, very similarly to what happens when the
+    // browser size changes. Throw this event in case we have other listeners on the resize event.
+    window.dispatchEvent(new Event('resize'))
```
robacarp, line 33: I see what you're saying, but do we have anything catching the resize event or is this preemptive optimization? I haven't had to listen for the resize event in ages because browser layout via CSS is so much more robust than what we used to do with dhtml layouts.

app/views/layouts/_toast.html.erb
```diff
@@ -1,31 +1,25 @@
...
-            &times;
-          <% end %>
-        </div>
-      <% end %>
+  <div class="toast toast-top toast-end transition-transform duration-100 ease-in"
+        data-controller="transition"
+        data-transition-toggle-class="translate-x-full"
+        data-transition-target="me"
+        data-transition-after-timeout-value="3000"
+  >
```
robacarp, line 8: I'm not sure what using the daisy-ui classes here benefits. With the tooltips, you get a lot for a little -- the interactivity and calculations about when and how a tooltip shows up are free and hidden.

Here, daisy is just aliasing ordinary CSS where "toast-top" really means "fixed top-4" and so on. I'm particularly surprised that the daisy component doesn't have a close button, or automatically transition out. That would make it feel worth it to me.

app/views/layouts/_toast.html.erb
```diff
@@ -1,31 +1,25 @@
...
-          %>
-            &times;
-          <% end %>
-        </div>
-      <% end %>
+  <div class="toast toast-top toast-end transition-transform duration-100 ease-in"
+        data-controller="transition"
+        data-transition-toggle-class="translate-x-full"
+        data-transition-target="me"
+        data-transition-after-timeout-value="3000"
```
robacarp, line 7: There's a situation here where a user mouse-over's a toast alert, and then the alert fades out anyway. Since it's often true that users mouse follows their eyes, should this cancel the timeout when they mouse enter?

I also think it's helpful to provide an explicit close button. In the user studies I've witnessed about this stuff, people are accepting when something fades away on it's own (as long as they're not reading it). But if they notice something and want it to go away but don't see the option, it's frustrating and distracting.

app/javascript/controllers/transition_controller.js
krschacht: Ah yes, good idea

app/javascript/controllers/transition_controller.js
krschacht: Yes, right now the scroll position of the page is recalculated on browser resize. This ensures that (a) if I'm at the bottom of the page and I resize my browser, I stay at the bottom of the page and (b) if I'm scrolled up reading a specific message and I resize my browser, it keeps me at that message.

Currently, there are multiple circumstances **besides** browser resizing in which the browser height and thus the scroll position need to be recalculated. Right now I'm enabling those other stimulus controllers to "communicate" with the scroll position controller by simply throwing a resize event.

This particular one right here is because hiding the left-hand column needs to trigger a scroll position recalculate. Basically, even though the browser itself didn't change size the right may have just done so. The other places I'm dispatching this event are within composer since that thing grows & shrinks in size.

It's a little bit of an overload of the resize event. Actually, sheesh, I should just create a custom event for this. I'll do that in a follow-up PR.

app/views/layouts/_toast.html.erb
krschacht: I think you're right, but I'm going to come back to that. I think there is a chance we re-think toasts further as part of larger design changes so I want to let things settle a bit more.

It probably would be nice to have an explicit X, but in this particular case it's actually hard to mouse over and click it before it disappears so I'll punt that to future consideration.

PR #155, opened 2024-02-23
Improve the styling of the settings pages

krschacht:
In PR #145 I moved the user settings out of a modal and to full-pages.

This new PR does another pass to improve the design of those pages. Most notable is that I extend the same two-column layout from the home experience into the /settings page as well. I elevate the HTML structure from layout/messages back into layout/application so that every place can use it.

This has the side effect of bringing the two-column design to the login and registration pages, which is a little bit off, but I think it's okay. We'll eventually want to add a bit more content to those pages to help explain the app and we can lean into that two-column layout.

One thing I didn't really tackle was improving the design & consistency of forms, overall. We're still mostly using rails scaffolding forms still.

Some screenshots:

image

image

image


app/controllers/settings/application_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  def set_settings_menu
+    # controller_name => array of items
+    @settings_menu = {
+      assistants: Current.user.assistants.ordered.map {
+                    |assistant| [ assistant, edit_settings_assistant_path(assistant) ]
+                  }.to_h,
+      people: { 'Account': edit_settings_person_path }
+    }
+  end
+end
```
krschacht, line 16: I haven't used this pattern before of a base controller for a set of routes, but it worked out quite nicely.

app/views/assistants/_assistant.html.erb
```diff
@@ -8,6 +8,7 @@
               flex justify-between items-center
               mb-1 p-1 pl-2 pr-2 mr-5
             hover:bg-gray-100 dark:hover:bg-gray-700
+            bg-gray-50 dark:bg-gray-900
```
krschacht, line 11: @florianfelsing One small thing I missed in your design change: content scrolls beneath the very top sticky assistant. This was an easy fix to add the background back:
image

app/controllers/settings/application_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+  private
+
+  def set_settings_menu
+    # controller_name => array of items
+    @settings_menu = {
+      assistants: Current.user.assistants.ordered.map {
+                    |assistant| [ assistant, edit_settings_assistant_path(assistant) ]
+                  }.to_h,
+      people: { 'Account': edit_settings_person_path }
+    }
```
krschacht, line 14: I haven't used this pattern before of a base controller for a set of routes, but it worked out quite nicely.

This particular `@settings_menu` felt a little kludgy to build it up like this, but it seems reasonable enough.

app/views/layouts/application.html.erb
```diff
@@ -14,8 +14,46 @@
     <%= turbo_refreshes_with method: :morph, scroll: :preserve %>
     <%= content_for :head %>
   </head>
```
krschacht, line 16: I deleted the layout/messages template and elevated most of that into layouts/application.

But there is still a messages/left_column and messages/right_column

app/controllers/settings/application_controller.rb
robacarp, line 1: Interesting idea, I've never seen that either. 👍

robacarp (approved):
it's nice to see some of the secondary pages taking shape

PR #156, opened 2024-02-27
Cleanup logic around event firing for auto scroll

krschacht:
I have not been completely satisfied with the scroll behavior. I went deep trying to track down the issue and ultimately realized it was related to a misunderstanding of how debounce works. The main thing this PR does is fix debounce and adding it's compliment: throttle. The tl;dr is that debounce() fires last of a series of duplicate calls whereas throttle() fires the first. I put more details in a comment at the top of each file.

The notable wins in this PR are:

* The initial scrolling down of the page on load happens much faster.
* As you type things into the composer textarea and it stretches in height, you don't notice a lag in the stretch.
* As the reply is streaming in, the scroll down happens much faster so you don't notice a lag between text appearing and scrolling.
* As the reply is streaming in, you can scroll up the page to read a different message without having to "fight" with the auto-scroller taking you back down.
* If you're scrolled all the way to the bottom and you resize the browser, you stay scrolled down.
* If you are scrolled up to a specific message and you resize the browser, you stay looking at that message.

There is quite a lot of small refactoring so I'm not sure how much it's worth reading the PR closely. I'll highlight the notable changes in logic with comments.

I am now confident that our scroll behavior will be more reliable than ChatGPTs! Their issues have long bugged me and after debugging all of ours I have a new appreciation for how tricky it is to get right. In fact, the desire to be better than them is partly what motivated me to track all these issues down. :)


app/javascript/controllers/message_scroller_controller.js
```diff
@@ -2,54 +2,60 @@ window.lastMessageControllerInstance = null
...
 
   connect() {
-    if (window.lastMessageControllerInstance) {
-      window.lastMessageControllerInstance.disconnect()
-      window.lastMessageControllerInstance = this
-    }
-    this.scrollableTarget = document.getElementById('right-content')
-    window.addEventListener('resize', this.debouncedScrollDownIfWasBottom) // resizing browser & composer size changes
+    if (window.lastMessageControllerInstance) window.lastMessageControllerInstance.disconnect()
+    window.lastMessageControllerInstance = this
```
krschacht, line 18: This was a mistakenly inside the conditional

app/javascript/controllers/message_scroller_controller.js
```diff
@@ -2,54 +2,60 @@ window.lastMessageControllerInstance = null
...
-    let instantScroll = instant ?? this.instantValue
-    setTimeout(() => {
-      window.wasScrolledToBottom = true // even if we don't get the full way, it was the intention
-
-      this.scrollableTarget.scrollTo({
-        top: this.scrollableTarget.scrollHeight,
-        behavior: instantScroll ? "auto" : "smooth"
-      })
-    }, instantScroll ? 300 : 0) // without the delay sometimes the page hasn't finished rendering and it doesn't go to the bottom
+  throttledScrollDown = throttle(() => this.scrollDown(), 50)
```
krschacht, line 44: Changing this from debounce() to throttle() was the biggest win. When it was debounce, it was delaying the execution by 50ms and that was causing most of the problems

app/javascript/controllers/message_scroller_controller.js
```diff
@@ -2,54 +2,60 @@ window.lastMessageControllerInstance = null
...
+      top: this.scrollableTarget.scrollHeight,
+      behavior: this.instantlyValue ? "auto" : "smooth"
+    })
+
+    if (this.instantlyValue) {
+      // This occurs immediately after page load; we jump to the bottom as fast as we can. However,
+      // sometimes this fires and jumps before the page is fully loaded so scrollHeight it calculates
+      // may not be correct yet. As a precaution, we add one more event to fire on window load to scroll
+      // down a bit further. This was hard to test so I'm not yet certain it solves the problem.
+      window.addEventListener('load', this.throttledScrollDownIfScrolledToBottom, { once: true })
```
krschacht, line 58: I eliminated the 300ms timeout in the case of an instant scroll down and instead added this second extra scroll onLoad

app/javascript/controllers/scrollable_controller.js
```diff
@@ -1,26 +1,25 @@
...
 
-  debouncedCheckScroll = debounce(this.checkScroll, 50)
-
+  throttledCheckScroll = throttle(() => this.checkScroll(), 50)
   checkScroll() {
     const target = this.scrollableTarget
     const isAtTop = target.scrollTop === 0
     const scrollOffset = target.scrollHeight - target.scrollTop
-    const isAtBottom = Math.abs(scrollOffset - target.clientHeight) <= 2 // occasionally these differ by a pixel or two
+    const isAtBottom = Math.abs(scrollOffset - target.clientHeight) <= 40 // occasionally these differ by a few pixels
```
krschacht, line 13: I'm certain I needed to change this after fixing the other debounce > throttle, but I think it makes the "sticking scroll to bottom" a bit more reliable

app/javascript/controllers/scrollable_controller.js
```diff
@@ -37,7 +36,7 @@ export default class extends Controller {
   }
 
   scrolled() {
-    this.debouncedCheckScroll()
+    this.throttledCheckScroll()
```
krschacht, line 39: This was causing the user-initiated scroll to be delayed and it was never catching the final end-of-scroll event, because debounce throws that one away.

app/javascript/controllers/textarea_autogrow_controller.js
```diff
@@ -12,20 +12,21 @@ export default class extends Controller {
...
 
   disconnect() {
-    this.element.removeEventListener('input', this.debouncedAutogrow)
-    window.removeEventListener('resize', this.debouncedAutogrow)
+    this.element.removeEventListener('input', this.throttledAutogrow)
+    window.removeEventListener('resize', this.throttledAutogrow)
   }
 
-  debouncedAutogrow = debounce(() => this.autogrow(), 50)
+  throttledAutogrow = throttle(() => this.autogrow(), 50)
```
krschacht, line 24: This was causing the slight delay when auto-growing the textarea

app/jobs/get_next_a_i_message_job.rb
```diff
@@ -25,12 +25,16 @@ def perform(conversation_id, assistant_id)
...
     puts "\nFinished GetNextAIMessageJob.perform(#{conversation_id}, #{assistant_id})" if Rails.env.development?
 
   rescue => e
     puts "Error in GetNextAIMessageJob: #{e.inspect}"
     puts e.backtrace
   end
+
+  def self.broadcast_updated_message(message)
+    message.broadcast_replace_to message.conversation, locals: { only_scroll_down_if_was_bottom: true }
+  end
```
krschacht, line 39: Moved this into a static method so tests can easily execute it. I caught a bug that tests were not excercising because of this so now we test for conversation parts streaming in.

test/application_system_test_case.rb
```diff
@@ -34,11 +34,17 @@ def assert_active(selector, error_msg = nil, wait: nil)
     assert_equal element, page.active_element, "Expected #{selector} to be the active element, but it is not. #{error_msg}"
```
krschacht, line 34: Everything from here downward is: (a) adding new tests, (b) addressing some flappy test cases, and (c) creating some new test helpers to simplify a bunch of testing of scroll position

PR #159, opened 2024-02-27
Add confirmation before deleting assistant

krschacht:
I accidentally deleted my assistant. :)

@robacarp I just discovered that when I added `data: { confirm: "Are you sure?" }` to the form button and it didn't work. I tracked it down to this removal of UJS. ChatGPT is telling me that there is some baseline rails functionality such as this delete confirmation that are still handled by UJS and that it can coexist with stimulus, so this PR adds it back. Any thoughts?

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,3 +1,5 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "@hotwired/turbo-rails"
 import "controllers"
+import Rails from "@rails/ujs"
+Rails.start()
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,5 +1,3 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "@hotwired/turbo-rails"
 import "controllers"
-import Rails from "@rails/ujs"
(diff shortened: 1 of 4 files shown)


app/javascript/application.js
```diff
@@ -1,3 +1,5 @@
 // Configure your import map in config/importmap.rb. Read more: https://github.com/rails/importmap-rails
 import "@hotwired/turbo-rails"
 import "controllers"
+import Rails from "@rails/ujs"
+Rails.start()
```
robacarp, line 5: I mentioned this elsewhere in a comment, but for the sake of it being here where the context is helpful, I have no real objection to this. If the only thing we're hoping for out of UJS is data-confirm, I might opt to write a stimulus controller for that.

krschacht: Yea, that’s true, it would be easy. I’ve been confused about UJS but I just did enough reading to finally understand — hotwire really is intended as a full replacement for it. In fact, it looks like they’ve added:

data-turbo-confirm=“are you sure?”
https://github.com/hotwired/turbo/pull/379

and even: data-turbo-method=“delete”

I’ll do the proper fix for this.

robacarp: Oh good find, I wasn't aware of that but it sure seemed like UJS had gone stale.

PR #164, opened 2024-02-29
Eliminate extra broadcasts to fix flicker

krschacht:
I kept noticing an issue with the assistant reply as it was streaming in. It looked like sometimes a word or two would appear and then disappear and then more of the text would appear. But it was hard to precisely identify what the issue was because it happened really fast and any time I would increase the delay between messages (within the get_ai job) then the issue would go away. There was some kind of race condition.

I added so me listening for the turbo render event and watched the logging closely and I was definitely seeing turbo stream actions hitting the client out of order. The clearest issue was around the initial messages. Sometimes broadcast `replace` actions would be processed by the client before the broadcast `append` had happened. Once it finally got started with the message, I think some of the later `replace` actions were sometimes received out of order but I'm not certain of this.

I also noticed some extra refresh actions that were being sent. This PR fixes some clear issues and it adds logging to help me potentially identify future issues. Specifically:

1. The initial assistant reply is trigger by an after_create callback hook within `message.rb`. This ensures it appears on the screen before the job is even picked up to generate the AI response.
2. Message creation no longer has a default `touch` on the belongs_to conversation. Instead, I added a single explicit touch trigger from within the get_ai job when it's wrapping up. This eliminates some extra refreshes that we don't need.

The tests were updated to reflect these changes.

There still may be an issue with the get_ai job replace actions being processed out of order, but I couldn't replicate to be certain. I added some short-term logging in application.js. This way if I ever think I see a word disappear during streaming then I can open console and inspect the log that's outputted there to know for sure. If there is, I started working on a fix but it's a bit hacky so I don't want to add it since my changes in this PR may have fixed the issue.


app/jobs/get_next_a_i_message_job.rb
```diff
@@ -1,40 +1,58 @@
...
+          GetNextAIMessageJob.broadcast_updated_message(@message)
+          last_sent_at = Time.current
+        end
 
-        if Time.current.to_f - last_sent_at >= 0.1
-          GetNextAIMessageJob.broadcast_updated_message(@new_message)
-          last_sent_at = Time.current.to_f
+        if @conversation.latest_message != @message
+          raise ResponseAborted
         end
```
krschacht, line 26: @robacarp You're right, it is better to pass in message_id. I refactored to make that change. However, I had a good reason to keep conversation, just a bad implementation. :)

I had added this extra check to make sure the last message in the conversation had not changed. Check at how I've updated this above ^.

What I'm attempting to do is two-fold:

1) If there's any evidence the conversation has moved on from this point (i.e. `user_has_replied_to_chat_since_queuing`) then we just skip this job. This is now implemented in a better way.

2) I would also like to abort the incoming streaming conversation if the user replies. Sometimes chatgpt will give me a long response and I butt in and say something different to take it in a new direction. This should be detected and should stop streaming.

This line 26 right here is how I've attempted to implement it, but it doesn't work. I added all sorts of debugging here including changing this line to `@conversation.messages.ordered.reload.last` and it seems that while this job is processing, if another Message gets created in a different thread, this job will not be able to detect that new message! I'm baffled as to why. But if you look at this debug output you'll see what I mean:

Notice that it's streaming a reply for message `1048910246` and then message `1048910248` and a separate job starts streaming the reply to that but **the first job still thinks that xx46 is the latest message**. Any ideas what's going on?

```
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    |  will
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    | ow
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    | 's
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    |  branches
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    |  low
worker-1    | ### Continue? 1048910246 == 1048910246
base-1      | Turbo::StreamsChannel transmitting "<turbo-stream action=\"replace\" target=\"message_1048910248\"><template>\n\n<div\n  id=\"message_1048910248\"\n  class=\"mb-5 ml-9 sm:ml-3 md:ml-8 group\"\n  data-role=\"message\"\n  data-timestamp=\"1709328512903\"\n  data-controller=\"message-scroller\"\n  data-message-scroller-scroll-down-va... (via streamed from Z2lkOi8vaG9zdGVkLWdwdC9Db252ZXJzYXRpb24vNDY0ODc0NDE1)
worker-1    |  help
worker-1    | ### Continue? 1048910248 == 1048910248
worker-1    | .
worker-1    | ### Continue? 1048910248 == 1048910248
worker-1    |  Just
worker-1    | ### Continue? 1048910248 == 1048910248
worker-1    |  let
worker-1    | ### Continue? 1048910248 == 1048910248
base-1      | Turbo::StreamsChannel transmitting "<turbo-stream action=\"replace\" target=\"message_1048910248\"><template>\n\n<div\n  id=\"message_1048910248\"\n  class=\"mb-5 ml-9 sm:ml-3 md:ml-8 group\"\n  data-role=\"message\"\n  data-timestamp=\"1709328513061\"\n  data-controller=\"message-scroller\"\n  data-message-scroller-scroll-down-va... (via streamed from Z2lkOi8vaG9zdGVkLWdwdC9Db252ZXJzYXRpb24vNDY0ODc0NDE1)
worker-1    |  me
worker-1    | ### Continue? 1048910248 == 1048910248
worker-1    |  know
worker-1    | ### Continue? 1048910248 == 1048910248
worker-1    | ,
base-1      | Turbo::StreamsChannel transmitting "<turbo-stream action=\"replace\" target=\"message_1048910246\"><template>\n\n<div\n  id=\"message_1048910246\"\n  class=\"mb-5 ml-9 sm:ml-3 md:ml-8 group\"\n  data-role=\"message\"\n  data-timestamp=\"1709328513145\"\n  data-controller=\"message-scroller\"\n  data-message-scroller-scroll-down-va... (via streamed from Z2lkOi8vaG9zdGVkLWdwdC9Db252ZXJzYXRpb24vNDY0ODc0NDE1)
worker-1    | 
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    | D
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    | ipping
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    |  in
worker-1    | !### Continue? 1048910246 == 1048910246
base-1      | Turbo::StreamsChannel transmitting "<turbo-stream action=\"replace\" target=\"message_1048910248\"><template>\n\n<div\n  id=\"message_1048910248\"\n  class=\"mb-5 ml-9 sm:ml-3 md:ml-8 group\"\n  data-role=\"message\"\n  data-timestamp=\"1709328513168\"\n  data-controller=\"message-scroller\"\n  data-message-scroller-scroll-down-va... (via streamed from Z2lkOi8vaG9zdGVkLWdwdC9Db252ZXJzYXRpb24vNDY0ODc0NDE1)
worker-1    |  the
worker-1    | 
worker-1    | ### Continue? 1048910246 == 1048910246
worker-1    |  water### Continue? 1048910248 == 1048910248
```

robacarp: I assume that the query response to `conversation.latest_message` is cached.

This is actually pretty tricky to solve, because I'd be weary of simply changing it to `reload` the latest message each time you loop -- that would be a lot of queries. But even if you do successfully abort this loop iteration, you're not actually aborting the loop which calls this block -- that's openai gem code.

Since we have redis, I might reach for a redis key to make this happen. You can effectively read a redis key each time this block is executed without issue. It's still polling, but I don't know if there's a way around that. We could listen on a pubsub channel (perhaps even the one turbo uses?) but that still doesn't give this a way to abort the fiber when it happens.

It's possible that the ruby openai code itself isn't going to allow you to modify this -- though it is seems like a scenario that could be solved with a [throw and catch](https://stackoverflow.com/questions/3716801/what-is-catch-and-throw-used-for-in-ruby), if you could figure out where to catch from.

krschacht: Nice, good idea. I'm going to do that in a followup PR. I think redis keys are the way to go. 

I don't think it was cached query because I tried adding a reload, but i'm okay not figuring out why this was happening and just using redis since that is a better long-term solution anyway.

Implementing this will lay the foundation for adding an explicit cancel buton to the streaming.

I did some testing on try/catch and it appears to work when I throw an exception here. I added some additional logic inside the deeper block to make sure it passes this exception onward but once it does then I can catch it inside the job.

robacarp: To be overtly clear, ruby has two mechanisms here -- `raise` and `rescue` (Exceptions) and `throw` and `catch` [(a kind of signals)](https://docs.ruby-lang.org/en/master/Kernel.html#method-i-catch). The difference between the two is perhaps semantics, but probably healthy to consider a signal here instead of an exception, if for no other reason than that you can explicitly handle Exception (ie, errors) in a different way.

PR #165, opened 2024-02-29
Add keyboard shortcuts for new chat

florianfelsing:


app/javascript/controllers/new_conversation_controller.js
```diff
@@ -0,0 +1,12 @@
...
+export default class extends Controller {
+  static values = {
+    path: String
+  }
+
+  navigate() {
+    const new_conversation_path = this.pathValue
+    window.location.href = new_conversation_path
+  }
+}
```
krschacht, line 12: Two thoughts on this:

1) I think we could do this as a generic controller. Imagine we have a different button or link next week that we want to bind a keyboard controller to? It’d be nice if we don’t have to create a new controller. Maybe this is keyboardable_controller.

If the action is: ->keyboardable#trigger

then the trigger action could determine if it’s a button and simply click it, or if it’s a link it navigates it

2) I could be wrong but I think there is a special new navigate function that we need to use rather than setting location.href. I believe Turbo.visit(url, “advance”) or something? Advance is surely the default so maybe you don’t need that

test/system/conversations_test.rb
```diff
@@ -3,11 +3,40 @@
...
+  end
+
+  test "creating new chat with meta+j" do
+    visit conversation_messages_path(@conversation)
+
+    page.driver.browser.action.key_down(:meta)
+                              .send_keys("j")
+                              .key_up(:meta)
+                              .perform
+
```
krschacht, line 37: There is a test helper method for sending keystrokes so you don’t need to grab the page object. I believe send_keys? Search for that

I use it for the other keyboard short cuts which all have tests. It lets you use the same naming as the stimulus key names

krschacht (changes requested):
A couple improvements

app/javascript/controllers/new_conversation_controller.js
florianfelsing: Yes, I think (1) makes a lot of sense. Working on this now.

And you are right about (2). This is the related piece of documentation: https://turbo.hotwired.dev/reference/drive, where it also states:

_If action is unspecified, Turbo Drive assumes a value of "advance"._

So we can indeed omit the "advance".

florianfelsing:
Thanks for your feedback @krschacht :)

krschacht (approved):
This looks great, I’ll merge in now. It’s a nice win. I’ve been wanting a shortcut for new conversations. :)

PR #176, opened 2024-03-04
Bind windows keyboard shortcuts

### LINKED ISSUE #175: Add windows shortcuts

krschacht:
Fixes #175 

Two improvements:

1. Adds Ctrl+ for all the keyboard shortcuts since this is standard for Windows users
2. Makes the `Cmd/Ctrl + J` keyboard shortcut open a new conversation for the **current** assistant rather than the first assistant. This is nice if you have multiple assistants you use.


app/javascript/controllers/keyboardable_controller.js
```diff
@@ -7,6 +7,6 @@ export default class extends Controller {
 
   navigate() {
     const new_conversation_path = this.pathValue
-    Turbo.visit(new_conversation_path) 
+    Turbo.visit(new_conversation_path)
```
krschacht, line 10: @florianfelsing I didn't mean to make this change, but my code editor strips trailing whitespace

krschacht:
@florianfelsing This PR was pretty straightforward, but take a quick look. I don't have a windows machine to test on but I just tried using Ctrl on my mac and they all worked, so it should be fine.

I discovered one notable bug while testing this: if you collapse the left sidebar then, as chat it re-opens. I'm not sure what's causing it. I'll create an issue now.

florianfelsing:
If you want I can pull down the branch and test it on my machine later @krschacht.

It's nothing critical, though, and the chance that this will not work is absolutely minimal. So I think we can also just merge and I will report back in the unlikely case that something is not working as expected.

app/javascript/controllers/keyboardable_controller.js
florianfelsing: Mine, too, usually. I probably messed something up in my JS auto-formatting config when I made some modification last week. Thanks for pointing this out!

PR #180, opened 2024-03-05
Re-requesting messages and other streaming improvements

### LINKED ISSUE #173: Get the re-generate response icon working beneath each assistant reply

krschacht:
* Makes the re-generate button work beneath assistant responses (fixes #173)
* Show a blinking cursor while the client is waiting on the server
* Fix an issue that was still causing some flicker at the beginning of generation

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -20,7 +20,6 @@ class Message < ApplicationRecord
   scope :ordered, -> { order(:created_at) }
 
   after_create :start_assistant_reply, if: -> { user? }
-  after_create_commit :broadcast_message
 
   private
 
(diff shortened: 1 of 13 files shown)


app/models/message.rb
```diff
@@ -39,8 +38,4 @@ def validate_conversation_user
   def start_assistant_reply
     conversation.messages.create! role: :assistant, content_text: "", assistant: conversation.assistant
   end
-
-  def broadcast_message
-    broadcast_append_to conversation, partial: "messages/message", locals: { scroll_down: true, timestamp: (Time.current.to_f*1000).to_i }
-  end
```
krschacht, line 45: @robacarp Notably, I was still seeing some flickering occasionally at the beginning of message generation. I finally tracked it down to this.

After you submit the form with a new chat message, these initial messages appear in the body with the full page render / morphing. However, we were also broadcasting them and so there was a race condition where sometimes the broadcast would happen before the morph had completed so when the morph ran it would roll back a moment in time. So we actually only need the broadcasting for updating the assistant's reply (within the job).

PR #182, opened 2024-03-07
Fix typo in README: converstions -> conversations

yasulab:
Sorry for the tiny pull request! But I found it when I am digging into this exciting project. 👀✨


robacarp (approved):
Awesome thank you @yasulab !

PR #185, opened 2024-03-09
Add conversation starters

florianfelsing:


florianfelsing:
@krschacht 

The layout has been implemented, but the conversation starters aren't functional yet.

These two points remain open at this stage:

(1) Submission of conversation starter messages
(2) We'll want to make the conversation starters dynamic and random. Right now I have just hardcoded four of them. I think we wanna define at least a dozen or so conversation starters of which we choose 4 random ones each time a new conversation is initiated. I'll leave it up to you whether we wanna cover this within the scope of 0.6.

Would you have some guidance/suggested path for implementation for (1) and some thoughts on (2)?

Thanks!

krschacht:
@florianfelsing On (1) you should be able to submit to the same endpoint that the current "composer" form submits to. Search for `button_to` in the codebase and you'll see an example of a one-liner that does a non-get to an endpoint. I bet you can include some params as part of that and change the method to POST/PUT. Worse case you'll need to use a form tag and construct a whole form with hidden elements, but I bet the button_to should work.

One thing I'm still figuring out is that it used to be: `method: :update` (or whatever action) but I saw mention that it changed to `turbo_method: :update` since the old method was for UJS but now we're using Turbo. However, in our codebase it looks like we're still using the old method so I can't quite explain that. I'm still figuring out the nuances of this.

krschacht:
On (2) we might as well do that one now. It should be as simple as an array
in the controller that you just  shuffle and grab the first four. Something
like that.

florianfelsing:
Ready for review @krschacht!

krschacht:
@florianfelsing some small comments in code organization but core behavior and styling looks great. Oh, but one thing I didn’t try was how this looks on mobile portrait. Did you consider that responsive case?

florianfelsing:
Thanks for the feedback @krschacht - I've addressed these issues now.

As for mobile portrait view, we are not copying ChatGPT there to 100%, but I think that our version looks better. I would re-assess in any case, though, once I've added our logo above the welcome message.

For your reference:

Screenshot from 2024-03-12 16-20-41

Screenshot from 2024-03-12 16-20-16

krschacht:
@florianfelsing nice work! And agreed, what you did for mobile looks great and better than ChatGPT.

PR #217, opened 2024-03-19
Add logo to project

florianfelsing:


florianfelsing:
There we go @krschacht :)

Please let me know if you would like to see the logo in some other place as well.

Also: Where would you like to put the ZIP file that contains all the source files that belong to the logo (AI files, Typo, ...)?

krschacht:
Can you attach the zip here as a comment? Or if github doesn't let you do
that, maybe you can throw it somewhere that gives you a URL and you can
post the URL here so I can download?

florianfelsing:
Sure!

[hosted_gpt_brand_files.zip](https://github.com/allyourbot/hostedgpt/files/14657651/hosted_gpt_brand_files.zip)

app/views/messages/new.html.erb
```diff
@@ -4,13 +4,18 @@
...
+        </div>
+        <span class="text-2xl font-semibold tracking-tighter">
+          How can I help you today?
+        </span>
+      </div>
     </div>
 
-    <div class="grid w-full grid-cols-1 gap-2 px-6 md:grid-cols-2">
+    <div class="z-10 grid w-full grid-cols-1 gap-2 px-6 bg-transparent md:grid-cols-2">
       <%= render partial: "messages/main_column/conversation_starter",
```
krschacht, line 19: @florianfelsing I just tried out this PR locally and it looks great! Thanks so much for taking the lead on all this. It's surprising how much a logo & the introduction of a brand color begin to up the professionalism of the app.

My one question: what's the introduction of the `z-10` and `z-0` do? I always get a little wary of introducing explicit z-index, but I know we do end up needing it sometime. I was curious and tried removing these in my dev environment and I didn't see a change in any appearance (i tried light / dark and different viewport sizes). Do you think we need the z's?

krschacht:
@florianfelsing One more thought: could you add the logo to the top of the README as well? That's the first thing most people will see when deciding to install so it'll make a nice first impression. I recently updated the demo video within that README but soon I'll have some more features to add to it and I'll add the logo into the demo video as well.

app/views/messages/new.html.erb
florianfelsing: You're right, we didn't really need them here in this case - removed them.

krschacht (approved):
Thanks for making these changes! sorry for being slow on the review, I’ve been traveling. I just took a look again and I’m merging in now.

PR #229, opened 2024-03-23
Fix "Converstion" typo

marckohlbrugge:


krschacht:
Oh jeez, good catch. Thanks for the fix @marckohlbrugge . I’m merging now.

Also, I would love to hear about your experience with the app! Is it working well? What missing features are highest on your wishlist? Start a discussion post and share a little about your thoughts so far. It really helps to keep us contributors all motivated and focused on the priorities.

marckohlbrugge:
Haven't had a chance yet to run the app, but looked through the code and looked really well done!

I once recreated the ChatGPT web interface myself, but your solution is more elegant 👍

krschacht:
@marckohlbrugge Thanks. I tried to make it really easy to get setup with Render so you can have the app deployed in 5 minutes. It's far from done but I now use it as my daily go-to instead of ChatGPT. I hope it can be that way for you, and if it's not yet there, let me know what things would close the gap so that it can be your go-to. That's how I've been prioritizing.

PR #240, opened 2024-03-25
Fix docker tmp/server.pid & docker DB connections

krschacht:
Two changes to docker:

1. Occasionally exits and it leaves a `tmp/pids/server.pid` file in the host's app/tmp directory. This causes problems when you restart docker. The solution to fix this was discussed in [this thread](https://github.com/allyourbot/hostedgpt/discussions/187#discussioncomment-8751255). We are blacklisting the host's tmp from being visible to the docker image. The docker's ephemeral tmp will disappear when docker exists. In my limited testing, this appears to have fixed it but I can't reliably repro the situation so I'm not certain.
2. Occasionally my docker worker will stop picking up jobs. I'll see `[SolidQueue] Claimed 1 jobs` and then after a few seconds I'll see an error about "all database connections are taken up." I'm guessing that puma may be using all the connections so I've changed puma to only use 2 threads, by default, within development. I hope this fixes but I can't reliably repro.

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -35,6 +35,7 @@ services:
     ports: ["3000:3000"]
     volumes:
       - .:/rails
+      - /rails/tmp
 
   worker:
     <<: *base
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -7,7 +7,9 @@
 # Any libraries that use thread pools should be configured to match
 # the maximum value specified for Puma. Default is set to 5 threads for minimum
 # and maximum; this matches the default thread size of Active Record.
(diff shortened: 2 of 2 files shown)


robacarp (commented):
I'm not 100% clear on the rails_max_threads -- is that because that value is reflected in the database connection pool size?

krschacht:
@robacarp Yes, that was my guess. I know in the Heroku world the DB had a max # of connections and puma threads along with workers consumed those processes. I’m really unclear why I occasionally get this error in the docker instance. I’m not certain the max_threads thing will fix it, but worst case, I figured it doesn’t hurt to only run two threads while developing.

PR #242, opened 2024-03-25
Format markdown (e.g. code) and add keyboard shortcut for copying

krschacht:
This PR started off just formatting code blocks, but it now formats all markdown (particularly code blocks) in both the user's messages and the AI's replies. Examples:

Single backticks `like this` and triple backticks:

```
like this
```

Are visually formatted. This is what they look like:

image

In addition, when you press `Cmd+Shift+C` it will copy the very last response, or if the last response includes a code block instead of copying the full last response it will copy the last code block within the last response.

In addition, there is a copy icon within each code blocks which can be clicked to copy to clipboard.

Also, as part of this PR the `keyboard-controller` was refactored to support a `target`. If a target is specified, that's the item which the `click()` is executed on. And if multiple targets exist, then `click()` is executed on the **last** of the targets. This was my simple way of determining which element to trigger *copy* on when the keyboard shortcut is invoked.


krschacht:
@olimart This PR was a little tricky. If you have any feedback on the implementation, I welcome it. I got it all working, I just have a couple automated tests to add.

olimart:
@krschacht will take a look (I had this one on my radar as well) 👍

app/helpers/messages_helper.rb
```diff
@@ -1,2 +1,71 @@
...
+  end
+
+  def paired_single_backticks(text)
+    text.gsub(/(?<!`)(`)(?=[^`]*`)([^`]+)`(?!`)/,
+      '<code><span class="hidden">`</span>\2<span class="hidden">`</span></code>')
+  end
+
+  def paired_triple_backticks(text)
+    text.gsub(/(?<!`)(```)([a-zA-Z0-9]*)(?:\n)?((?:\n?.*?)*)```(?!`)(?:\n)?/m) do
+      language = $2
```
krschacht, line 21: I just got a response from chatgpt that looks like this:

3. **Canopy by Hilton San Francisco SoMa** - A modern hotel with flexible meeting spaces that can be adjusted for any type of gathering. Its location in the SoMa district is also ideal for exploring the city during downtime.

It occurred to me that chatgpt just replies in markdown. Here I went and prioritize the code markdown but maybe I did it in such a way that will lead me to generate gsub for all the key markdown tags. That won’t be too hard, but if I would have stepped back and just said: how do I parse the whole response as markdown, it may have led me to a different solution. I’m inclined to ship what I did, anyway, but before starting to parse other tags I should investigate ruby markdown parsers.

olimart: What does Claude return?
Maybe check what Brakeman has to say about these regex (which will become a default gem in Rails 8)

Since you have a copy of Campfire, maybe check how they handle it. I know they're loading a bunch of language files.

Else, may be enough for now but leveraging some parsers look like a more robust approach on the long run.
Here is an example https://www.driftingruby.com/episodes/streaming-llm-responses

krschacht: This episode was a helpful link. It led me down a rabbit hole and ultimately to Redcarpet, which is a nice markdown parser that's pretty robust. I switched our my regex approach to use this. The bonus is *all* markdown now gets parsed. It's a nice win.

app/views/messages/_message.html.erb
```diff
@@ -46,8 +46,19 @@
...
+        class="prose break-words whitespace-pre-wrap leading-normal"
+      ><%= formatted_text = format(message.content_text) %><div
+        class="
+          animate-breathe
+          w-3 h-3
+          rounded-full
+          bg-black dark:bg-white
+          inline-block
+          ml-1 <%= !thinking && 'hidden' %>"></div
+      ></div>
```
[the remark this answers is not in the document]
krschacht: I had the messy formatting because this div had whitespace-pre-wrap so all line breaks within it were being rendered to the user. However, now that all markdown is converted to HTML I no longer need the pre-wrap. I've cleaned this up.

krschacht:
@olimart Thanks for the feedback on this. I cleaned up the code a bit more, got all markdown parsing done, and am merging this in!

PR #250, opened 2024-03-27
Turbo was executing the logout_path when the mouse entered the logout…

Tabenkin:
Turbo anticipates user actions by calling urls when the mouse enters the element. For the Logout button, Turbo was calling the logout_path and terminating the user's session.

Adding data: { turbo: false } to the button disables this behaviour on the logout button.


krschacht (approved):
👍 thanks for this fix! I'll merge in now

PR #262, opened 2024-04-05
Fix image upload race: wait for before display & processing

krschacht:
The image takes a couple seconds to process the different size variants (within documents.rb) yet it's trying to render the view and run the AI job right away. This PR adds some intelligence to it:

* Have the `GetNextAIMessageJob` blocks until the image is finished processing before sending to the AI's API
* When the conversation is rendering the list of messages, do not block if the image is not finished processing, instead insert a redirect URL and retry a few times until it properly loads. Show a spinner while we wait.


app/views/messages/_main_column.html.erb
```diff
@@ -142,7 +142,7 @@
           %>
         <% end %>
 
-        <div id="document-previews" class="absolute hidden space-x-3 left-7 top-3 show-previews:preview-container" data-image-upload-target="preview">
+        <div id="document-previews" class="absolute hidden space-x-3 left-9 top-3 show-previews:preview-container" data-image-upload-target="preview">
```
krschacht, line 145: the image attachments were a bit too close to the left edge

PR #269, opened 2024-04-12
Fix more flappy system tests

krschacht:


app/views/messages/_message.html.erb
```diff
@@ -162,37 +162,37 @@
...
+              <div data-role="image-loader" class="hidden p-5" data-image-loader-target="loader">
+                <%= spinner size: 6, class: "text-black dark:text-white" %>
+              </div>
+            </article>
+          </main>
+          <form method="dialog" class="modal-backdrop no-focus-outline">
+            <button id="modal-backdrop">close</button>
+          </form>
+        </dialog>
+      <% end %>
```
krschacht, line 195: The modal was inside of the message footer div which shows & hides on hover. I think that's why it sometimes was failing the `modal.visible?` check

PR #273, opened 2024-04-14
Add support for versioned messages within the Message model

krschacht:
This PR is my attempt to isolate all the backend logic changes to begin supporting versioned messages. All of the messages in a conversation now have:

* `index` attribute and indicates what number message in the sequence
* `version` indicates which message version this is for a given index
* `versions` is a dynamic attribute that lets you know all of the versions for a message, at this given index, assuming this index is a point where the conversation branched (i.e. `branched` = true) 

With those model changes, a bunch of query logic was updated so that it's pretty easy to create messages and not have to think too hard about what the index and version are.

An old conversation looks like this:

```
0.1
1.1
2.1
3.1
```
The first number is the message index, the second number after the decimal is the version of that message. In this start case it's the first version of every message. Now, message index 2 is going to be edited creating a new version and the assistant will reply. Now the messages for this conversation look like this:

```
0.1
1.1
2.1 2.2
3.1 3.2
```

The messages at `index` of 2 both have `branched` = true and if you check the `versions` attribute for those you'll get `[1,2]`. Even though there are two versions of message index 3 it was not branched there so versions will be an empty array. In other words, message 2 is where the UI is going to let you do next/previous and not at message 3.

By default, when you do `conversation.messages.for_conversation_version(1)` you will get the original message sequence back. But when you do `...(2)` or the shorthand `conversation.messages.latest_version_for_conversation` you get the following message sequence back:

```
0.1
1.1
2.2
3.2
```

And it supports more complicated branching such as this. In this example, the user went back to message 1 and branched again from there so there are now 3 different versions of this conversation and the message at index 0 is the only message that is shared between all three versions.

```
0.1        
1.1       1.3
2.1  2.2  2.3
3.1. 3.2
```

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -100,8 +100,8 @@ class MessagesController < ApplicationController
       :branched_from_version,
       documents_attributes: [:file]
     )
-    if modified_params.has_key?(:content_text) && modified_params[:content_text].blank? # when we rerequest the content_text: nil is mistakenly getting converted to ""
-      modified_params[:content_text] = nil
+    if modified_params.has_key?(:content_text) && modified_params[:content_text].blank?
+      modified_params[:content_text] = nil # nil and "" have different meanings
     end
     modified_params
   end
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -11,14 +11,6 @@ class MessagesController < ApplicationController
   before_action :set_conversation_starters, only: [:new]
 
   def index
-    if @version.blank?
-      version = @conversation.messages.order(:created_at).last&.version
-      redirect_to conversation_messages_path(
-        params[:conversation_id],
-        version: version
-      ) if version
-    end
-
     @messages = @conversation.messages.for_conversation_version(@version)
     @new_message = @assistant.messages.new(conversation: @conversation)
     @streaming_message = Message.where(
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -2,6 +2,7 @@ class MessagesController < ApplicationController
   include ActiveStorage::SetCurrent
   include HasConversationStarter
 
+  before_action :set_version, only: [:index, :update]
   before_action :set_conversation, only: [:index]
   before_action :set_assistant, only: [:index, :new, :create]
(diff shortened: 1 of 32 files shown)


app/controllers/messages_controller.rb
```diff
@@ -31,8 +32,11 @@ def edit
   def create
     @message = @assistant.messages.new(message_params)
 
-    if @message.save
-      GetNextAIMessageJob.perform_later(@message.conversation.latest_message.id, @assistant.id)
+    if @message.save # remember, an extra assistant message is auto-created
+      GetNextAIMessageJob.perform_later(
```
krschacht, line 36: kill this comment

app/controllers/messages_controller.rb
```diff
@@ -46,7 +50,7 @@ def create
 
   def update
     if @message.update(message_params)
-      GetNextAIMessageJob.perform_later(@message.id, @message.assistant.id)
+      GetNextAIMessageJob.perform_later(@message.id, @message.assistant.id) if @message.previous_changes.any?
```
krschacht, line 53: remove this job altogether since we no longer support editing messages

app/controllers/messages_controller.rb
```diff
@@ -59,16 +63,19 @@ def destroy
...
+  end
+
   def set_conversation
     @conversation = Current.user.conversations.find(params[:conversation_id])
   end
 
   def set_assistant
     @assistant = Current.user.assistants.find_by(id: params[:assistant_id])
-    @assistant ||= @conversation.messages.ordered.last.assistant
+    @assistant ||= @conversation.messages.for_conversation_version(@version).last.assistant
```
krschacht, line 78: use the conversation helper

app/controllers/messages_controller.rb
```diff
@@ -88,11 +95,15 @@ def message_params
...
+      :version,
+      :role,
       :cancelled_at,
-      :rerequested_at,
+      :branched,
+      :branched_from_version,
       documents_attributes: [:file]
     )
-    if modified_params[:content_text].blank? # when we rerequest the content_text: nil is mistakenly getting converted to ""
+    if modified_params.has_key?(:content_text) && modified_params[:content_text].blank? # when we rerequest the content_text: nil is mistakenly getting converted to ""
```
krschacht, line 106: wrap line

PR #283, opened 2024-04-19
Interface updated for message editing

krschacht:
* Wired up the edit (pencil) icon beneath messages. Clicking it activates inline editing.
* Cancel reverts the form and saving submits the form and creates a new version of the message

Note: editing does not work for messages with images. I cut this scope and noticed that ChatGPT also does not support this.

* All the system tests have been updated including new tests for edit functionality
* I've made another attempt to address some of the flappy system tests


app/views/messages/_message.html.erb
```diff
@@ -45,123 +49,177 @@
         <%= message.user? ? 'You' : message.assistant.name %>
       </div>
 
+      <turbo-frame id="message-<%= message.id %>">
+
       <div class="hidden" data-clipboard-target="text"><%= message.content_text %></div>
```
krschacht, line 54: All the diff the remaining diff in this file is simply indenting the block since it's now inside the turbo-frame

app/views/messages/_message.html.erb
```diff
@@ -45,123 +49,177 @@
...
-            %>
+          <% if message.user? && !message.has_document_image? %>
+            <%= link_to edit_assistant_message_path(message.assistant, message),
+              class: "cursor-pointer hover:text-gray-900 dark:hover:text-white flex items-center",
+              data: {
+                role: "edit",
+                turbo_frame: "message-#{message.id}",
+              } do %>
+              <%= icon "pencil", variant: :outline, size: 18, title: "Edit" %>
+            <% end %>
```
krschacht, line 165: Actually, this edit link_to was also changed within this diff

PR #302, opened 2024-04-25
[docker/compose] don’t expose ports and change pg volume name

matthewbennink:


krschacht:
@matthewbennink No problem. But that's a good idea to change the volume name, I should do make that change for everyone. But do you think others will want the ports closed or is that specific to what you're doing? I assume that's specific to some testing you're doing because I don't think the app will work with the ports are closed, although I'm not actually sure of that.

matthewbennink:
I'll provide some context, which may help.

For the ports, it was because the ports were already in use on my machine. At first, I tried to simply change them, but that required changing them within all of the connection strings as well. I was pleasantly surprised to notice that my local Docker setup was such that the containers were able to communicate with each other on those ports without them needing to be exposed on the host machine. I won't be able to access with `psql` or `redis-cli` from the host, but I _am_ able to run on port 3000 (which I made sure was open) and I'm also able to access through `docker compose run base rails console`.

As for the volume, it's because I received `PostgreSQL Database directory appears to contain a database; Skipping initialization`, and that's likely because some other Docker Compose file had a `pgdata` volume somewhere.

In summary, these changes were enough for me to run `docker compose up` locally and have a functioning app.

krschacht:
Dang, super helpful! I definitely want to tweak this so no one else runs
into these roadblocks in the future. Thanks for sharing.

PR #313, opened 2024-05-01
Improve Dockerfile & render.yml to fix conflicts + add health chk

krschacht:
Fix a bunch of issues @matthewbennink noticed over in this discussion:

https://github.com/allyourbot/hostedgpt/pull/302#issuecomment-2078162590

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,13 +6,13 @@ ENV BUNDLE_CACHE=/tmp/bundle \
   BUNDLE_JOBS=2 \
   PORT=3000
 
-WORKDIR /rails
+WORKDIR /app
 COPY Gemfile Gemfile.lock .tool-versions ./
 
 RUN --mount=type=cache,id=gems,target=/tmp/bundle \
   bundle install
 
-ENTRYPOINT ["/rails/bin/docker-entrypoint"]
+ENTRYPOINT ["/app/bin/docker-entrypoint"]
 CMD ["./bin/rails", "server", "-b", "0.0.0.0"]
 
 FROM base as development
@@ -40,6 +40,6 @@ RUN mkdir -p log tmp bin
 
 RUN adduser rails -D -h /rails -s /bin/sh && \
   chown -R rails:rails db log tmp bin && \
-  chmod 755 /rails/bin/docker-entrypoint
+  chmod 755 /app/bin/docker-entrypoint
 
 USER rails:rails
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -40,6 +40,6 @@ RUN mkdir -p log tmp bin
 
 RUN adduser rails -D -h /rails -s /bin/sh && \
   chown -R rails:rails db log tmp bin && \
-  chmod 755 /rails/bin/docker-entrypoint
+  chmod 755 /app/bin/docker-entrypoint
 
 USER rails:rails
```
krschacht, line 45: `rails` was not causing any problems but it made it a bit more difficult to grock these different scripts.

bin/docker-entrypoint
```diff
@@ -1,22 +1,17 @@
...
+    rm -f ./tmp/pids/server.pid 2>/dev/null
+    echo "Running db:prepare"
+    ./bin/rails db:prepare
+  elif echo "${1}" | grep -q "rails$" && [ "${2}" == "test" ]; then
+    export RAILS_ENV=test
+    echo "Running db:prepare"
+    ./bin/rails db:prepare
+  else
+    echo "Skipping db:prepare"
+  fi
```
krschacht, line 14: On Render migrations are handled with the `render-build.sh` script and on Fly migrations are handled with the build command within `fly.toml`. Basically, no situation yet actually needs this script to handle migrations and they're being double-called.

docker-compose.yml
```diff
@@ -3,20 +3,18 @@ version: '3.1'
 services:
   redis:
     image: redis
-    ports: ['6379:6379']
     healthcheck:
       test: ["CMD", "redis-cli","ping"]
 
   postgres:
     image: postgres:16
-    ports: ['5432:5432']
```
krschacht, line 12: These ports don't need to be exposed to the host machine and can sometimes cause conflicts.

docker-compose.yml
```diff
@@ -3,20 +3,18 @@ version: '3.1'
...
     restart: always
     environment:
       POSTGRES_USER: app
       POSTGRES_DB: app_development
       POSTGRES_PASSWORD: secret
     volumes:
-    - pgdata:/var/lib/postgresql/data
+    - hostedgpt_pgdata:/var/lib/postgresql/data
```
krschacht, line 17: This was getting prefixed by the directory name but if you repeat directory names then you can get a conflict on the volume name on the host machine.

PR #332, opened 2024-05-07
Fix markdown rendering with multiple blocks in one msg

### LINKED ISSUE #323: Indented markdown backticks do not render properly
Hello:

- Indented

  ```
  What's this
  ```

- Indented

I saw a small rendering issue with markdown. The two specific issues I noticed were (1) an extra line appeared at the end of the first couple markdown blocks and (2) further down the whole markdown sections got broken when it missed a triple backtick.

I don't know the cause of issue (1) and will need to inspect the code. But the cause of issue (2) is indented backticks, I think.

This is a screenshot:

image

This is the HTML string:

```
To check if Redis is running on your computer, you can use the `redis-cli` command with the `ping` subcommand. Here's how:\n\n```\nredis-cli ping\n```\n\nIf Redis is running, it will respond with:\n\n```\nPONG\n```\n\nIf Redis is not running, you'll see an error message like:\n\n```\nCould not connect to Redis at 127.0.0.1:6379: Connection refused\n```\n\nYou can also check the status of the Redis service using the appropriate command for your operating system:\n\n- On Linux with systemd:\n  ```\n  sudo systemctl status redis\n  ```\n\n- On macOS with Homebrew:\n  ```\n  brew services list\n  ```\n\n- On Windows:\n  ```\n  redis-server --service-name status\n  ```\n\nThese commands will provide more detailed information about the Redis service status.
```

Here is a better example of issue (1). There is an extra blank line in the first two markdown blocks but not in the third one!

image

This is the HTML string:

```
"To stop Redis that was installed and started with Homebrew on macOS, you can use the following command:\n\n```\nbrew services stop redis\n```\n\nThis command will gracefully stop the Redis service.\n\nIf you want to restart Redis later, you can use:\n\n```\nbrew services start redis\n```\n\nAnd to check the status of the Redis service, use:\n\n```\nbrew services list\n```\n\nThis will list all the services managed by Homebrew and their current status."
```

al3rez:
resolves #323


krschacht:
@al3rez Nice to see you! Thanks so much for figuring this one out. The fix looks good, nice and simple. Can you add a quick test for it just to ensure someone doesn’t accidentally regress it in the future?

krschacht:
Hey, I saw you pushed a test, thanks. But it looks like there’s a failure. I couldn’t quickly tell if that was an old test or the test you just added.

al3rez:
We'll push a fix soon.

Issue #340, opened 2024-05-09
Support auth via HTTP headers


krschacht:
Requests like #336, #333 will come up with requirements to support other means to authenticate besides username+password. Read: SSO to support "business" use cases (e.g. using generic oAuth 2, EntraID, Google SSO)

In my case to self-host applications like hostedgpt, I tend to use a "ZeroTrust" approach, where the application is not exposed to the public internet. ZeroTrust approaches, like with Tailscale ensure authentication of a user already on the infrastructure level. So no need to re-authenticate the user.

To support SSO, there are two ways to implement it:

1) Add support to omniauth and support with this a random set of SSO providers. This is a maintenance hell for this proejct
2) Allow to skip username+auth and allow to conduct authentication using a HTTP header. This outsources the authenticaton logic to 3rd party services. This is what I would recommend

# Deep dive in auth via HTTP header

A 3rd party service (e.g. a reverse proxy or a ZeroTrust appliance) would set a header, which if it exists means the user has been authenticated.

This is supported by Reverse Proxies like Traefik (see for example [BasicAuth](https://doc.traefik.io/traefik/middlewares/http/basicauth/) and especially [ForwardAuth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/) where they support a 3rd party auth server), Caddy with [forward_auth](https://caddyserver.com/docs/caddyfile/directives/forward_auth), Nginx with `auth_request_set` (source: [blog post](https://developer.okta.com/blog/2018/08/28/nginx-auth-request#bonus-who-logged-in))

ZeroTrust services like Tailscale can be used with [`$ tailscale serve`](https://tailscale.com/kb/1242/tailscale-serve). Tailscale sets in such a case so called "Identity headers" (ref. <https://tailscale.com/kb/1312/serve?q=header#identity-headers>)

The key is: the name of the "identity headers" is not standardized. It would need to be exposed as a configuration setting (e.g., environment variable). 

[...]

PR #344, opened 2024-05-09
Support auth via HTTP request header

### LINKED ISSUE #339: Support Google Auth login

tmaier:
Related discussion: #338
Closes #339, #340 

This is WIP. It will not work. I am just pushing it out in the open very early on.

Design decisions:

- Configuration happens via `Feature` with `features.yml` and environment variables defined within. This allows to configure this features on infrastructure level (e.g. via kubernetes or docker compose)
- The AuthN mechanism supported are username+password, google and http auth. You can enable username+password together with google. Enabling Http auth means google and username+password will be disabled.
- When a new user arrives and clicks on "log in with Google" or when a new user arrives with the right HTTP request header set, the `Person`/`User` will be created automatically.
   - This means, the `users#new` and `users#create` controller actions are only used for password-based AuthN
   - This means http-request-header-based AuthN needs to provide username and email address as well


README.md
```diff
@@ -24,12 +24,34 @@ ChatGPT uses your private conversations history to train its models. [OpenAI dis
...
 - [Changelog](#changelog)
 
+# Configuration
+
+Following features can be configured with environment variables:
+
+- `REGISTRATON_FEATURE` - Set to `false` to disable registration.
+  Default is `true`.
+- `AUTHENTICATION_APPROACH` - Set to `google` to use Google OAuth for authentication, `http_header` to use HTTP header authentication, or `password` to use password authentication.
+  Default is `password`.
```
krschacht, line 41: It’d be nice to be able to support multiple: google & password, for example. Would it ever make sense to support all three or if you are using HTTP_HEADER is that necessarily the only way to auth?

README.md
```diff
@@ -24,12 +24,34 @@ ChatGPT uses your private conversations history to train its models. [OpenAI dis
...
+- `AUTHENTICATION_HTTP_HEADER_NAME` - Name of the HTTP request header, which can be used to derive the name of a new user.
+  Required if `AUTHENTICATION_APPROACH` is set to `http_header`.
+  Default is `Tailscale-User-Name`.
+- `AUTHENTICATION_HTTP_HEADER_UID` - Name of the HTTP request header, which will be used as unique identifier of the user.
+  Required if `AUTHENTICATION_APPROACH` is set to `http_header`.
+  Default is `Tailscale-User-Login`.
+- `GOOGLE_CLIENT_ID` - Google OAuth client ID.
+  Required if `AUTHENTICATION_APPROACH` is set to `google`.
+- `GOOGLE_CLIENT_SECRET` - Google OAuth client secret.
+  Required if `AUTHENTICATION_APPROACH` is set to `google`.
```
krschacht, line 54: If you want to enable google auth with just a normal gmail account (not a custom google org) do you somehow get a google_client_id and secret? Or are those things only required when you’re doing a custom google org/domain?

I’ve never actually implemented google registration before.

docker-compose/tailscale/Caddyfile
```diff
@@ -0,0 +1,30 @@
+{
```
krschacht, line 1: Is this caddyfile something that tailscale needs?

db/schema.rb
```diff
@@ -137,7 +137,9 @@
     t.string "email"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.string "uid"
     t.index ["personable_type", "personable_id"], name: "index_people_on_personable"
+    t.index ["uid"], name: "index_people_on_uid"
```
krschacht, line 142: People are unregistered humans, Users are registered humans. Do you want this `uid` on the User table or Person? 

To be honest, introducing Person was a bit of a premature abstraction on my part. I was modeling the registration off the Basecamp pattern where deactivated users convert to a Tombstone object, invited users are People but no User or Tombstone.

app/models/user.rb
```diff
@@ -5,7 +5,7 @@ class User < ApplicationRecord
   has_secure_password
   has_person_name
 
-  validates :password, length: { minimum: 6 }, allow_nil: true
+  validates :password, length: { minimum: 6 }, allow_nil: Feature.authenticate_with_password?
```
krschacht, line 8: It feels backwards that this as ever true — wouldn’t we want to NOT allow nil passwords, if we’re supporting password-based auth?

config/routes.rb
```diff
@@ -16,18 +16,31 @@
...
+    post "/login", to: "sessions/passwords#create"
+    get "/register", to: "users#new"
+    get "/logout", to: "sessions/passwords#destroy"
+  end
+
+  constraints lambda { |_| Feature.authenticate_with_google? } do
+    get "/login", to: "sessions/google#new"
+    get "/auth/google/callback", to: "sessions/google#create"
+    get "/logout", to: "sessions/google#destroy"
+  end
```
krschacht, line 43: I don’t love the complexity of the routes & controllers now. I’m staring at this and trying to think of how it can be simplified so it’s not so much to wrap your mind around. I don’t have a great suggestion yet.

Do we need to conditionalize routes? What if we just set up all the routes and then within them we have before_filters that forward if you shouldn’t be allowed to use it. I think that may be a simpler convention.

Also, do we really need to namespace controllers? Something smells off about that. For example, PasswordsController doesn’t feel right. Password isn’t the model we’re concerned with: we’re Creating / Reading / Updating / Deleting passwords. Like, we’re not Creating new Google’s (which is what a GoogleController would imply). I’ll think more about this …

krschacht:
@tmaier Looking good! I did a quick pass through it and left a handful of thoughts on things that struck me as I’m starting to wrap my mind around the new architecture.

README.md
tmaier: Password and Google could co-exist. HTTP Auth would not make sense to co-exist with anything else

In many cases, one would like to disable password auth when one has Google Auth enabled

db/schema.rb
tmaier: About the premature abstraction: 

I think the key question is how you would like to envision registration to happen in the long term. Out in the open for everyone like it is now?

Or is it a more private service, where an initial user gets created and others get invited.

tmaier: I am going to move the uid to the User model

docker-compose/tailscale/Caddyfile
tmaier: There is a readme file next to it (actually two readme files).

This is an example how one could get this up and running with a Caddy Reverse Proxy, which is responsible for the Tailscale connection. Caddy needs a Caddyfile.

I can also remove this entirely and put it into a gist or a blog post or something

README.md
tmaier: To enable Google SSO, you will need (a) a Google Cloud account, (b) configure a "consent screen" (c) register your application. Then you get the client id and the client secret.

config/routes.rb
tmaier: It was an idea to isolate the logic. But I agree, it is not that nice. I will change it.

app/models/user.rb
tmaier: Agreed. My brain was twisted.
`allow_nil: true` should not be here

tmaier:
I wonder how to properly support HTTP request authentication

Situation:

- The User/Person models require email, name

Complication:

- HTTP request authentication may just set one header, which allows to find/create a user. Everything else may not be  shared. The value set, does not need to be an email address.

Resolution options:

1. We assume more headers are available, which allow us to set these values
2. We make some (e.g. name optional, especially when HTTP-login is ongoing.
3. We create a new model, next to `User`, maybe `HTTPUser`

README.md
tmaier: I will change it, as described above

README.md
tmaier: I will add this to the README later

README.md
```diff
@@ -24,12 +24,34 @@ ChatGPT uses your private conversations history to train its models. [OpenAI dis
...
+- `REGISTRATON_FEATURE` - Set to `false` to disable registration.
+  Default is `true`.
+- `AUTHENTICATION_APPROACH` - Set to `google` to use Google OAuth for authentication, `http_header` to use HTTP header authentication, or `password` to use password authentication.
+  Default is `password`.
+- `AUTHENTICATION_HTTP_HEADER_EMAIL` - Name of the HTTP request header, which can be used to derive an email address of a new user.
+  Required if `AUTHENTICATION_APPROACH` is set to `http_header`.
+  Default is `Tailscale-User-Login`.
+- `AUTHENTICATION_HTTP_HEADER_NAME` - Name of the HTTP request header, which can be used to derive the name of a new user.
+  Required if `AUTHENTICATION_APPROACH` is set to `http_header`.
+  Default is `Tailscale-User-Name`.
```
tmaier, line 47: This is what I would like to remove. This would mean email and name will not be set.

app/controllers/sessions_controller.rb
```diff
@@ -32,4 +25,29 @@ def destroy
...
+    person = Person.find_by(email: params[:email].strip.downcase)
+
+    @user = person&.personable
+  end
+
+  def find_user_for_google_authentication
+    auth = request.env['omniauth.auth']
+    return if auth.blank?
+
+    @user = User.find_or_create_by!(uid: auth[:uid]) do |user|
```
tmaier, line 47: Here, I could also check for `Feature.enabled?(:registration)`... (see `#find_current_user_based_on_http_header`

app/controllers/users_controller.rb
```diff
@@ -36,6 +37,12 @@ def update
 
   private
 
+  def authorize_password_login
+    return if Feature.authenticate_with_password?
+
+    halt 403
```
tmaier, line 43: This is the most simple way to show a 403

krschacht:
I don’t quite follow this, can you say more about it:

> HTTP request authentication may just set one header, which allows to find/create a user. Everything else may not be shared. The value set, does not need to be an email address.

I think what I’m hearing is that a user can be in this other system, click a link, and they’ll be directed to some route within our app which should register and authenticate them all in this moment? And this request does not pass any information about the user except a key in this header, so we won’t know anything else about them? If I’m following this correctly, how feasible is it for this request to be required to include a little more information (eg name and email).

In general the way I think we should approach this is to introduce a Credentials model. We move details from User over to Credentials and we make this table be STI so it can be the backing for EmailCredential model and a HTTPCredential model.

tmaier:
@krschacht  I was imprecise in my words:

1. We will always get something like a unique identifier for a user. In some cases it may be an email address. In some cases not.
2. We might not get any other data, like name or email address. This is the complication in respect to `User` and `Person` models

krschacht:
With those three ENV you are going to rename to: is NAME the full name? And what’s the USER one?

app/controllers/users_controller.rb
tmaier: Alternate: Like `ensure_registration`: Redirect to `redirect_to root_path`

docker-compose/tailscale/Caddyfile
tmaier: I removed it.

README.md
tmaier: I updated the README and refactored the feature flags

README.md
tmaier: Done

tmaier:
@krschacht can you have one more look at this PR? I refactored it.

krschacht:
@tmaier I took a quick look and I think it’s on the right track. In your mind, what parts are still remaining?

I think I’d like to land a small refactor of Features ahead of this. You need to extend its capabilities for your work here in this PR, and I also need to extend its capabilities in another branch I’m working in. So tomorrow I will try to spin up a new PR that just focuses on Features, I’ll closely reference what you’re doing in this branch to incorporate the extensions that you need. Then after you get a chance to review it and I merge into main, I can help in pulling main back into here.

app/views/sessions/new.html.erb
```diff
@@ -6,34 +6,57 @@
...
   <h2 class='text-3xl font-semibold'>Welcome back</h2>
 
-  <span class='text-sm text-center'>
-    Don't have an account? <%= link_to "Sign up", register_path, class: "underline" %>
-  </span>
+  <% if Feature.authenticate_with_password? %>
+    <span class='text-sm text-center'>
+      Don't have an account? <%= link_to "Sign up", register_path, class: "underline" %>
+    </span>
+  <% end %>
```
krschacht, line 13: In addition to hiding this link, somneone could visit this /users/new route directly so we should add a redirect in users_controller on the new route. Do we also need one on the create route? Probably would be a good idea so maybe a before_filter that applies to them both.

app/models/person.rb
```diff
@@ -6,7 +6,7 @@ class Person < ApplicationRecord
 
   validate :personable_id_unchanged, on: :update
   validates_associated :personable
-  validates :email, presence: true, uniqueness: true
+  validates :email, email: true, presence: true, uniqueness: true
```
krschacht, line 9: I did a double take on this line and was pretty sure it was a mistake, then after asking chatgpt I realized it's referencing the custom validator. Can we put that at the end and rename the custom validator? So:

validates :email, presence: true, uniqueness: true, valid_email: true

(and then rename the validator to ValidEmailValidator, a bit wordy but I think it makes the validates line much clearer)

app/controllers/users_controller.rb
krschacht: How about we rename the filter and check them both in one:

`before_action :ensure_registration_allowed`

then:

```
def ensure_registration_allowed
  if Feature.disabled?(:registration) || Feature.disabled?(:authenticate_with_password)
    redirect_to ...
  end
end
```

(I added disabled? in my other PR)

krschacht:
@tmaier I gave this PR a close read and added a few small comments. I also pulled out the Feature work into it's own PR. Give this a quick review, then I can merge into main and we can finish up this PR: https://github.com/AllYourBot/hostedgpt/pull/372

I'd love to get your PR merged in this week, if possible. I have some additional work I'm doing where I want users to be able to give their Assistant access to their Google Calendar / Gmail (optionally) and I'm building off the google auth in this PR.

krschacht:
@tmaier I landed the Features PR into main, I then pulled main back into your branch here. When I did this, I resolved the merge conflicts, I did the find & replace on Feature.* and Setting.* so your code matches the new names.

Note: there is one final rename that I did:

```
    http_header_auth_uid: <%= ENV["HTTP_HEADER_AUTH_UID"] || "X-WEBAUTH-UID" %>
```

I made the default `X-WEBAUTH-UID` to match the symmetry there. I hope this is okay.

I am doing some google auth related work in another branch so I'm spending a little more time studying how this PR approaches that.

krschacht:
Okay @tmaier , I merged that other PR into main and then merged main back into this. I hope this doesn't randomize your efforts too much! You have a lot of good changes in here that I tried to preserve, but I may have accidentally introduced a bug while resolving merge conflicts. Thanks for your help on all this auth stuff!

krschacht:
Hi @tmaier do you think you’ll have time to wrap up this PR this week? You got it really close, and after peeling off the Google Auth and getting that merged in this shouldn’t take too much to get this one finished.

config/options.yml
```diff
@@ -10,4 +10,4 @@ shared:
     google_auth_client_secret: <%= ENV["GOOGLE_AUTH_CLIENT_SECRET"] || Rails.application.credentials[:google_auth_client_secret] %>
     http_header_auth_email: <%= ENV["HTTP_HEADER_AUTH_EMAIL"] || "X-WEBAUTH-EMAIL" %>
     http_header_auth_name: <%= ENV["HTTP_HEADER_AUTH_NAME"] || "X-WEBAUTH-NAME" %>
-    http_header_auth_uid: <%= ENV["HTTP_HEADER_AUTH_UID"] || "X-WEBAUTH-USER" %>
```
tmaier, line 13: The thing is, when you google for `X-WEBAUTH-USER`, you will find examples which use this name. This is what I would stick with this instead of `-UID`

db/migrate/20240509231839_add_uid_to_users.rb
```diff
@@ -0,0 +1,6 @@
+class AddUidToUsers < ActiveRecord::Migration[7.1]
+  def change
+    add_column :users, :uid, :string
```
tmaier, line 3: TODO: Rename `uid` to `auth_uid`

db/schema.rb
```diff
@@ -300,6 +300,7 @@
     t.string "auth_uid"
     t.index ["auth_uid"], name: "index_users_on_auth_uid", unique: true
     t.index ["last_cancelled_message_id"], name: "index_users_on_last_cancelled_message_id"
+    t.index ["uid"], name: "index_people_on_uid", unique: true
```
tmaier, line 303: TODO: Remove this index

tmaier (commented):
I did a quick review of the code.

What else is missing?

- [x] Tests

app/controllers/concerns/authenticate.rb
```diff
@@ -27,4 +33,24 @@ def user_signed_in?
...
+    if Feature.registration?
+      Current.user = User.find_or_create_by!(auth_uid: request.headers[Setting.http_header_auth_uid]) do |user|
+        user.create_person!(email: request.headers[Setting.http_header_auth_email])
+        user.name = request.headers[Setting.authentication_http_header_name] || person.email
+      end
+    else
+      Current.user = User.find_by(auth_uid: request.headers[Setting.http_header_auth_uid])
+    end
+    Current.person = Current.user&.person
+  end
```
krschacht, line 55: TODO: This private helper method should probably end with the same consistency check `Current.user = nil if Current.person.nil?`

app/models/person.rb
krschacht: Agreed, let's just remove the custom validator and go with the `format: { with: `

app/views/sessions/new.html.erb
```diff
@@ -7,14 +7,14 @@
   <h2 class='text-3xl font-semibold'>Welcome back</h2>
 
   <% if Feature.registration? %>
-    <span class='text-sm text-center'>
+    <span class="text-sm text-center">
```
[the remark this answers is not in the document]
krschacht: I've been trying to default to double quotes everywhere

tmaier:
@krschacht I think everything is done now. I never worked with `ActionDispatch::IntegrationTest`. I hope the test cases are ok for you.

krschacht:
@tmaier Since this is auth, I re-read through everything again one more time and tested out the various scenarios. I did a few tiny cleanup commits along the way and there is only one material change I made: 

I flipped the feature flags in such a way that got me into an infinite redirect loop (directing me to root which directed me to login but I had disabled password & google so it was redirecting me to root). I fixed this by simply allowing us to render a login (and registration) page even though all the options are hidden. This has the added advantage of ensuring that I see just a single flash message explaining what's wrong, which should be easier for people to diagnose what state the system is is. I don't think the flash messages were properly persisting through the redirect.

(this is the very last commit, if you want to take a look at it)

But otherwise, I'm ready to click merge on this. Any hesitation at all before I do? Just double checking, since it feels dangerous for me to click merge after just making a last-minute change to an auth flow.

Thanks again for taking the lead on all these great auth improvements! i don't know when I would have gotten to this.

app/controllers/application_controller.rb
```diff
@@ -1,3 +1,11 @@
 class ApplicationController < ActionController::Base
   include Authenticate
+
+  private
+
+  def ensure_authentication_allowed
```
tmaier, line 6: Consider renaming it to be specific which authentication is being checked.

For example: `ensure_session_based_authentication_allowed`

app/controllers/application_controller.rb
```diff
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
 
   def ensure_authentication_allowed
     if Feature.disabled?(:password_authentication) && Feature.disabled?(:google_authentication)
-      redirect_to root_path, alert: "Password and Google authentication are both disabled."
+      flash[:alert] = "Password and Google authentication are both disabled."
```
tmaier, line 8: Wouldn't it be better to render a classical 401 page instead?

E.g. an exception app or `rescue_from` ...

krschacht: Yes, good point. That's much simpler.

krschacht:
I made those last two changes, good catches. Merging it in now! Whew.

tmaier:
Great! Thank you very much! This was a bit of a collaborative effort. I like it.

PR #362, opened 2024-05-19
Google tools

krschacht:
This PR was getting too big so I pulled the first tool — weather — out in it's own PR so I can land the abstractions. Once I get that merged into main then I'll come back to this.


app/services/open_weather.rb
```diff
@@ -0,0 +1,27 @@
...
+    Setting.openweather_key
+  end
+
+  #describe :get_current_weather, "Get the current weather in a given location"
+  def self.get_current_weather(city_s:, state_s:, country_s: "US")
+    location = get(
+      "http://api.openweathermap.org/geo/1.0/direct?q=#{city_s},#{state_s},#{country_s}&limit=1&appid=#{key}"
+    ).first
+
+    response = get("https://api.openweathermap.org/data/3.0/onecall?lat=#{location.lat}&lon=#{location.lon}&units=imperial&appid=#{key}")
```
mgomes, line 14: FWIW, https://github.com/open-meteo/open-meteo doesn't require an API key. Battled with this myself.

krschacht: @mgomes oh thanks! It did bug me that this needed another key. I’m working really hard to keep the complexity and dependencies of this app really, really low. (I’m proud that I got ActiveJob, ActionCable, *and* image uploading to all use postgres! :)  So even this one single extra key was bugging me. I’ll definitely look into open meteo.

BTW, I didn’t know anyone was following along at home with this PR. :) Any thoughts on it or things you’d like to see?

And in case you didn’t see the post, I got voice working in the app already: https://github.com/allyourbot/hostedgpt/discussions/348  The feature flag still defaults to being off while I chip away at the long-tail of edge cases around voice, but I’m using voice daily myself and it’s getting pretty good. So now that I’m talking to GPT all day long I want to completely replace all my Google Home use cases with this (weather, timer, music) in addition to lots of other stuff (gmail, google calendar, and google tasks). So that’s how I’ve been approaching this stream of work. But any thoughts.

mgomes: Very cool. I am really interested in the tool integration as well. One of the reasons ChatGPT performs so much better than other models (including plan GPT4) is because of the Python sandbox it can use to write and run Python scripts.

I've built something similar for Ruby and I am looking to see how well it'd fit into hostedgpt. It uses Heroku's one-off dynos to run code in short-lived dynos. It can be extended to other languages as well.

krschacht: @mgomes That could be really interesting to incorporate! I've never taken much advantage of the python sandbox since I do so little in python. For me the only time it seems to kick in is when I ask it a math question and it does some quick calculation in python. I'm curious to hear more about the times you end up using it.

My thought has been: if we can get a ruby sandbox, then I could solve all the hallucinating problems with ChatGPT because it could test out any code it's suggesting. Ideally, the sandbox isn't just ruby but also has a set of common gems pre-loaded (e.g. Rails) so that it can quickly hop into a rails console, verify that a method call actually works in the way that it says.

PR #367, opened 2024-05-21
Assistant model updates

Rmpanga:
# Description of Changes

### Change 1 Created a new column called `external_id` to the following models
1. Assistant 
2. Conversation
3. Run
4. Step 

**What does `external_id` represent?**
-  It represents the backend api's id. If we created an OpenAI assistant Open AI has an ID that they provide clients. This ID is used to perform API calls on their assistant. Hence why we need to store it somewhere

### Change 2: The `external_id` column is unique only for assistant and conversation models
- I felt like these two HAD to be unique, can make an agruement for Run and step to be unique as well but don't feel as strongly on this.

### Change 3: Added annotation gem
- I love this gem because you can look at model class and know exactly what its database properties are. 

# Testing done

- [x] Confirmed migration succeeds
- [x] Confirmed rollback is successful along with a following migration 
- [x] TODO: Confirmed app still functions


krschacht:
The external_id makes sense. Thanks for adding this!

Can we remove the annotate gem, though? I am really trying to minimize external dependencies since every gem is one to keep updated over time. I also don’t the comment at the top of the model files. People can just open schema.rb or often a dev already has `bin/rails dbconsole` open anyway `\dt table;` or has `bin/rails c` open and `model.attributes`. I feel like most devs have fallen into some habit already since having a big comment at the top of models isn’t a standard pattern we’re all invariably falling back to one of these other methods which every project supports.

krschacht:
BTW, I have an overall thought about this direction, let me throw it back in our other thread: https://github.com/allyourbot/hostedgpt/discussions/345

Rmpanga:
> `rails dbconsole`

Yes I can remove the annotate gem. I find it useful but if other don't that no problem.

krschacht:
I went ahead and cleaned this up so I could get it merged in.

PR #370, opened 2024-05-22
Add the first tool: checking today's weather

krschacht:
This puts an abstraction in place for Tools. Re-works the AI backends for OpenAI (and get message job) to be able to handle all the logic for tool calling. There is good test coverage for all these new cases.

The tool that's in place is pretty trivial; it just shows today's weather. But it's a good start.


app/services/toolbox.rb
```diff
@@ -0,0 +1,133 @@
...
+  end
+
+  def self.formatted_function_parameters_with_types(name)
+    function_parameters(name).map { |param| formatted_param_properties(param) }.to_h
+  end
+
+  def self.formatted_param_properties(param)
+    raise "The param '#{param}' is not properly named for the type to be inferred (e.g. is_child, age_num, name_str)" if param.to_s.exclude?('_')
+
+    case param.to_s.split('_')
```
mgomes, line 87: @krschacht What do you think about a refactor to remove types from the argument names? The type inference feels complicated and affects the argument names. I am thinking something like method decorators could be nice.

So right now the weather function looks like this
```ruby
describe :get_current_and_todays_weather, <<~S
    Some descrption.
S

def self.get_current_and_todays_weather(city_s:, state_province_or_region_s:, country_s: nil)
# code
end
```

The refactor would look like:

```ruby
function_description <<~S
    Some descrption.
S
function_params city: :string, state_province_or_region: :string, country: :string
def self.get_current_and_todays_weather(city:, state_province_or_region:, country: nil)
# code
end
```

The two function calls get automatically associated with the method and perhaps stored in a class var.

mgomes: It might also be cleaner to no implement all tools as class methods and instead require the tool class to be instantiated first.

krschacht: I initially had extra declarations for those, but it felt verbose. And also, it slowed down prototyping.

I want to create a lot of tools really soon. I want to support being able to create a class with a few methods and as long as you inherit from Toolbox, it works as a tool. I really want to minimize the extra configuration you have to do. However, OpenAI **requires** a description of the function and types on all the params. I managed to auto-generate a reasonable description of the function. But I could not figure out any way to auto-generate the types. I tried a bunch of different approaches with default values and parsing source, but in the end, a simple thing that resembles `.to_s` and `.to_i` worked quite well. I found it to be succinct, it's a subtle reminder that there is something special about this function, but without bogging you down in learning a bunch of new declarations in order to add something to the Toolbox. And when I had two separate declarations, as I was prototyping and moving things around, they were regularly out of sync. I'd add a param on one line but forget to add it on another. I'd rename in one place but not in another. It was starting to feel like I was back in the land of Java with so much explicit declaration. :) Anyway, that was my thinking! Let's live with it for a little while and see how it plays out.

Oh, on singleton class vs instantiatiable class, I didn't have a strong feeling there. It was just that we will never actually instantiate more than one of these. Isn't the convention to make those singletons? It is annoying to type "self." in front of each one but I probably should have just wrapped it all in `class << self` — I'm not sure why I didn't do that. That would make it cleaner.

mgomes: Thanks for the explanation and background!

Is it currently generating descriptions for each argument? I don't see that.

I think my counterargument would be that encoding types/descriptions into the parameter names also requires learning and makes their behavior less explicit to new devs. I'm also worried to use underscores in my parameter names in case it interferes with the parser. It doesn't seem like it would be a big lift for devs to annotate the methods they expect to be visible to the LLM -- the syntax can feel similar to RBS or Sorbet.

Re: the singleton class -- I'm not sure, but I do think a tool can be called more than once within a single request by the LLM.

krschacht: We’ll try for a little while. I’m not primarily optimizing for learning curve — I agree the approach I picked will take a little bit extra learning. But I think the benefits are overcome before you’re done writing your first tool.

I agree with the Sorbet comparison. I used Sorbet for 1 year on a new product with a team of 3 engineers and at the end of 12 months we all agreed to remove all types and rip Sorbet out. :) That’s a longer conversation with many considerations, but one of the roots was: having to keep data in sync across two places creates lots of unexpected out of sync’ness.

We’ll try it with the clever param naming for a little while and see how it plays out. I’ve never tried this approach before, but I”m curious to see how it bakes.

Regarding the singleton, I’m not sure if a tool can be called more than once in the same request, but let’s assume we should support that. Even if the LLMs dont do that now they easily could add it.

It’s a really good consideration to make sure the code isn’t backing us into a corner and not supporting multiple calls to the functions. Is there an aspect of how the classes are structured, as singletons, that does that? I might have messed that up.

mgomes: Sounds good. Re: the per-argument descriptions, are those getting autogenerated somehow or is there a way to define those?

mgomes: Oh I missed your question about singletons.

In your OpenMeteo tool, for example, you are passing your params to your private methods each time you invoke them. Another approach would be to initialize instance vars for each of the calling params. This would make most of the private methods parameter-less. In a larger tool, that would be a good refactor.

If you start setting and relying on instance vars on a singleton class that you instantiate more than once, you might start hitting race conditions similar to class vars.

PR #374, opened 2024-05-24
Don't render signup link if registration feature is disabled

mgomes:
Hides the signup link since the controller just redirects back to root path if the registration feature is disabled.

### DIFF
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -6,7 +6,7 @@
 
   <h2 class='text-3xl font-semibold'>Welcome back</h2>
 
-  <% unless Feature.registration? %>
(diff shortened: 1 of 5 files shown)


mgomes:
Going to fix the system tests.

krschacht:
@mgomes but the system test shouldn't fail by making this change because the feature is enabled by default so the link **should** be there. I actually think your conditional is backwards :)

mgomes:
@krschacht Ok, fixed the logic and added a system test. I had to move some of the private methods in the feature tests to helpers. I think they'll be useful, though, as more features get introduced and need to be tested in isolation.

krschacht:
Good call, thanks for cleaning that up! Merging in

PR #375, opened 2024-05-24
Add support for Google Auth registration & login

krschacht:
This is peeling some of @tmaier work in #344 into it's own PR.

* A feature flag now lets you enable Google Oauth registration
* When this is turned on additional steps (explained in README) need to be followed to create a google auth service
* A new Google button appears on login & signup forms
* New endpoints that receives success after google auth or failure after google auth

I added tests for all the google_oauth_controller cases but some additional tests that are missing:

- [x] Verify that turning the feature on causes a button to appear on login & signup
- [ ] Verify that when logged in, cancel takes you back to settings. When logged out, cancel takes you back to login.


krschacht:
@tmaier Want to take a look at this PR? The thing that made this tricky is that I want the app to support **registering** with Google, but I also want it to support being able to give your Assistant access to Gmail, Google Calendar, and Google Tasks. I managed to handle this by creating multiple omniauth providers, each with a different set of permissions, and then differentiating these cases on the callback.

This led me to create a sessions/google_oauth_controller with a monster create action. This creates a new session and, if needed, also creates a new user (so it's a little off being nested under sessions but I couldn't think of any clean way to separate those things).

I think I successfully pulled melded together your efforts to add google auth with my efforts to add these connect-to-google permissions. However, this PR focuses on auth and just lays the foundation for the connect-to-google stuff. I have another PR coming which will turn those google services into tools that Assistant can have access to. Once we merge this into main, then your PR can end up just adding on the final HTTP_AUTH stuff.

I hope peeling off some of your work into these two additional PRs didn't randomize you too much! I really appreciate the work you've done on this and was just trying to minimize our merge conflicts by doing it this way.

tmaier:
Hi @krschacht I reviewed the code. It looks good :)

I will pick up the the other PR, when this one landed, ok?

PR #383, opened 2024-05-30
Restrict messages endpoint to Current.user

krschacht:


app/controllers/messages_controller.rb
```diff
@@ -79,7 +79,7 @@ def set_assistant
   end
 
   def set_message
-    @message = Message.find(params[:id])
+    @message = Current.user.messages.find(params[:id])
```
matthewbennink: If you set `has_many :messages, :through => :conversations` inside the `User` class, you should be able to do this.

krschacht: I thought about that, but it's a little confusing because you'd expect `Current.user.messages` to be all of the user's messages. But that result set will include messages created by the assistant.

Also, I've been preparing for a world in which there are more than two actors in any conversation (OpenAI has said this is coming).

Imagine there are two different assistants in one conversation, that doesn't add much complexity. But imagine a world where there are two different users in a conversation. `Current.user.messages` would include messages that another user wrote. That's confusing.

And, which messages should you be able to see? Any messages which are connected to a conversation that you have access to.

So it seems a little convoluted that I retrieve the message and then check it's conversation. Especially in a world where conversations are only attached to a single user:

```
@message = Message.find(params[:id])
redirect_to root_url, status: :unauthorized if @message.conversation.user != Current.user
```

But I'm basically preparing for a world in which conversations may have additional participants. I think it's proper for us, even right now, to view `conversation` as the "access control element".

krschacht: I'll go ahead and merge in now so that this fix is in main (I got distracted earlier and forgot to merge). But continue the conversation if you have more thoghts. I can always do a follow up PR if needed.

PR #384, opened 2024-05-30
Add language models to seed

nwarwick:
I was running into an issue setting up the application locally, specifically at the seeding step (see image below). I realized it was due to the fact that it was trying to create assistant records without having language model records to connect them to.

Screenshot 2024-05-30 at 2 23 07 PM

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -3,7 +3,7 @@ require "active_record/fixtures"
 if Rails.env.development?
 
   puts "loading fixtures"
-  order_to_load_fixtures = %w[people users tombstones assistants conversations runs messages steps]
+  order_to_load_fixtures = %w[people users tombstones language_models assistants conversations runs messages steps ]
 
   ActiveRecord::Base.transaction do
     ActiveRecord::Base.connection.disable_referential_integrity do
--- a/test/seeds_test.rb
+++ b/test/seeds_test.rb
@@ -2,7 +2,7 @@ require 'test_helper'
 
 class SeedsTest < ActiveSupport::TestCase
   def setup
-    Rails.application.load_seed 
+    Rails.application.load_seed
   end
 
   test 'seeds run without errors' do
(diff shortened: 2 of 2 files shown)


krschacht:
@mwarwick I really appreciate you submitting a fix for this! I inadvertently broke this when I merged in a recent PR. Adding a test case for seed is exactly what I needed.

BTW, what's motivated you to set up HostedGPT?

And as you get started using it, please let me know how it's working out for you. I monitor the Discussions (and PRs) closely and am working on this project a lot so I love to hear any feedback about what could improve it for you.

nwarwick:
My pleasure @krschacht!

> BTW, what's motivated you to set up HostedGPT?

I primarily just wanted to take a peek at how you solved a few things (It looks/works great btw). I am building a new Rails app which includes a chat interface with an LLM.

krschacht:
@nwarwick Nice. It may be worth you knowing this then... one thing I really want to do is basically take almost all the code within this app and put it into an engine. That way people who want to build a new app like, say, "ChatGPT for data analytics" can mount the engine, override a few views, add some appropriate hooks like, say, after_message_processed.

I know of a few people using this in their startups, but they need to make so many changes/additions that they basically copy all code into their own app. And while I don't mind this at all, people who do this then won't benefit from all the future downstream changes I make.

So if this HostedGPT codebase had an appropriate abstraction in place (and I think engine's are the right abstraction) then it would set people up for deep customization without permanently forking/copying.

I don't know when I'll get to that, it'll be at least a couple months. But I mention this in case it impacts your plans! And, in case, maybe, possibly, perhaps, you decide you want to help with that. ;)

nwarwick:
That's a great idea @krschacht! For our usecase we need to provide extra context via a RAG system which is in another app and deals with hitting OpenAI. I will definitely be keeping a close eye on this project and helping where I can!

krschacht:
Nick, I've been considering adding langchainrb as the core library that
this HostedGPT is built on top of. It has RAG built in. There was another
contributor investigating this to see what would be involved in integrating
langchainrb. Anyway, I just mention that in case it's helpful. I don't have
a timeline for you on that one.\

PR #391, opened 2024-06-06
Adds `rack-cors` gem and configures it via an ENV variable

matthewbennink:
Adds the [rack-cors](https://github.com/cyu/rack-cors) gem and configures it if the `ALLOWED_REQUEST_ORIGINS` ENV variable is present. This enables another site to send HTTP requests on behalf of the logged-in user and to connect to the Action Cable Websocket to receive message updates.


krschacht:
@matthewbennink I get the basics of CORS so I understand what this is accomplishing, but I don’t completely understand the code. I assume you tested it. :) I’m going to merge in without testing it myself since it looks like it’ll be fully skipped if no one touches the setting.

PR #393, opened 2024-06-06
Respect prompt's language when generating the autotitle

ceicke:
When generating the autotitle, respect the language that the text prompt of the user was written in.

### DIFF
--- a/app/jobs/autotitle_conversation_job.rb
+++ b/app/jobs/autotitle_conversation_job.rb
@@ -29,7 +29,7 @@ class AutotitleConversationJob < ApplicationJob
     <<~END
       You extract a 2-4 word topic from text. I will give the text of a chat. You reply with the topic of this chat,
       but summarize the topic in 2-4 words. Even though it's not a complete sentence, capitalize the first letter of
-      the first word unless it's some odd anomaly like "iPhone". Make sure that your answer matches the language of 
+      the first word unless it's some odd anomaly like "iPhone". Make sure that your answer matches the language of
       the text of the chat tht I give you.
 
       Example:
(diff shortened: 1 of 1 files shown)


krschacht:
Guten tag @ceicke! That’s about all the german I know. :) Thanks for these improvements! I’m getting ready to merge them in.

I’d love to hear more about how you’re using the project and if there is anything you’d like to see from it?

ceicke:
Hey @krschacht 
I use LLMs really for everything work/non-work related but found myself not willing to shell out 2x20€ per month for OpenAI and Anthropic. so hosting this on my own server gives me the option to just pay per use. My daughter uses the instance I am hosting for studying for her school exams.

ceicke:
In terms of what I’d like to see:

- localization (I started a branch on my dev machine yesterday)
- Dall-e intergration
- Local LLM integration (probably simplest with Ollama)
- Some RAG extension for the cherry on top

krschacht:
@ceicke Thanks for sharing! I could definitely use help with localization. I've never done localization for a rails app before even though I know there is a standard built-in way to do it. If you can help with getting localization setup and all the current strings localized that would help me and others to follow that pattern going forward.

On your other items:

> Local LLM

Local LLM integration is almost done, the first big PR was merged in and the second half of it has been tee'd up as a PR and it's being finalized. I hope this will get finished sometime in the next week.

> Dall-e integration

I want this for myself too, eventually. It's not one of my top tools so I haven't prioritized it, but I have put all the pieces in place so this wouldn't be too difficult to tackle, if you wanted to take it on. A pattern for tools exists, we already have an API for Dall-E, and we have the ability to save & display images within conversations. I could give you guidance on what an implementation would look like, if you ever want to turn tackle it.

> Rag extension

I do eventually want to this for myself, as well. I did a bit of investigating into another ruby library that does some of the hard work (langchainrb) but a good bit of research and planning would still need to be done for this.

PR #394, opened 2024-06-08
Fix registration regression for Language Model

sealocal:
A small change to ensure that respective Language Models are created for the Assistants that are created upon registration.

 In production, the DB is not seeded, while in development the DB is seeded with fixtures that create these Language Model records.

I have not tested this.


matthewbennink:
I've run into this same issue a couple of times since the `LanguageModel` came about since I was starting with a fresh database and so not running the individual migrations, and since it wasn't a dev environment, I wasn't running `db/seeds.rb` either. The quick fix for me was to copy/paste [these lines](https://github.com/AllYourBot/hostedgpt/blob/main/db/migrate/20240523080700_create_language_models.rb#L12-L45) and run them inside a Rails console once the database has been setup.

Perhaps we could have a script somewhere that does this once, e.g., `bin/setup_language_models` or something like that. Or maybe we could augment `db/seeds.rb` to load the language models only in non-development environments.

krschacht:
@sealocal Thanks for starting this PR. And @matthewbennink thanks for pointing out the full solution. I'll admit, I'm having a little trouble ensuring I don't break new-production setup with this app since there are so many situations I'm supporting! For each change like this, I really need to test development setup, existing production setups, new production setups on Fly, Render, and Heroku. :(

In any case, I definitely need to fix this but I'm going to change the solution slightly so that there is less chance I break it again. As Matthew was pointing out, there is already a seed setup and that really should just run for development and production. And then there will be less chance I break this again since I drop & prepare the DB in development all the time. Let me make some changes on this PR.

krschacht:
Okay, merging in. I kind of took over this PR but it was still super helpful b/c you alerted me to the issue, it caused Matthew to remind me we had a whole table in a migration, so it pointed the way for me to fix it.

krschacht:
@sealocal Thanks again Mike! And  I’d love to hear more about how you’re using the project and if there is anything you’d like to see from it?

PR #396, opened 2024-06-08
Refactor credentials and authentication & add Gmail auth

krschacht:
We already have 3 authentication schemes — password, Google OAuth, and Http Header — and all of these are stored in the user model. However, I want to add API authentication (Http Bearer token) and auth for multiple Google tools (gmail, google calendar, and google tasks).

This PR refactors the authentication code. The main things it does are:

* Better organize the Authenticate module so it's broken up by auth type
* Rather than simply storing a user_id in the session to consider someone logged in, I introduced a `Client` model which corresponds with each browser. I didn't call it Browser because we'll eventually consider iOS and Android app installs to be clients and very soon we'll create an client for API access
* Add a `Credential` model which is STI and move the existing auth to `PasswordCredential`, `GoogleCredential`, and `HttpHeaderCredential`
* Add an `Authentication` model. When credentials are verified an authentication is created for the current client, linking it to the credentials which just used. A client can be forced to logout from the server by deleting it's authentication.
* I went ahead and added a `GmailCredential` to ensure the architecture works well. I will follow up with a separate PR to add the APICredential


test/controllers/concerns/authenticate/by_http_header_test.rb
```diff
@@ -0,0 +1,117 @@
+require 'test_helper'
```
krschacht, line 1: @tmaier I did a big refactor of auth & credentials because I'm about to add a few more methods. This current PR here attempted to change no logic, I simply broke things into separate files and moved the auth details from the User model into a dedicated HttpAuthCredential model (likewise for Password and Google).

You don't need to review this whole PR, but could you review this test file to sanity check things? I'll flag one specific question below

krschacht: Also, maybe you can sanity check this locally to make sure I didn't break things. I'm pretty confident everything still works because we had good test coverage before and I just moved code around, but I don't have this HTTP Header auth situation setup in my local dev so it's not easy for me to do a final test of it.

PR #397, opened 2024-06-09
Localize the models

ceicke:
Start with localizing the messages from the models.

### DIFF
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -30,7 +30,14 @@ class Conversation < ApplicationRecord
   def self.grouped_by_increasing_time_interval_for_user(user)
     nav_conversations = user.conversations.ordered
 
-    keys = ["Today", "Yesterday", "This Week", "This Month", "Last Month", "Older"]
+    keys = [
+      I18n.t('models.conversation.groupings.today'),
+      I18n.t('models.conversation.groupings.yesterday'),
+      I18n.t('models.conversation.groupings.this_week'),
+      I18n.t('models.conversation.groupings.this_month'),
+      I18n.t('models.conversation.groupings.last_month'),
+      I18n.t('models.conversation.groupings.older'),
+    ]
     values = [
       nil,
       Date.current.beginning_of_day,
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -49,11 +49,11 @@ class Message < ApplicationRecord
   end
 
   def validate_conversation
-    errors.add(:conversation, I18n.t('errors.messages.invalid')) unless conversation.user == Current.user
+    errors.add(:conversation) unless conversation.user == Current.user
   end
 
   def validate_assistant
-    errors.add(:assistant, I18n.t('errors.messages.invalid')) unless assistant.user == Current.user
+    errors.add(:assistant) unless assistant.user == Current.user
   end
 
   def start_assistant_reply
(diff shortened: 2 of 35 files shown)


db/schema.rb
```diff
@@ -111,6 +111,13 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "llms", force: :cascade do |t|
+    t.string "name"
+    t.text "description"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
```
krschacht, line 120: Did you mean to add this table? I suspect this is from another branch. You can just revert this whole file.

ceicke: Fixed

app/models/conversation.rb
```diff
@@ -30,7 +30,14 @@ class Conversation < ApplicationRecord
...
 
-    keys = ["Today", "Yesterday", "This Week", "This Month", "Last Month", "Older"]
+    keys = [
+      I18n.t('models.conversation.groupings.today'),
+      I18n.t('models.conversation.groupings.yesterday'),
+      I18n.t('models.conversation.groupings.this_week'),
+      I18n.t('models.conversation.groupings.this_month'),
+      I18n.t('models.conversation.groupings.last_month'),
+      I18n.t('models.conversation.groupings.older'),
+    ]
```
krschacht, line 40: I think this is a good example of a view consideration sneaking it's way into the model. It'd be really nice if we could push this up to a partial somehow. 

One slightly hacky way to do it would be to turn these keys into symbols rather than strings—or even better, we turn these into the strings that represent the i18n name! So this array would be: `keys = [ ".today", ".yesterday" ...]` and then the view which iterates over these keys (somewhere we must have `.each do |header, _|` we do display it with `<%= t header %>`

I got this hack by realizing that views support [lazy lookups of keys](https://guides.rubyonrails.org/i18n.html#lazy-lookup)

app/models/message.rb
```diff
@@ -58,11 +58,11 @@ def create_conversation
   end
 
   def validate_conversation
-    errors.add(:conversation, 'is invalid') unless conversation.user == Current.user
+    errors.add(:conversation, I18n.t('errors.messages.invalid')) unless conversation.user == Current.user
   end
 
   def validate_assistant
-    errors.add(:assistant, 'is invalid') unless assistant.user == Current.user
+    errors.add(:assistant, I18n.t('errors.messages.invalid')) unless assistant.user == Current.user
```
krschacht, line 65: We can also get these with simply `errors.add(:assistant)`

config/locales/en.yml
```diff
@@ -1,31 +1,228 @@
-# Files in the config/locales directory are used for internationalization and
```
krschacht, line 1: Let's break these long files up like this:
https://guides.rubyonrails.org/i18n.html#organization-of-locale-files

I think that will make it much easier for us to manage them over time, especially when we have multiple languages and we may eventually break up translation tasks (or get GPT to help us :)

app/models/conversation.rb
ceicke: Yeah, I hear you. I would put any changes of the logic into a separate PR. I cannot add issues, so maybe it's something you could do so that it doesn't get forgotten :-)

krschacht: Done: https://github.com/AllYourBot/hostedgpt/issues/439

Issue #236, opened 2024-03-24
Allow creation of master API keys to be shared across users


krschacht:
Some people want to setup an instance of HostedGPT and let a couple special people use it. In this scenario, they don’t want everyone to have to enter their own API keys they want to let them just piggy back on some master API keys and foot the bill. In this scenario, my assumption was: we could read the API keys out of some central config.

If we implemented it this way then we would want to allow trusted people to register but prevent random people from registering. Maybe there needs to be an option to set a registration code? And, when I tell a trusted friend my registration code and they sign up, if that friend goes to their account settings how should the API key fields behave? Maybe the fields should still be there so they can input their own keys instead of using the master ones. Probably that’s the case.

Note: We’ll need some way to lock down user registration once this is implemented.

krschacht:
@olimart I mentioned this in a PR, but I’m pulling it out into a real issue. FYI in case you want to tackle it.

olimart:
Food for thoughts:

- Implement a new model `ApiKey` with type/provider `Claude, ChatGPT,...` & `shared` attribute.
- Update logic for current user key to become `user's key || shared ApiKey`

This implies:
- To keep fields on user settings to override shared key
- New `admin/api_keys` controller & views to manage keys
  - Add `admin` attribute to the very first user created to restrict keys management
- Migrate user keys to the new `ApiKey` model
- Eventually, new onboarding screen to input API key(s)

krschacht:
I like how you’re thinking. What you laid out makes sense as the place to end up. My inclination would be to get there in a least a couple PRs. One PR could lay down the core logic and just use an ENV as the fallback “shared key” and then a second PR could implement the new admin views & routes.

Want to take this one?

krschacht:
Nope, no progress on this one so you're safe to changes in this arena

matthewbennink:
This is interesting for our use case where we'll have a handful of students who are learning about AI chat interfaces. We'll be providing the API keys and yet we want to be sure we can keep usage within reasonable limits on a per-user basis.

As for cost tracking, is that something like keeping track of token usage per conversation? It seems like it'd be hard to do per message, especially if there's a moment where HostedGPT is unable to send all of the preceding messages, but instead sends summarized messages or perhaps only the last N messages of a conversation to keep the token count underneath the `max_tokens` limit. Or were you imagining tracking in terms of USD based on a pricing table?

[...]

PR #399, opened 2024-06-10
Allows for shared, default API keys.

matthewbennink:
If either of the default ENV keys are set, then the user will have access without needing to provide their own API key(s). I haven't added any tests and this doesn't protect anyone who adds these ENV variables while also allowing users to freely register. Is this simple enough?

Would it be worth adding anything to the README or might it be better to wait until the "feature" is more mature, perhaps allowing for the secrets to be managed within the database, for instance, since that would better support other LLMs outside of OpenAI and Anthropic.

Relates to #236.


app/services/chat_completion_api.rb
```diff
@@ -127,4 +127,4 @@ def self.model_token_limit(name)
       'gpt-3.5-turbo-instruct' => 4096
     }[name]
   end
-end
```
matthewbennink, line 130: My editor likes to add these newlines at the end of files. If that's problematic, I'll be sure to disable it and can revert this change.

krschacht: Do you mean your editor removes these newlines? If I'm reading this diff correctly it looks there was a newline after the "end" and now it's removed it?

It's fine, but I would love to standardize whitespacing handling somehow. Do you know if there a dotfile I can add to the project that will cause all IDEs to handle whitespace the same?

docker-compose.yml
```diff
@@ -26,6 +26,8 @@ services:
       - DATABASE_URL=postgres://app:secret@postgres/app_development
       - DEV_HOST=${DEV_HOST:-localhost} # Set if you want to use a different hostname
       - ALLOWED_REQUEST_ORIGINS=${ALLOWED_REQUEST_ORIGINS}
+      - DEFAULT_OPENAI_KEY={DEFAULT_OPENAI_KEY}
+      - DEFAULT_ANTHROPIC_KEY={DEFAULT_ANTHROPIC_KEY}
```
krschacht, line 30: I believe this needs $ it front of the curlies. At the last the one above it has that

app/services/chat_completion_api.rb
matthewbennink: I’ve had success with https://editorconfig.org/

matthewbennink: I’ve used https://prettier.io/ as well, but find it most useful for languages other than ruby.

krschacht (changes requested):
@matthewbennink Thank you for tackling this one. I think it's sufficient for a v1 but can we implement slightly differently:

- [ ] Having the conditional sprinkled throughout the app will inevitably break, especially with the various contributors we have let's instead override it at the model level then it'll "just work". So inside user.rb do `def openai_key; self.attributes["openai_key"] || ...`
- [ ] The one problem with this strategy is that a user can go to their settings and they'll be able to steal the API key so the way you can fix this is simply `settings/people/form` and explicitly adding the value to the text field, like this: `<%= user_fields.text_field :openai_key, value: person.user.attributes["openai_key"], ...` Now it will never pre-populate with the fallback value, it will only pre-populate with a user-specific value. AND users will still be able to override the default with their own key.
- [ ] Let's add a Feature for this, in addition to a setting. Meaning, in the options.yml file we can call the feature boolean: `default_llm_keys`, defaulting to false, and then add two Settings in here and then add the corresponding settings keys (`default_openai_key`, ...)
- [ ] No where in the app code should you reference ENV..., instead always reference Setting.default_openai_key. This makes it so that users can configure settings with `bin/rails credentials:edit` or with ENV variables

While you're at it, can you do me a favor... :) 
- [ ] We should follow this same pattern with ALLOWED_REQUEST_ORIGINS so add a Setting for that and update the code to use Setting.allowed_request_origins instead of ENV...

docker-compose.yml
krschacht: While you're adding these, can you do me a favor and make sure every Setting ENV is listed here? We're missing a few.

Also, add a comment such as:

```
  environment: -- Be sure to add this environment variable to options.yml
```

And do the reverse too, in the `options.yml` file on the settings line do:

```
  settings: -- Be sure to add these ENV to docker-compose.yml
```

This project is getting nuanced enough it's even hard for me to remember all the subtle connections so sprinkling these hints will help future me (and you) :)

app/services/chat_completion_api.rb
krschacht: Done. Thanks. I just did a commit in main that adds editorconfig and added it to the README

config/initializers/allowed_request_origins.rb
```diff
@@ -3,7 +3,7 @@
 #
 #   export ALLOWED_REQUEST_ORIGINS=https://myhost.com,https://myotherhost.com
 #
-if (allowed_request_origins = ENV['ALLOWED_REQUEST_ORIGINS'].to_s.split(',')).any?
```
matthewbennink, line 6: I tried changing this to `Setting.allowed_request_origins`, but `Setting` isn't loaded at the time this file is run. I'd expected it to be auto-loaded, but I'm not sure it's safe to do so, or even desired. Thoughts?

docker-compose.yml
```diff
@@ -23,10 +23,19 @@ services:
       context: .
       target: development
     environment:
+      # Be sure to add environment variables to config/options.yml
       - DATABASE_URL=postgres://app:secret@postgres/app_development
       - DEV_HOST=${DEV_HOST:-localhost} # Set if you want to use a different hostname
-      - ALLOWED_REQUEST_ORIGINS=${ALLOWED_REQUEST_ORIGINS}
       - OVERMIND_COLORS=2,3,5
+      - ALLOWED_REQUEST_ORIGINS
```
matthewbennink, line 30: Without interpolation, these ENV variables are only set inside the container if they're set in the host. If using interpolation, then an empty string is passed in to the Docker container, which isn't ideal.

test/models/user_test.rb
```diff
@@ -87,4 +87,13 @@ class UserTest < ActiveSupport::TestCase
       users(:keith).update!(last_name: nil)
     end
   end
+
+  test "it treats blank String API keys as nil to allow for default keys" do
```
matthewbennink, line 91: Does this feel OK to you? I realized that clearing the value inside the form would set it to an empty string in the model. I've handled it at the model layer by returning `nil` if we have a blank value stored, which appears to be handled correctly everywhere since `blank?` or `present?` are being used where needed.

matthewbennink:
@krschacht - I've made the requested changes. I wasn't sure if we needed the `features` ENV variables inside `docker-compose.yml` as well, but I _think_ we do since we may want to enable/disable features via the host env variables. Let me know what else needs done.

config/initializers/allowed_request_origins.rb
krschacht: Ohh, hm, that kind of makes sense b/c initializers must run early. In that case let’s just leave it as is. We have a few other “deep” things like DATABASE_NAME and stuff like that which are ENV only so maybe there is a clear distinction between server level settings and app-level settings.

docker-compose.yml
krschacht: Nice, that one is new to me.

test/models/user_test.rb
krschacht: Yes, that seems like the right solution. Good catch.

krschacht (approved):
This looks great. Thanks for taking this one on!

I like to do merges in the morning in case there is surprise breakage so I’ll merge tomorrow AM.

test/models/user_test.rb
```diff
@@ -87,4 +87,13 @@ class UserTest < ActiveSupport::TestCase
...
+    stub_features(default_llm_keys: false) do
+      user = users(:keith)
+      user.update(openai_key: '', anthropic_key: ' ')
```
krschacht, line 94: @matthewbennink after merging in I noticed this is missing a bang. The update can fail and then this test mistakenly passes. Years ago I automatized almost always doing using bang on update and save since we typically expect it to succeed and want it to fail loudly if it doesn't (and controllers handle exceptions gracefully).

I'll quickly fix this in main, so just an FYI for you

PR #400, opened 2024-06-10
Fix database.yml password for production

ceicke:
There was always an = sign appended the production database password.


krschacht (approved):
Ugh, I can't believe I let that one slip in. I'm surprised that didn't blow up more stuff! I get lots of production environments just ignore that password field.

PR #407, opened 2024-06-11
Add support for API clients

krschacht:
A request will be authenticated if there is a proper Bearer token in the request and JSON should be returned for all endpoints.

Automatically returning a response for actions that end with redirect_to was a little tricky. The way this was implemented was to assign a special significance to `redirect_to` with a `status: :see_other`. When this status is present, we take this as a sign of success and we render a simple success JSON response.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,24 @@
 class ApplicationController < ActionController::Base
   include Authenticate
 
+  def default_render
+    if api_request?
+      json_payload = user_defined_ivars.map { |i| [ i.to_s[1..], instance_variable_get(i) ] }.to_h
+      render json: json_payload, status: :ok
+    else
+      super
+    end
+  end
+
   private
+
+  def api_request?
+    request.format.json?
+  end
+
+  def user_defined_ivars
+    instance_variables.select { |i| !i.to_s.starts_with?('@_') }
+  end
 
   def ensure_manual_login_allowed
     return if manual_login_allowed?
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,7 @@
 class ApplicationController < ActionController::Base
   include Authenticate
 
+  skip_before_action :verify_authenticity_token, if: :api_request?
   before_action :set_system_ivars
 
   def default_render(*args)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,6 +1,8 @@
 class ApplicationController < ActionController::Base
   include Authenticate
 
+  before_action :set_system_ivars
+
   def default_render
     if api_request?
       json_payload = user_defined_ivars.map { |i| [ i.to_s[1..], instance_variable_get(i) ] }.to_h
@@ -12,12 +14,20 @@ class ApplicationController < ActionController::Base
 
   private
 
+  def set_system_ivars
+    @system_ivars = public_ivars
+  end
+
+  def public_ivars
+    instance_variables.select { |i| !i.to_s.starts_with?('@_') }
+  end
+
   def api_request?
     request.format.json?
   end
 
   def user_defined_ivars
-    instance_variables.select { |i| !i.to_s.starts_with?('@_') }
+    public_ivars - @system_ivars
   end
 
   def ensure_manual_login_allowed
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -41,7 +41,7 @@ class ApplicationController < ActionController::Base
   end
 
   def user_defined_ivars
-    public_ivars - @system_ivars
+    public_ivars - @system_ivars - ["system_ivar"]
   end
 
   def ensure_manual_login_allowed
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,12 +3,26 @@ class ApplicationController < ActionController::Base
 
   before_action :set_system_ivars
 
-  def default_render
+  def default_render(*args)
     if api_request?
       json_payload = user_defined_ivars.map { |i| [ i.to_s[1..], instance_variable_get(i) ] }.to_h
       render json: json_payload, status: :ok
     else
-      super
+      super(*args)
+    end
+  end
+
+  def redirect_to(*args)
+    if api_request?
+      options = args.extract_options!
+      status = options[:status]
+
+      if status == :see_other
+        render json: options.merge(redirect_to: args.first), status: :ok
+        return
+      end
+    else
+      super(*args)
     end
   end
 
--- a/app/controllers/concerns/authenticate/by_bearer_token.rb
+++ b/app/controllers/concerns/authenticate/by_bearer_token.rb
@@ -24,5 +24,6 @@ module Authenticate::ByBearerToken
   def render_unauthorized
     self.headers['WWW-Authenticate'] = 'Token realm="Application"'
     render json: { error: "Authentication Bearer token was invalid" }, status: :unauthorized
+    nil
   end
 end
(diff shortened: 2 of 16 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,8 +1,25 @@
 class ApplicationController < ActionController::Base
   include Authenticate
 
+  def default_render
```
matthewbennink, line 4: This won't be called if we redirect within the action itself, so only _some_ actions will get JSON payloads in their response, namely those that would otherwise render from an HTML template.

app/controllers/application_controller.rb
```diff
@@ -1,8 +1,25 @@
...
+    end
+  end
+
   private
 
+  def api_request?
+    request.format.json?
+  end
+
+  def user_defined_ivars
```
matthewbennink, line 19: There were two ivars that I didn't expect to see - `rendered_format` and `marked_for_same_origin_verification` - but otherwise it seemed to work great!

app/controllers/concerns/authenticate/by_bearer_token.rb
```diff
@@ -0,0 +1,28 @@
+module Authenticate::ByBearerToken
+
+  private
+
+  def find_client_by_bearer_token
+    authenticate_with_http_token do |double_token, options|
+      client, client_token = parse_double_token(double_token)
```
matthewbennink, line 7: If the `Authentication` isn't present for any reason, then this returns `nil` for `client` and an error is raised when trying to compare `client.token` (client being `nil`) with `client_token`. We could change the condition to be `if client && client_token && ...` perhaps.

app/models/current.rb
```diff
@@ -6,7 +6,7 @@ class Current < ActiveSupport::CurrentAttributes
   def self.initialize_with(client: nil)
     self.client = client
 
-    if client&.authenticated?
+    if client&.is_a?(Client) && client&.authenticated?
```
matthewbennink, line 9: This could be simplified to `if client.is_a?(Client) && client.authenticated?`. `nil.is_a?(AnythingButNilClass)` will be `false` (assuming no meta-programming).

app/controllers/application_controller.rb
krschacht: Good point. I have an idea for how to detect which ivars are created from rails internals and exclude those

app/models/current.rb
krschacht: Interesting, that .nil syntax is new to me. I’m intrigued.

But I don’t quite follow your train of thought. What’s the thing you’re simplifying?

app/controllers/application_controller.rb
krschacht: I figured out a way to ONLY include user created instance variables in the response. A simple before_action to remove anything that existed before the action ran.

app/controllers/application_controller.rb
krschacht: I figured out a way to ONLY include user created instance variables in the response. A simple before_action to remove anything that existed before the action ran.

matthewbennink:
```
{
  "system_ivars": [
    "@rendered_format",
    "@marked_for_same_origin_verification"
  ],
```
😂 - we'll need to remove `@system_ivars` from the list as well, but it worked great otherwise!

app/controllers/concerns/authenticate/by_bearer_token.rb
krschacht: Good catch. I fixed it and added a test to catch it.

app/models/current.rb
krschacht: Good catch, I just removed the &. For awhile I thought this nil check was all I needed but then I ran into one weird situation where something other than a client was passed into here. To be honest, maybe I should have caught that in an alternate way but my fix was to confirm that client is indeed a client. But with that in place, the nil check was no longer needed.

thanks for pointing out

krschacht: Actually, I take that back. I should really fix this properly. The fact that a non-Client was being passed in is a deeper code small. I'm fixing the type mismatch at the root so then I don't need this is_a?(Client) check

krschacht:
@matthewbennink Give this one another look. I think I was actually able to solve those good points that you raised (ivars and redirect). Is this a good overall solution now?

app/controllers/application_controller.rb
krschacht: Ack! Yes :) Done

matthewbennink:
Yea, I think this makes sense overall. We can request information and generally get everything that's needed for the HTML view, and if we mutate anything, we'll be given a redirect URL that will give us the information we may need, or another redirect perhaps.

Thinking through creating messages, if we send a POST to `/assistants/:assistant_id/messages`, we'll get a redirect to `/conversations/:conversation_id/messages?version=1` and that will provide the full list of messages. Our message is likely the most recent "user" message or perhaps it's the `streaming_message`, but maybe that doesn't matter so much since we'll have the full list and can reconcile as needed with some local state.

I did a quick search for where instance variables are set and, apart from the `User`, I don't see any that seem particularly dangerous if they're leaked. The assumption is that if the user has access to the action, they have access to any data within instance variables we set.

For `User`, it's providing the OpenAI and Anthropic keys even if they're set via the ENV vars, so that might be something to think through. The `@user` is only set in a couple places, once in `users#update` where we redirect afterwards, so it won't be leaked with the current logic, and the other in `authentications#create` where it doesn't seem necessary to use an instance variable at all; I think we could get away with just `user` since it's only referenced within the controller method, not the `new` view template.

More importantly, the `@user` variable is set even though the user isn't authenticated yet, so it _may_ be possible to send a request to `/login` and get a response back for the person with a given email even if the wrong password is supplied. It doesn't work right now because we render the `new` template, so `default_render` isn't called and instead we get this error:
```
base-1      | web      | ActionView::MissingTemplate (Missing template authentications/new, application/new with {:locale=>[:en], :formats=>[:json], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby]}.
```

Anyway, if we just remove the `@user` instance variable from `authentications#create`, I think this will be safe. We could also adjust the `api_request?` method to only return `true` if the client is authenticated, e.g., `request.format.json? && Current.client.authenticated?`.

matthewbennink:
Oh - the only other thing I forgot to note that we'll need to deal with is skipping `verify_authenticity_token` when it's an API request. I haven't figured that one out yet. It may just be another case of overriding a method and calling `super` for non-API requests, but it could happen in a separate PR.

krschacht:
> If we just remove the `@user` instance variable from `authentications#create`, I think this will be safe.

I'm loving the thoroughness. Great catch on @user. I just removed those two unnecessary setting of them.

However, it really seems wrong that we should need to be concerned with a user seeing the @user object. The fact that we're even having to ask ourselves this suggests that the `openai_key` implementation is wrong. One some level I already realized this b/c I didn't want to "leak" these keys via a user's account settings, and now looking back I think my suggested fix was wrong. We fixed it at the **form** level but really we should have fixed it at the model level.

I think the right fix is: we no longer hack the openai_key attribute. Instead, on the user model we add a method on the model like:

```
def preferred_openai_key
  openai_key || ENV...
end
```

> We could also adjust the `api_request?` method to only return `true` if the client is authenticated, e.g., `request.format.json? && Current.client.authenticated?`.

I don't quite get this because every controller action already has a before_filter concerning authentication. There's a default app-wide in application_controller and some controllers override that for various actions.

If we also add an authentication check in the redirect_to, which is at the bottom, then we could end up with a discrepancy between the auth check at the start of an action and the end. So that doesn't seem right.

@matthewbennink Is there a situation I'm not thinking of where we need redirect_to to check it again?


> authenticity_token

I think it's just `skip_before_action :verify_authenticity_token, if: :api_request?` I added that to the application_controller. I'm not quite sure how to test it but I'm pretty sure it's that simple

matthewbennink:
> I don't quite get this because every controller action already has a before_filter concerning authentication.

I think I had the unauthenticated routes in mind at the time, such as signing in. But we're no longer setting instance variables for any of those now that `@user` has been changed, so I think we're set. If we ever do set an instance variable prior to authenticating, it could potentially be accessed, but it feels unlikely and also wouldn't be associated with any user.

> I'm pretty sure it's that simple

Yep

PR #409, opened 2024-06-11
Ensure you can only attach images to composer

krschacht:


drnic:
Why are we restricting to images only?

Is this because they are implicitly easy to display inside the chat window?

Is the requirement to support all other document types only that we have a way to visualise what documents are included in the chat?

Here is chatgpt's visualisation of PDF docs in a chat:
IMG_2117

drnic:
Bonus question: are there concerns/reasons for `<%= document_form.file_field :file, accept: "image/*", class: "hidden", data: { image_upload_target: "file" } %>` to only allow a single file? The data model looks to accept 1+ Documents per Message.

krschacht:
@drnic There is no good reason for only a single file. When I first implemented the feature, I added support for multiple files deep in the domain model but when I got to the UI I was eager to have the feature finished so I cut scope and only made it possible to attach one image. I believe adding support for multiple images is just an issue of getting the front-end to elegantly handle it. After that it should "just work" although there may be some debugging.

On the question of PDFs, that's a much more complex question. Supporting those (and DOCs) is more complex.

The OpenAI API is a little bit messy. Their main API, the "chat api" has only supported image attachments for the longest time. However, the new 4o model just added support for audio attachments in the last 2 weeks so they expanded support a little bit:
https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages

For quite some time now, when LLMs parse PDFs (or any other docs) it's RAG (retrieval augmented search). So you parse the document into a bunch of chunks, save those chunks in their original form and vectors of those, and then you add a function the LLM can call to retrieve relevant information (typically this function call does a combination of vector search and old school keyword search).

When OpenAI added support for PDFs to the chatgpt interface, they introduced a completely new API called the Assistants API back in 2023. It's quite a bit more complex to build apps against. The original API is stateless. We store all the state within the rails app and pass it with each message request. But the Assistants API is stateful. API call has to be made to create an assistant (with system prompt as part of it). A separate API call to add a message to the conversation history — so you don't submit the full history with each new message.

The main advantage of this was that it was more efficient token processing because you aren't resubmitting the long context history. Also, this API includes OpenAI-created tools: specifically document search (i.e. RAG) and code interpreter (https://platform.openai.com/docs/assistants/tools). But this API did not support images within messages for the longest time. I believe 3 months ago they finally added support for images so it finally has full feature parity with the original API plus these bonus things.

But... during all this time, all the other competitors copied OpenAI's original API. Anthropic, Groq, and all the local models. So it's quite easy for the app to support multiple services. Plus, a few months ago OpenAI added "caching" to their old API so now it won't charge you for all the conversation history with each subsequent message. And just last month, Anthropic added support for PDF attachments with their stateful API:
https://docs.anthropic.com/en/docs/build-with-claude/pdf-support

It looks like they are not doing RAG but instead just extract all the contents of the PDF and add it to the conversation history. Will OpenAI now add support for this to their old API? It's really unclear if OpenAI is looking to deprecate their old API and move to the new one? I suspect that was their original plan, but I think they've dragged their feet so long that now waaaay to many companies are old their original API so I kind of suspect they'll eventually phase out the "new" Assistants API as a failed experiment. But that's just a weak hunch.

To support PDFs in HostedGPT, the options are:

1) Keep using the old "chat API" so that supporting multiple services is easy and implement our own PDF-extraction feature which dumps the full document into the conversation history (and disable this for Anthropic's Claude since it does it automatically).

2) Implement a RAG system to support, which would potentially support adding a ton of documents to an assistant. Langchainrb has one which I poked around a bit, but RAG is complicated and can require quite a bit of tuning. If we did this, all services could support asking questions about documents.

3) Move OpenAI over to the Assistants API for OpenAI but refactor the backend so it continues to work with the old API format for all the other services.

This space moves too fast... :)

drnic:
Ah, because something works in chatgpt.com it does not necessarily exist in any of their APIs. Bah.

> PDF-extraction feature which dumps the full document

This was the approach I was going to take initially as I assumed it 1. would use fewer tokens; 2. would work against the most LLM models.

Thanks for the generous summary of the history + ideas for implementations.

PR #411, opened 2024-06-13
Add missing scope and method to Assistant

nwarwick:
Signup wasn't working for me and it appeared to be due to a missing scope on the Assistant model. I also ran into an issue where a missing method on the model was causing an error.

Before:
Screenshot 2024-06-13 at 11 49 51 AM

Screenshot 2024-06-13 at 11 58 47 AM

### DIFF
--- a/app/models/assistant.rb
+++ b/app/models/assistant.rb
@@ -17,6 +17,7 @@ class Assistant < ApplicationRecord
   validates :name, presence: true
 
   scope :ordered, -> { order(:id) }
+  scope :not_deleted, -> { where(deleted_at: nil) }
 
   def initials
     return nil if name.blank?
--- a/app/models/assistant.rb
+++ b/app/models/assistant.rb
@@ -31,4 +31,8 @@ class Assistant < ApplicationRecord
   def to_s
     name
   end
+
+  def not_deleted?
+    deleted_at.nil?
+  end
 end


krschacht:
@nwarwick I really appreciate you helping debug this and tee'ing up a PR! But I can't quite figure out why this was causing problems for you as that scope and helper method should already exist. It's a little buried as it comes from the `Scopes::AutoGenerated` mixin within ApplicationRecord. I just auto-create a bunch of scopes and helpers for datetime fields across all the models (e.g. `deleted_after()`, `deleted_on()` in addition to `not_deleted`)

I just pulled down main, opened a console, and did this as a sanity check:

```
[2] pry(main)> Assistant.not_deleted.length
  Assistant Load (2.3ms)  SELECT "assistants".* FROM "assistants" WHERE (assistants.deleted_at is null or assistants.deleted_at > '2024-06-13 16:31:26.979242')
=> 8
[3] pry(main)> Assistant.not_deleted.last.not_deleted?
  Assistant Load (2.1ms)  SELECT "assistants".* FROM "assistants" WHERE (assistants.deleted_at is null or assistants.deleted_at > '2024-06-13 16:32:53.667493') ORDER BY "assistants"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> true
```

On your machine, could you try reverting the change you made, starting a fresh console and replicating this same test?

I wonder if your app was running, you pulled down some new changes from git without restarting the server, and maybe the way I'm doing those scopes is not working with rails auto-reloading? That's my best guess right now. But after you run my same test, paste the output if it doesn't work b/c I might see a clue.

nwarwick:
Yep looks like something was just out of sync @krschacht. I re-cloned the repo and made sure everything was up to date and now it seems to be working. My apologies! Thanks for looking into it regardless.

krschacht:
No problem at all! I'm glad it was that simple. I'm happy every time someone is using the project because it helps us all discover more issues sooner and identify opportunities to improve it.

nwarwick:
I'm really liking exploring it! Learning lots of new tricks like this active record mixin.

PR #414, opened 2024-06-14
renames API key methods on User to avoid keys leaking into JSON objects

matthewbennink:
From feedback on #407, renames the methods to avoid leaking the keys. The OpenAI key injected into JS for the voice feature does not fallback to a default API key.


krschacht (approved):
Thanks @matthewbennink , merging now. I also merged my big Credentials API PR so that's in main.

PR #424, opened 2024-06-18
Attach images to conversations as URLs

krschacht:


jlvallelonga:
from: https://docs.anthropic.com/en/docs/build-with-claude/vision#can-claude-read-image-urls

image

so just openai for now

config/application.rb
```diff
@@ -37,10 +37,17 @@ class Application < Rails::Application
...
+    config.app_url_protocol = Setting.app_url_protocol
+    config.app_url_host = Setting.app_url_host
+    config.app_url_port = Setting.app_url_port
+    config.app_url = "#{Setting.app_url_protocol}://#{Setting.app_url_host}:#{Setting.app_url_port}"
+
     # Active Storage
     if Feature.cloudflare_storage?
       config.active_storage.service = :cloudflare
     else
+      puts "using database service"
```
jlvallelonga, line 50:
```suggestion
```

lib/active_storage/service/postgresql_service.rb
```diff
@@ -148,10 +148,15 @@ def url_helpers
...
-        { host: ActiveStorage::Current.host }
+        url_opts = ActiveStorage::Current.url_options
+        return url_opts if url_opts.is_a?(Hash)
       end
+
+      return {
+        protocol: Rails.application.config.app_url_protocol,
+        host: Rails.application.config.app_url_host,
+        port: Rails.application.config.app_url_port,
+      }
```
jlvallelonga, line 159: fixes this: https://github.com/AllYourBot/hostedgpt/pull/398#issuecomment-2168135853

the problem is that the `ActiveStorage::Current.url_options` are set only during a request lifecycle (with ActiveStorage::SetCurrent) and workers don't run within the context of a request. This fix also allows you to get the url from within the rails console.

see: https://github.com/rails/rails/blob/main/activestorage/app/controllers/concerns/active_storage/set_current.rb
and: https://github.com/rails/rails/issues/40855#issuecomment-2350744657
and: https://github.com/rails/rails/pull/42520

krschacht: on dang, nice debugging on this. i couldn’t figure out what the difference was within the worker

krschacht:
Oh dang, I didn’t realize that Claude actually requires base64 encoding. I thought when I saw in the OpenAI docs that they suggest passing in the URLs that this meant it was the preferred method and I just assumed that would be the same for the other models.

That decreases the priority of this then. If it’s not too much more work, I think it’s still worth getting this for openai since that’s the most used model. But if you run into any blockers then we can always elect to merge in the improvements you have and defer the rest. I’m just sharing this aloud since I’m not sure where you’re at with this or if you’ve hit a blocker.

krschacht:
@jlvallelonga our production setup is using the credentials file. I did a recent commit on the other repo to add that. You can edit it with `bin/rails credentials:edit --environment production` and it lives in the repo. Only the encryption key is in the production env.

For github actions we should just add env to the rubyonrails.yml file

jlvallelonga:
ah I just merged this thinking it was approved. Let me know if you want any changes and I can make those in another branch @krschacht

krschacht:
@jlvallelonga no problem, I'll give it a close look right now

krschacht:
New PR: https://github.com/AllYourBot/hostedgpt/pull/532

PR #425, opened 2024-06-19
341 ManageLLMs (III)

stephan-buckmaster:
This is from discussions in the 341-manage-llms-2 pull request, https://github.com/AllYourBot/hostedgpt/pull/389

It extends the 341-manage-llms-2 changes, by removing the concept of a system-wide language model, by creating copies of the system-wide language model records for each user, then removing the system-wide records themselves.

This then makes it confusing to users to allow configuring the OpenAI/Anthropic api-key in two places:
* in their user settings,
* as a token of their "OpenAI" APIService

This change also removes this, and so it turns into a larger change. For example, the implementation of the feature  for default openai/anthropic api-keys gets moved.

All the tests are passing, but I haven't added tests of actual new behaviour yet.


stephan-buckmaster:
Merged in the latest from upstream (via the 341-manage-llms-2 branch). I feel it will be hard to maintain the process of managing the users' language models through migrations. 

We can update the "best" models with simple one-liners. Same for the initial configs. But going into the user configuration for existing users feels unusual. Possibly introduce a "second-best"? They really can do this themselves too.

krschacht:
I merged this PR back into (II) just because I found it easier to do a final review in one place. Closing. I'll add comments over there.

PR #446, opened 2024-06-30
More debugging of voice but specific to Chrome on Pixel devices

krschacht:
The voice code currently enables a Transcriber and a Microphone at the same time. The mic was for specific volume details (and the visual effect) whereas the transcriber is the core.

Chrome on my Pixel does not allow using both at once. This code addresses that issue and some other rough edges at the same time. Key changes:

* Disable Microphone and rely solely on Transcriber
* Improve AI's system message with time & date so it can answer them independently
* Improve on some copy on the Google Tools part of account settings
* runAfter and runEvery now have a method to run() them before their time is due
* There is still some glitchiness when there are only 2 messages on the page. It's difficult to determine whether we just came from creating a new conversation and both messages are already there because the page was slow to load (and the AI was fast in generating a response), or because the conversation was freshly opened at the point of 2 messages and the user is just submitting a new question. The reason this is an issue is because in the later case, sometimes the voice will repeat the previous response while waiting for the new reply. I attempted to improve this a bit through tiny changes to logic, extra logging, and improved cancelling of the previous speaking queue. In other words, in the later case it will still queue up one speaking queue and I rely on the second response coming quickly enough to cancel the first.


app/javascript/blocks/interfaces/speaker_interface.js
```diff
@@ -14,7 +14,6 @@ export default class extends Interface {
 
   new() {
     $.audioService = new AudioService
-    //Microphone.attach($.audioService.player)
```
krschacht, line 17: Related to the visual effect

app/views/messages/_message.html.erb
```diff
@@ -63,7 +63,7 @@ end
         ><%= format_for_copying message.content_text %></div>
         <div
           class="hidden"
-          data-speaker-target="text <%= message.assistant? && "assistantText" %>"
+          data-speaker-target="text <%= message.assistant? ? "assistantText" : "userText" %>"
```
krschacht, line 66: I actually don't think this is helping. I thought I could use the presence userText to distinguish between the two tricky cases but I'm almost certain it's not helping. I should eventually revert this when I come up with a better strategy.

PR #456, opened 2024-07-10
Better support local dev - local DB documentation & env vars

draknor:
Update README to better document db setup for local dev, and update database.yml to support local username+password.

NOTE: This *does* change the default username + password for local (non-Docker) dev, so there's a potential compatibility issue there. Not sure if there's a good way around it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,12 +229,10 @@ Every time you pull new changes down, kill docker (if it's running) and re-run:
 HostedGPT requires these services to be running:
 
 - Postgres ([installation instructions](https://www.postgresql.org/download/))
-  - Default database username: `app`
-  - Default database password: `secret`
-  - Default dev database: `hostedgpt_development`
-  - Default test database: `hostedgpt_test`
-  - You can override via environment variables - see `config/database.yml`
-  - *Note that the docker compose workflow uses `DATABASE_URL` instead, which overrides anything in `config/database.yml`*
+  - Use environment variable `DATABASE_URL` to specify the connection parameters
+  - *Make sure to set that environment variable in your setup!*
+  - Recommended (development): `DATABASE_URL=postgres://app:secret@localhost/hostedgpt_development`
+  - Recommended (test): `DATABASE_URL=postgres://app:secret@localhost/hostedgpt_test`
 - rbenv ([installation instructions](https://github.com/rbenv/rbenv))
 - ImageMagick (`brew install imagemagick` should work on Mac )
 
--- a/README.md
+++ b/README.md
@@ -228,11 +228,8 @@ Every time you pull new changes down, kill docker (if it's running) and re-run:
 
 HostedGPT requires these services to be running:
 
-- Postgres ([installation instructions](https://www.postgresql.org/download/))
-  - Use environment variable `DATABASE_URL` to specify the connection parameters
-  - *Make sure to set that environment variable in your setup!*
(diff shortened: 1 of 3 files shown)


draknor:
Ha, I guess I could just NOT set defaults for username + password - seems so obvious in hindsight! I'll have to test that - I haven't setup my local dev DB to allow anonymous access.

krschacht:
Even if you have a username and password you can provide that within the
DATABASE_URL

krschacht:
@draknor let me know what you think of just setting DATBASE_URL instead of this? The format is postgres://username:password@localhost/hostedgpt_development

(Password and/or username can be left off)

This is a rails convention and the production configs use this to set the database details.

draknor:
Yep - I revised my approach to just focus on DATABASE_URL -- how's the new commit look?

krschacht:
@draknor Did you need to change anything in database.yml in order to get it to work? I could be wrong, but I thought that DATABASE_URL would just work as is. I think all we need to do is elaborate within the README

README.md
```diff
@@ -229,6 +229,10 @@ Every time you pull new changes down, kill docker (if it's running) and re-run:
 HostedGPT requires these services to be running:
 
 - Postgres ([installation instructions](https://www.postgresql.org/download/))
+  - Use environment variable `DATABASE_URL` to specify the connection parameters
+  - *Make sure to set that environment variable in your setup!*
+  - Recommended (development): `DATABASE_URL=postgres://app:secret@localhost/hostedgpt_development`
+  - Recommended (test): `DATABASE_URL=postgres://app:secret@localhost/hostedgpt_test`
```
krschacht, line 235:
```suggestion
- Postgres (`brew install imagemagick` or other [install instructions](https://www.postgresql.org/download/))
  - By default postgres will create a default user and things should just work, but if you configure a specific username and password then set the environment `DATABASE_URL=postgres://username:password@localhost/hostedgpt_development` (replacing username & password with your values, optionally you can change the database name)
```

draknor: @krschacht I changed database.yml to be more explicit that we're using DATABASE_URL, per the recommendation in the Rails Guide [3.21 Connection Preference](https://edgeguides.rubyonrails.org/configuring.html#connection-preference)

(So yes - you are correct that DATABASE_URL works as-is; editing database.yml is just to reduce developer confusion).

New commit added!

krschacht: @draknor Got it. Yes, you’re right, I think it’s good that we explicitly set `url:` as you’ve done. That makes it easier to discover. In fact, let’s even add a comment that gives an example of the format for the URL.

The thing I worry about is if any users have made use of the ENV which are being removed from database.yml, we might break their setup with this release. Reading the Rails Guide that you linked to, it sounds like details from DATABASE_URL will take precedence over any explicitly specified things. I’m just inclined to add back all the red removals from database.yml (the default `host` and the default `database` and what not).

README.md
```diff
@@ -228,7 +228,8 @@ Every time you pull new changes down, kill docker (if it's running) and re-run:
 
 HostedGPT requires these services to be running:
 
-- Postgres ([installation instructions](https://www.postgresql.org/download/))
+- Postgres (`brew install imagemagick` or other [install instructions](https://www.postgresql.org/download/))
```
jlvallelonga, line 231:
```suggestion
- Postgres (`brew install postgresql@16` or other [install instructions](https://www.postgresql.org/download/))
```

Issue #458, opened 2024-07-14
Fix bin/dev overmind check in Render environment


krschacht:
We just made an update to bin/dev to confirm that Overmind install worked. However, this is failing for Render.

We need to ONLY do the bin/dev overmind extra checking if RENDER environment variable does not equal true.

PR #457, opened 2024-07-12
Workaround: Overmind installing wrong binaries for aarch64-linux

htruong:
When we run on aarch64-linux*, overmind downloads and installs the wrong binary for the architecture. Detect if that's the case, and then download the correct binaries.

This originally came up when installing the app on a Raspberry Pi. An issue has been opened with the Overmind project: https://github.com/DarthSim/overmind/issues/185

PR #459, opened 2024-07-14
Update bin/dev on Render

krschacht:
After the recent PR #457, Render is now reporting an error `Workaround: Overmind installing wrong binaries for aarch64-linux`

This PR attempts to fixes the issue #458

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -11,7 +11,7 @@ fi
 # Remove once #185 is fixed for overmind
 # https://github.com/DarthSim/overmind/issues/185
 RUBY_PLATFORM=$(ruby -e "puts Gem::Platform.local")
-if [[ $RUBY_PLATFORM == aarch64-linux* ]]; then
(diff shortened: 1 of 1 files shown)


htruong:
That's funny, why did the `$RUBY_PLATFORM == aarch64-linux*` match in the Render case?

On my laptop:

```
#!/bin/bash

RUBY_PLATFORM=$(ruby -e "puts Gem::Platform.local")

if [[ $RUBY_PLATFORM == aarch64-linux* ]]; then
        echo "$RUBY_PLATFORM is an aarch64-linux platform"
else
        echo "$RUBY_PLATFORM is not aarch64-linux"
fi
```

```
$ ./foo.sh
x86_64-linux is not aarch64-linux
```

Why did it match, did Render actually use aarch64-linux?

krschacht:
I was also confused by this. But notably, even if they are using it, I would have expected this change to have been a positive one for Render but instead bin/dev was exiting with an error code. It’s a bit of a sloppy fix but it worked to just exception it. There continue to be a surprising number of little edge cases for all these different environments! :)

Issue #402, opened 2024-06-11
Add token cost tracking


krschacht:

matthewbennink:
The `estimated_token_count` on the message is useful, but if I understand correctly, we'll need to also add up all of the token counts of prior messages, assuming all of them are sent.

For example, suppose we send 3 messages, each with 100 tokens, and we get 3 replies, each with 100 tokens, and the system message is 200 tokens. Our first message will be 200+(100+100)=400 tokens, our second message will be 200+(100+100)+(100+100)=600 tokens, and our third message will be 200+(100+100)*3=800 tokens. Given the [OpenAI pricing](https://openai.com/api/pricing/) for GPT-4o of $5 for 1M input tokens and $15 for 1M output tokens, and with our 500 input tokens and 300 output tokens, we'll get some cost. **Is this your understanding?**

As well, when we have very long conversations, there will be a moment when some of the preceding messages may be dropped or summarized to reduce the token usage, or at least get it within the context limit. In other words, the number of tokens in the latest message may not help us fully work out the cost of the request. It's almost like we need to keep track of each individual API request and the number of (estimated) input and output tokens.

If the API can give us the actual number of input tokens,  even better, which _seems_ possible with OpenAI if we pass the `include_usage` [streaming option](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options). I don't see any mention of `include_usage` inside the OpenAI ruby gem git repo ([search query](https://github.com/search?q=repo%3Aalexrudall%2Fruby-openai+include_usage&type=code)), but might be something that can be surfaced there.

krschacht:
@matthewbennink Yes, that's why I was thinking we update estimated_price twice. When we are generating a new message, we pass a newly created message into get_next_message_job (it's persisted to the db already) which, in turn, passes it into ai_backend.

I think the moment that ai_backend is sending it's request to the API, which includes all of the previous messages in the conversation, we can add up the tokens and save the preliminary estimated_cost on the blank message.

Then when the response comes back to get_next_message_job, we do a final message.save and we can do one more token cost estimate and add it onto the estimated_price we previously calculated.

I didn't know about include_usage, that's cool! The OpenAI gem just passes the hash of params that we send straight on to OpenAI, so it should be supported.

matthewbennink:
[...]

PR #460, opened 2024-07-14
Draft: Add token counting per message

lumpidu:
Implement token counting per message by using the returned usage stats from the LLM API backend. Anthropic sends token stats already by default. As for OpenAI, the stream needs to be first configured to return those.

- Add migration for new Message attributes input_token_count, output_token_count
- Add concern Message::TokenCount for Message model
- Implement capturing of returned token count from Anthropic & OpenAI and set Message attributes inside the async API-specific jobs. For OpenAI, configure the stream to return usage stats
- Add basic usage dropdown-menu beneath last message that shows the number of input/output tokens when clicked on the "$" sign. This is just a quick demo for the functionality and will likely be changed to something else

This MR is related to #402

### DIFF
--- a/app/services/ai_backend/anthropic.rb
+++ b/app/services/ai_backend/anthropic.rb
@@ -28,6 +28,15 @@ class AIBackend::Anthropic < AIBackend
 
     response_handler = proc do |intermediate_response, bytesize|
       chunk = intermediate_response.dig("delta", "text")
+
+      # input and output tokens are sent in different responses
+      if (input_tokens = intermediate_response.dig("message", "usage", "input_tokens"))
+        @message.input_token_count += input_tokens
+      end
+      if (output_tokens = intermediate_response.dig("message", "usage", "output_tokens"))
+        @message.output_token_count += output_tokens
+      end
+
       print chunk if Rails.env.development?
       if chunk
         stream_response_text += chunk
--- a/app/services/ai_backend/open_ai.rb
+++ b/app/services/ai_backend/open_ai.rb
@@ -35,6 +35,7 @@ class AIBackend::OpenAI < AIBackend
         tools: Toolbox.tools,
         stream: response_handler,
         max_tokens: 2000, # we should really set this dynamically, based on the model, to the max
+        stream_options: {include_usage: true}
       })
     rescue ::Faraday::UnauthorizedError => e
       raise ::OpenAI::ConfigurationError
@@ -53,6 +54,12 @@ class AIBackend::OpenAI < AIBackend
     proc do |intermediate_response, bytesize|
       content_chunk = intermediate_response.dig("choices", 0, "delta", "content")
       tool_calls_chunk = intermediate_response.dig("choices", 0, "delta", "tool_calls")
+
+      # input and output tokens are sent in the same response
+      if (input_tokens, output_tokens = intermediate_response["usage"]&.values_at("prompt_tokens", "completion_tokens"))
+        @message.input_token_count += input_tokens
+        @message.output_token_count += output_tokens
+      end
 
       print content_chunk if Rails.env.development?
       if content_chunk
--- a/app/views/messages/_message.html.erb
+++ b/app/views/messages/_message.html.erb
@@ -245,6 +245,24 @@ end
                 <% end %>
               </menu>
             </div>
+            <% if last_message %>
+              <% input_tokens = conversation.messages.sum(:input_token_count) %>
+              <% output_tokens = conversation.messages.sum(:output_token_count) %>
+
+              <div class="dropdown dropdown-top flex items-center ml-2">
+                <%= button_tag "$",
+                               tabindex: 0,
+                               role: :button,
+                               class: "text-gray-600 dark:text-gray-300 hover:text-gray-900 dark:hover:text-white",
+                               data: { role: "show-token-info" }
+                %>
+
+                <div tabindex="0" class="dropdown-content -ml-6 z-10 p-2 shadow-xl bg-base-100 rounded-box w-52 dark:!bg-gray-700">
+                  <p class="py-1 px-2 text-sm text-gray-700 dark:text-gray-300">Input tokens: <%= input_tokens %></p>
+                  <p class="py-1 px-2 text-sm text-gray-700 dark:text-gray-300">Output tokens: <%= output_tokens %></p>
+                </div>
+              </div>
+            <% end %>
           <% end %>
         </div>
       </turbo-frame>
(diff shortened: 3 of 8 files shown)


db/migrate/20240713130357_add_token_counts_to_messages.rb
```diff
@@ -0,0 +1,6 @@
+class AddTokenCountsToMessages < ActiveRecord::Migration[7.1]
+  def change
+    add_column :messages, :input_token_count, :integer
```
krschacht, line 3: I think this probably needs a default: 0 and a null: false. Probably just the null: false is what’s required, but I think it’s worth having default 0 too: this could make a difference when we’re doing a SQL query like `…sum(:input_token_count)` because activerecord won’t instantiate each record

app/services/ai_backend/anthropic.rb
```diff
@@ -28,6 +28,19 @@ def get_next_chat_message(&chunk_received_handler)
 
     response_handler = proc do |intermediate_response, bytesize|
       chunk = intermediate_response.dig("delta", "text")
+
+      input_tokens = intermediate_response.dig("message", "usage", "input_tokens")
+      if input_tokens
```
krschacht, line 33: stylistic feedback: since you are only using input_tokens here, you can combine these two lines:

```
if inputs_tokens = …
  @message …
end
```

app/services/ai_backend/anthropic.rb
```diff
@@ -28,6 +28,19 @@ def get_next_chat_message(&chunk_received_handler)
...
+        @message.input_token_count += input_tokens
+      end
+      # the initial number of output tokens is first returned in message.usage.output_tokens and
+      # the rest in usage.output_tokens in the final response
+      output_tokens = intermediate_response.dig("message", "usage", "output_tokens")
+      output_tokens = intermediate_response.dig("usage", "output_tokens") if (output_tokens.nil?)
```
krschacht, line 39: you can lose the `if` at the end of the line and instead to:

```suggestion
      output_tokens ||= intermediate_response.dig("usage", "output_tokens")
```

app/services/ai_backend/open_ai.rb
```diff
@@ -54,6 +55,15 @@ def stream_handler(&chunk_received_handler)
...
 
+      input_tokens = intermediate_response.dig("usage", "prompt_tokens")
+      if input_tokens
+        @message.input_token_count += input_tokens
+      end
+      output_tokens = intermediate_response.dig("usage", "completion_tokens")
+      if output_tokens
+        @message.output_token_count += output_tokens
+      end
+
```
krschacht, line 66: you can do the same collapsing of if … into the previous lines

app/services/ai_backend/anthropic.rb
```diff
@@ -28,6 +28,18 @@ def get_next_chat_message(&chunk_received_handler)
...
+      if (input_tokens = intermediate_response.dig("message", "usage", "input_tokens"))
+        @message.input_token_count += input_tokens
+      end
+      # the initial number of output tokens is first returned in message.usage.output_tokens and
+      # the rest in usage.output_tokens in the final response
+      output_tokens = intermediate_response.dig("message", "usage", "output_tokens")
+      output_tokens ||= intermediate_response.dig("usage", "output_tokens")
+      if output_tokens
+        @message.output_token_count += output_tokens
+      end
```
krschacht, line 41: I checked with the anthropic SDK team and we only need the final output_tokens count. That number is cumulative. We can ignore the initial one: https://github.com/anthropics/anthropic-sdk-python/issues/424#issuecomment-2227894093

```suggestion
      if output_tokens = intermediate_response.dig("message", "usage", "output_tokens")
        @message.output_token_count += output_tokens
      end
```

app/services/ai_backend/open_ai.rb
```diff
@@ -54,6 +55,14 @@ def stream_handler(&chunk_received_handler)
       content_chunk = intermediate_response.dig("choices", 0, "delta", "content")
       tool_calls_chunk = intermediate_response.dig("choices", 0, "delta", "tool_calls")
 
+      if (input_tokens = intermediate_response.dig("usage", "prompt_tokens"))
+        @message.input_token_count += input_tokens
+      end
+
+      if (output_tokens = intermediate_response.dig("usage", "completion_tokens"))
+        @message.output_token_count += output_tokens
+      end
```
krschacht, line 64: This may not be worth simplifying further, but since the two of these are closely related we could conceptually make it a single thought. You decide if this is clearer or not:

```suggestion
      if (input_tokens, output_tokens = intermediate_response["usage"]&.values_at("prompt_tokens", "completion_tokens"))
        @message.input_token_count += input_tokens.to_i
        @message.output_token_count += output_tokens.to_i
      end
```

app/views/messages/_message.html.erb
```diff
@@ -245,6 +245,24 @@ end
...
+                               class: "text-gray-600 dark:text-gray-300 hover:text-gray-900 dark:hover:text-white",
+                               data: { role: "show-token-info" }
+                %>
+
+                <div tabindex="0" class="dropdown-content -ml-6 z-10 p-2 shadow-xl bg-base-100 rounded-box w-52 dark:!bg-gray-700">
+                  <p class="py-1 px-2 text-sm text-gray-700 dark:text-gray-300">Input tokens: <%= input_tokens %></p>
+                  <p class="py-1 px-2 text-sm text-gray-700 dark:text-gray-300">Output tokens: <%= output_tokens %></p>
+                </div>
+              </div>
+            <% end %>
```
krschacht, line 265: @lumpidu I just realized a pretty decent issue with this which I didn't consider before: the _message partial updates *a lot* because we stream continual fragments. Having this partial trigger some SQL queries doesn't seem like a good idea.

I spent a little while looking at the UI and I made room for this next to the conversation title, in the left column. Here is the screenshot:
image

What do you think? It's this PR #463 

One implication of putting it over here is that we need to make sure it updates itself. I already have the front-end wired up with turbo so that if a conversation model changes at all, it will broadcast this update to the left side and replace the full `_conversation` partial. Every time a message finishes streaming, it does `message.conversation.touch` to trigger an update to the timestamp of the conversation object, and this also causes the conversation title to update because it has an after_touch hook (all this is within `conversation.rb`)

If you like the direction I'm heading with this, then I can merge in that PR 463 and then you can pull main back into this PR. You can then update this PR to be:

* Add 3 token columns to the conversation table: `input_token_count`, `output_token_count`, and `token_cost`
* Add an after_save hook within message which updates these three columns on the conversation
* Within get_next_ai_message_job find the conversation.touch and remove this, it's no longer necessary now that message.save always triggers the conversation to update
* Within conversation.rb change the `after_touch` to an `after_save`

krschacht: Great! It's now in main so you can pull it into this branch.

krschacht:
I can't push some fixes to this branch and I want to get it merged in, so I needed to create a new PR: https://github.com/AllYourBot/hostedgpt/pull/495

PR #464, opened 2024-07-16
Cleanup message partial

krschacht:
In preparation for some voice refactoring that I'm doing, this PR cleans up the message partial. Notable changes:

* The key change is that the message-scroller stimulus controller that exists per message (within _message partial) needs to be reconnected each time the partial updates with a streaming response. However, in the next PR I'm going to land, there is a new playback stimulus controller which should NOT reconnect each time the partial updates. The _message partial already has an outer most div and an inner div. Currently, both are replaced each time we do `broadcast_turbo_replace` so I change this to be `broadcast_turbo_update` in the worker. Replace does the whole div whereas update does the innerHTML of the div. I then move the existing message-scroller to the inner div.
* Move some _message logic into message_helper
* Version bump stimulus rails
* Rename a method in composer_controller for clarity
* Rename a method in message.rb for clarity

### DIFF
--- a/app/views/messages/_message.html.erb
+++ b/app/views/messages/_message.html.erb
@@ -25,8 +25,8 @@ end %>
   <div
     class="flex"
     data-role="inner-message"
-    data-controller="<%= (scroll_down || only_scroll_down_if_was_bottom) && "message-scroller" %>"
 
+    data-controller="<%= (scroll_down || only_scroll_down_if_was_bottom) && "message-scroller" %>"
     data-message-scroller-scroll-down-value="<%= scroll_down %>"
     data-message-scroller-only-scroll-down-if-scrolled-to-bottom-value="<%= only_scroll_down_if_was_bottom %>"
     data-message-scroller-instantly-value="<%= initial_page_load %>"
--- a/app/views/messages/_message.html.erb
+++ b/app/views/messages/_message.html.erb
@@ -17,9 +17,15 @@ end %>
 
 <div
   id="<%= dom_id message %>"
+  class="flex mb-5 ml-9 sm:ml-3 md:ml-8 group <%= message.only_tool_response? && !debug_tool_messages && "hidden" %>"
   data-role="message"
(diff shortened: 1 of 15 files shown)


app/views/messages/_message.html.erb
```diff
@@ -1,284 +1,274 @@
...
   data-role="message"
-  data-timestamp="<%= timestamp %>"
-  data-controller="<%= (scroll_down || only_scroll_down_if_was_bottom) && "message-scroller" %>"
-  data-message-scroller-scroll-down-value="<%= scroll_down %>"
-  data-message-scroller-only-scroll-down-if-scrolled-to-bottom-value="<%= only_scroll_down_if_was_bottom %>"
-  data-message-scroller-instantly-value="<%= initial_page_load %>"
+  data-subrole="<%= message.role %>-message"
+  class="flex mb-5 ml-9 sm:ml-3 md:ml-8 group <%= message.only_tool_response? && !debug_tool_messages && "hidden" %>"
 >
-  <div class="flex">
```
krschacht, line 30: This extra div is unnecessary so I move some of the styles up to the outer div

app/views/messages/_message.html.erb
```diff
@@ -1,284 +1,274 @@
...
-%>
+<% done_thinking = !thinking && message.content_text.present? %>
+<% debug_tool_messages = false %>
 
 <div
   id="<%= dom_id message %>"
-  class="mb-5 ml-9 sm:ml-3 md:ml-8 group <%= message.only_tool_response? && !debug_tool && "hidden" %>"
   data-role="message"
-  data-timestamp="<%= timestamp %>"
-  data-controller="<%= (scroll_down || only_scroll_down_if_was_bottom) && "message-scroller" %>"
```
krschacht, line 25: These stimulus controllers don't need to be on this outer div so I move them down to the `right column` div to clear up this space for some upcoming voice work.

PR #465, opened 2024-07-17
Cloudflare R2 file storage

jlvallelonga:
Adds Cloudflare R2 as option for storing files with Active Storage.

Task from here: https://github.com/AllYourBot/hostedgpt/wiki/Draft-project-plan

```
Changing ActiveStorage config from Postgres to Cloudflare's S3 alternative. Regardless of what we learn from our load test, we need to get these big files out of the database)
```

Starting this as an optional change, but happy to modify this to make it a requirement or change it to be the default if we want.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,6 +44,7 @@ gem "tiktoken_ruby", "~> 0.0.6"
 gem "solid_queue", "~> 0.2.1"
 gem "name_of_person"
 gem "actioncable-enhanced-postgresql-adapter" # longer paylaods w/ postgresql actioncable
+gem "aws-sdk-s3", require: false
 
 gem "omniauth", "~> 2.1"
 gem "omniauth-google-oauth2", "~> 1.1"
--- a/README.md
+++ b/README.md
@@ -115,6 +115,31 @@ The file `options.yml` contains a number of Features and Settings you can config
 - `VOICE_FEATURE` - This is an experimental feature to have spoken conversation with your assistant. It's still a bit buggy but it's coming along.
 - `GOOGLE_TOOLS_FEATURE` — This is an experimental feature that enables your assistant to access your Gmail (and soon Google Tasks and Calendar).
 
+Additionally, certain features can only be configured with environment variables. These are:
+
+- `CLOUDFLARE_STORAGE_FEATURE` - This is a feature that allows you to store message attachments in Cloudflare's R2 storage. By default this is false, but you can set it to `true` to enable it. You must configure Cloudflare first for it to work.
+
+### Configuring Cloudflare Storage
+
+First you need to sign up for Cloudflare. The free tier allows 10 GB of storage. After you sign up, you need to create a new bucket and an API token. The API token should have "Object Read and Write" access to your bucket. Take note of your Access Key ID and your Secret Access Key along with your Account ID. Set the following environment variables:
+
+- `CLOUDFLARE_ACCOUNT_ID` - Your Cloudflare Account ID
+- `CLOUDFLARE_ACCESS_KEY_ID` - Your Cloudflare Access Key ID
+- `CLOUDFLARE_SECRET_ACCESS_KEY` - Your Cloudflare Secret Access Key
+- `CLOUDFLARE_BUCKET` - The name of the bucket you created
+
+or you can add these to your Rails credentials file:
+
+```yaml
+cloudflare:
+  account_id: "your-account-id"
+  access_key_id: "your-access-key-id"
+  secret_access_key: "your-secret-access-key"
+  bucket: "your-bucket-name"
+```
+
+Then be sure to set the `CLOUDFLARE_STORAGE_FEATURE` environment variable to `true`.
+
 ### Configuring Google Tools
 
 You first need to follow all the steps in the [Google OAuth instructions](#google-oauth-authentication). The only step that is optional is that you can leave `GOOGLE_AUTHENTICATION_FEATURE` set to false, which means you don't have to enable new users to register with Google. However, following all the steps will also set up Google Auth so you can connect Google Tools to your assistants. After, you complete those steps, here is the additional configuration you need to do in order to enable the Google tools:
--- a/config/application.rb
+++ b/config/application.rb
@@ -33,6 +33,13 @@ module HostedGPT
     config.time_zone = "Central Time (US & Canada)"
     config.eager_load_paths << Rails.root.join("lib")
 
+    # Active Storage
+    if ENV["CLOUDFLARE_STORAGE_FEATURE"] == "true"
+      config.active_storage.service = :cloudflare
+    else
+      config.active_storage.service = :local
+    end
+
     config.options = config_for(:options)
   end
 end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -36,9 +36,10 @@ Rails.application.configure do
   # Disable request forgery protection in test environment.
   config.action_controller.allow_forgery_protection = false
 
-  # Store uploaded files on the local file system in a temporary directory.
+  # Store uploaded files in the db for testing
   config.active_storage.service = :test
-  config.active_storage.variant_processor = :mini_magick # vips is better but does not work on Mac Apple Silicon so using this in dev & test
+  # vips is better but does not work on Mac Apple Silicon so using this in dev & test
+  config.active_storage.variant_processor = :mini_magick
 
   config.action_mailer.perform_caching = false
 
--- a/config/storage.yml
+++ b/config/storage.yml
@@ -7,3 +7,11 @@ local:
 local_public:
   service: Postgresql
   public: true # the generated URLs will not expire
+
+cloudflare:
+  service: S3
+  endpoint: https://<%= ENV["CLOUDFLARE_ACCOUNT_ID"] || Rails.application.credentials.dig(:cloudflare, :account_id) %>.r2.cloudflarestorage.com
+  access_key_id: <%= ENV["CLOUDFLARE_ACCESS_KEY_ID"] || Rails.application.credentials.dig(:cloudflare, :access_key_id) %>
+  secret_access_key: <%= ENV["CLOUDFLARE_SECRET_ACCESS_KEY"] || Rails.application.credentials.dig(:cloudflare, :secret_access_key) %>
+  region: auto
+  bucket: <%= ENV["CLOUDFLARE_BUCKET"] || Rails.application.credentials.dig(:cloudflare, :bucket) %>
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,8 @@ services:
       POSTGRES_PASSWORD: secret
     volumes:
     - hostedgpt_pgdata:/var/lib/postgresql/data
+    ports:
+    - 5433:5432
     healthcheck:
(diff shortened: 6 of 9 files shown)


krschacht:
I’ll leave comments inline on the PR, but just to comment on the overall direction. I think it’s absolutely right to implement this as a configurable option via feature flag, but keep the default to be postgresql. For anyone running the app just for themselves, postgres is simpler setup. This S3-like option is really just for larger deployments.

config/storage.yml
jlvallelonga, line 1: further up in this file, there are `local:` and `test:` - any thoughts on combining those and renaming to `postgres:`?

docker-compose.yml
```diff
@@ -42,6 +44,11 @@ services:
       - HTTP_HEADER_AUTH_EMAIL
       - HTTP_HEADER_AUTH_NAME
       - HTTP_HEADER_AUTH_UID
+      - CLOUDFLARE_STORAGE_FEATURE=true
+      - CLOUDFLARE_ACCOUNT_ID
+      - CLOUDFLARE_ACCESS_KEY_ID
+      - CLOUDFLARE_SECRET_ACCESS_KEY
+      - CLOUDFLARE_BUCKET
```
krschacht, line 51: It makes sense to include all of these so that docker passes them through, but let’s not set any defaults here. And let’s keep the default as postgres, not cloudflare. And in terms of access these config ENV, we should do that through `Feature.` and `Setting.`. If you search the codebase for “google_auth” (not case sensitive) you can see a feature which has both Feature and Setting.

docker-compose.yml
```diff
@@ -10,6 +10,8 @@ services:
       POSTGRES_PASSWORD: secret
     volumes:
     - hostedgpt_pgdata:/var/lib/postgresql/data
+    ports:
+    - 5433:5432
```
krschacht, line 14: What does this do? The docker compose has been working for me without this.

config/environments/test.rb
```diff
@@ -36,9 +36,10 @@
   # Disable request forgery protection in test environment.
   config.action_controller.allow_forgery_protection = false
 
-  # Store uploaded files on the local file system in a temporary directory.
+  # Store uploaded files in the db for testing
```
krschacht, line 39: I think we should just delete this comment, especially when it’s already out of sync with storage.yml. :) People can just check storage.yml if they want to know how test is storing files.

I’m definitely in the camp of “comment very little unless it’s really needed to clarify because inevitably the comments will get out of sync from the code.” This is one of those comments that was generated by rails with the original “rails new” and I haven’t gone through the codebase and removed those generated comments which have become irrelevant.

config/environments/test.rb
```diff
@@ -36,9 +36,10 @@
...
   config.active_storage.service = :test
-  config.active_storage.variant_processor = :mini_magick # vips is better but does not work on Mac Apple Silicon so using this in dev & test
+  # vips is better but does not work on Mac Apple Silicon so using this in dev & test
+  config.active_storage.variant_processor = :mini_magick
```
krschacht, line 42: I think it’s worth reverting this. Even though it makes the line really long, which is a bit annoying, the few times I do add comments I try to bias towards having them on the actual line they’re explaining rather than their own line, only because it slightly decreases the frequency that someone updates a line without reading the comment.

I feel like I stated that in a confusing way. All I’m trying to say is: if the comment is on its own line, it’s more likely to get out of sync. So let’s keep it connected to the mini_magick thing even though it makes the line really long.

docker-compose.yml
jlvallelonga: not necessary, but useful if you want to connect to the postgres instance that's running in docker from your local machine via tools like postico. Happy to remove it if you want. I was just using it to look at my database

README.md
```diff
@@ -115,6 +115,31 @@ The file `options.yml` contains a number of Features and Settings you can config
 - `VOICE_FEATURE` - This is an experimental feature to have spoken conversation with your assistant. It's still a bit buggy but it's coming along.
 - `GOOGLE_TOOLS_FEATURE` — This is an experimental feature that enables your assistant to access your Gmail (and soon Google Tasks and Calendar).
 
+Additionally, certain features can only be configured with environment variables. These are:
+
+- `CLOUDFLARE_STORAGE_FEATURE` - This is a feature that allows you to store message attachments in Cloudflare's R2 storage. By default this is false, but you can set it to `true` to enable it. You must configure Cloudflare first for it to work.
```
krschacht, line 120: This might be a good opportunity to explain the default, so:

```suggestion
- `CLOUDFLARE_STORAGE_FEATURE` - This is a feature that allows you to store message attachments in Cloudflare's R2 storage. When this is false, attachments are stored within postgres. This is the default configuration. But you can set it to `true` to enable it. You must configure Cloudflare Storage first for it to work.
```

Gemfile
```diff
@@ -44,6 +44,7 @@ gem "tiktoken_ruby", "~> 0.0.6"
 gem "solid_queue", "~> 0.2.1"
 gem "name_of_person"
 gem "actioncable-enhanced-postgresql-adapter" # longer paylaods w/ postgresql actioncable
+gem "aws-sdk-s3", require: false
```
krschacht, line 47: I’m not familiar with require: false, I just looked it up and it sounds like when this is false the you need to explicitly set “require …” in the file that uses it. Is that the intent?

config/application.rb
```diff
@@ -33,6 +33,13 @@ class Application < Rails::Application
     config.time_zone = "Central Time (US & Canada)"
     config.eager_load_paths << Rails.root.join("lib")
 
+    # Active Storage
+    if ENV["CLOUDFLARE_STORAGE_FEATURE"] == "true"
```
jlvallelonga, line 37: can't use `Feature.` here because application hasn't finished starting yet

config/storage.yml
krschacht: Yes that makes sense

Gemfile
jlvallelonga: Oh interesting. Maybe rails requires it explicitly internally. Just going based on the guides here: https://guides.rubyonrails.org/active_storage_overview.html#s3-service-amazon-s3-and-s3-compatible-apis

config/application.rb
krschacht: Ohhh, shoot. Hmm.  I think then still add it to Feature and options.yml, just in case other parts of the app need to check it, and then here just repeat the “ENV || application…” which you put in the options.yml

docker-compose.yml
krschacht: Ahh, that makes sense. I think I used to have a couple ports exposed and then someone submitted a PR to remove them saying: it’s exposing ports which are conflicting with my local services and those ports arent being used.

So i think it helps ppl for us to expose less

config/environments/test.rb
jlvallelonga: yeah agreed with the comments. If the code explains itself without too much trouble then no need for further explanation with comments.

For this file I purposely left it here thinking that we should override anything that would be set in the application.rb config file here since I don't think we want to upload anything to cloud storage during tests. At least that's my initial thought. What do you think?

config/storage.yml
```diff
@@ -1,9 +1,27 @@
+# TODO: remove test, local, and local_public options from this file
+#  figure out how to migrate existing records to handle that change
```
jlvallelonga, line 2: ran into some issues with existing database records when I changed `local` to `database` so I left the existing `test`, `local`, and `local_public` here for now and added this TODO. I don't plan to leave this here, but will come back tomorrow and finish. I can either revert the name changes or do a database migration to try to update existing data with the new names. I would definitely want someone with more test data to try the migration if that's the route we take though since I have almost nothing in my db at this point

krschacht: No problem. And for trying things out in development, it really helps to load the fixture data into your dev database. I used to have the seeds file configured to do this but it caused some issues in production. This is an old commit back when it was still working:

https://github.com/AllYourBot/hostedgpt/blob/f66ee3b8d2bac67acfe7f8a174f29930d7176af6/db/seeds.rb

I think this should work as is, although it may be missing a few more recent database tables so check schema.rb and maybe add the missing items to the array. Note: the array is ordered so tables referenced by other tables are loaded first.

krschacht: Oh, after you update this file,  you run `bin/rails db:seed` to load it. If that doesn't work, you maybe have to run `bin/rails db:drop; bin/rails db:prepare`. I vaguely remember having some issues with referential integrity.

And if that still doesn't work, then maybe copy and paste the referential integrity hack from this version of the seeds file:
https://github.com/AllYourBot/hostedgpt/blob/3de36588ca922a22edd232c70fe27967576e78ad/db/seeds.rb

I really probably should have preserved this "load fixtures" trick but moved it into a rake task. If you get this working, maybe when you're done with this PR we tee up another PR that turns this into a clean rake task. The only reason I removed this is because it was causing some issues with the various production hosting environments due to stricter postgres permissions, but this is quite helpful in development.

config/environments/test.rb
krschacht: Agreed, we don't want to upload to cloud storage during tests. But since the default is postgres, we shouldn't need to override this for test and instead can use the postgres default in the test environment. I think it's a good thing that tests will use the postgres adapter because if we ever inadvertently break that adapter we'll notice failing tests.

jlvallelonga:
I think this is all good, but I'm working on running the seeds to verify. I was able to create some files on the main branch and then checkout this branch, see it breaking, and then migrate to see it working again. If we're ok with that then feel free to merge, but otherwise I'll try to have the seeds working before I sleep tonight or tomorrow

krschacht:
Hi Justin, I had a busy day today and didn’t get a chance to take a look,
but I should be able to tomorrow AM before you start (I’m still in Europe).

krschacht:
@jlvallelonga I gave this one more read through before merging in. I didn't love the special-casing of this one feature being ENV only (and needing to explain that) so I figured out how to make `Feature...` work within application.rb. You can see it in this commit, if you're curious: [`8873ede` (#465)](https://github.com/AllYourBot/hostedgpt/pull/465/commits/8873ede76c85c8ebf748646841b9830163c560a3)

While I was in there, I did a close read-through of all the optional features we cover within the README and I did copy editing of that to try and streamline it a bit. While doing this, I decided to remove any mention of the rails credentials file. We'll still support it, but I'll consider that an advanced thing for people who already know how rails credentials work. For everyone else, it's easiest and more standard to simply set ENV.

jlvallelonga:
> @jlvallelonga I gave this one more read through before merging in. I didn't love the special-casing of this one feature being ENV only (and needing to explain that) so I figured out how to make `Feature...` work within application.rb. You can see it in this commit, if you're curious: [`8873ede` (#465)](https://github.com/AllYourBot/hostedgpt/pull/465/commits/8873ede76c85c8ebf748646841b9830163c560a3)
> 
> While I was in there, I did a close read-through of all the optional features we cover within the README and I did copy editing of that to try and streamline it a bit. While doing this, I decided to remove any mention of the rails credentials file. We'll still support it, but I'll consider that an advanced thing for people who already know how rails credentials work. For everyone else, it's easiest and more standard to simply set ENV.

nice! definitely more clean that way

krschacht:
First week, first PR merged in 🍾 🎉  It’s always nice to get something full cycle. Now we’re off!

PR #466, opened 2024-07-19
Add note about db:fixtures:load to README

jlvallelonga:
following up on conversation here:

https://github.com/AllYourBot/hostedgpt/pull/465#discussion_r1681218434

turns out there's a db:fixtures:load task built-in so just adding a readme item


jlvallelonga:
@krschacht for the production issues, does the `if Rails.env.development?` check solve that. I'm not sure I understand your vision for the rake task on this one. We're only trying to load the seed data in dev/test right?

krschacht:
I was just thinking we should make this a rake task, so you and I (and future developers) can more easily populate the database. Maybe just `bin/rails db:fixtures` will load all the fixtures in the database.

You'd change this to be within lib/tasks/db/fixtures.rake

And then I don't think you need the environment check since you and I would just execute it deliberately when we want to quickly get some data in there for testing.

README.md
```diff
@@ -83,7 +83,9 @@ Deploying to Fly.io is another great option. It's not quite one-click like Rende
 1. Click Fork > Create New Fork at the top of this repository. Pull your forked repository down to your computer (the usual git clone ...).
 1. Install the Fly command-line tool on Mac with `brew install flyctl` otherwise `curl -L https://fly.io/install.sh | sh` ([view instructions](https://fly.io/docs/hands-on/install-flyctl/))
 1. Think of an internal Fly name for your app, it has to be unique to all of Fly, and then in the root directory of the repository you pulled down, run `fly launch --build-only --copy-config --name=APP_NAME_YOU_CHOSE`
-  * Say "Yes" when it asks if you want to tweak these settings
+
+- Say "Yes" when it asks if you want to tweak these settings
+
```
jlvallelonga, line 88: i think my vs code made these changes because of a markdown plugin. I assume it's doing the right thing, but if not then I can undo these extra changes

README.md
```diff
@@ -234,6 +243,7 @@ The easiest way to get up and running is to use the provided docker compose work
 1. Run tests: `docker compose run base rails test` The app has comprehensive test coverage but note that system tests currently do not work in docker.
 1. Open the rails console: `docker compose run base rails console`
 1. Run a psql console: `docker compose run base psql`
+1. Load fixtures: `docker compose run base rails db:fixtures:load`
```
jlvallelonga, line 246: this is the only change that I intended to make to this file

jlvallelonga:
I found out that there's actually a built-in rake task for this so I removed the seeds file and added a note to the readme. Loading fixtures can be done with `docker compose run base rails db:fixtures:load`

docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: "3.1"
```
jlvallelonga, line 1: was getting the following warning when running docker:
```
WARN[0000] .../hostedgpt/docker-compose.yml: `version` is obsolete
```

README.md
```diff
@@ -83,7 +83,9 @@ Deploying to Fly.io is another great option. It's not quite one-click like Rende
 1. Click Fork > Create New Fork at the top of this repository. Pull your forked repository down to your computer (the usual git clone ...).
 1. Install the Fly command-line tool on Mac with `brew install flyctl` otherwise `curl -L https://fly.io/install.sh | sh` ([view instructions](https://fly.io/docs/hands-on/install-flyctl/))
 1. Think of an internal Fly name for your app, it has to be unique to all of Fly, and then in the root directory of the repository you pulled down, run `fly launch --build-only --copy-config --name=APP_NAME_YOU_CHOSE`
-  * Say "Yes" when it asks if you want to tweak these settings
+
+- Say "Yes" when it asks if you want to tweak these settings
```
krschacht, line 87: I reviewed the markdown and actually, the old one was incorrect, but the change is incorrect too. :) We need an extra level of indentation so that it actually nests beneath the number and doesn't restart numbering after this.

```suggestion
    - Say "Yes" when it asks if you want to tweak these settings
```

README.md
krschacht, line 246: Let's give a bit more context as to why someone should run this command. Also, let's add this same line (without the "docker compose run base" prefix) to the section below. The very next section walks through setting up outside of docker and it can be helpful to have this step there too.

```suggestion
1. If you want a few fake users and a bunch of conversations and other data pre-populated in the database, you can load fixtures into the development database. This can be helpful, for example, if you want to test a migration and save yourself the time manually creating a bunch of data: `docker compose run base rails db:fixtures:load`
```

krschacht:
@jlvallelonga I can't believe I never knew about this! I spent a bunch of hours implementing and maintaining that old seeds file thing. Sheesh. And I swear I even googled about it because it wasn't easy to manually load fixture data from the seeds file. :) Anyway, TIL ...

PR #470, opened 2024-07-23
Add "Forgot Password" reset feature using Postmark

jlvallelonga:
adds forgot password flow to the app and utilizes Postmark


jlvallelonga:
still need to move some things around a bit, but this is the general idea

config/environments/production.rb
```diff
@@ -58,11 +58,6 @@
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
 
-  # Info include generic and useful information about system operation, but avoids logging too much
-  # information to avoid inadvertent exposure of personally identifiable information (PII). If you
-  # want to log everything, set the level to "debug".
-  config.log_level = ENV.fetch("RAILS_LOG_LEVEL", "info")
-
```
jlvallelonga, line 65: I moved this up to application while trying to look at logging in my dev env. Is there a reason to not have it there though? wanted to make sure before this gets merged

app/controllers/password_resets_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  layout "public"
+
+  def new
+  end
+
+  def create
+    user_agent = request.env['HTTP_USER_AGENT']
+    os = get_item_in_str(user_agent, KNOWN_OPERATING_SYSTEMS) || "unknown operating system"
+    browser = get_item_in_str(user_agent, KNOWN_BROWSERS) || "unknown browser"
```
jlvallelonga, line 13: there's probably a better place for these, but maybe I should just use this?: https://github.com/ua-parser/uap-ruby

app/views/layouts/public.html.erb
```diff
@@ -3,7 +3,15 @@
   <%= render "head" %>
   <body class="font-sans text-gray-950 dark:text-gray-100 system">
     <main class="flex bg-white dark:bg-gray-800  w-full dark:dark:text-gray-100">
-      <%= content_for?(:main) ? yield(:main) : yield %>
+      <div class="flex flex-col items-center justify-center w-full h-full space-y-4">
+        <div class="p-3 bg-white border border-gray-100 rounded-full">
+          <%= image_tag("logo.svg", width: "24", height: "24", class: "text-gray-950") %>
+        </div>
+
+        <h2 class='text-3xl font-semibold'><%= content_for?(:heading) ? yield(:heading) : "Welcome back" %></h2>
```
jlvallelonga, line 11: moved these lines into the public layout since the only views using this layout all had these lines duplicated

config/environments/production.rb
krschacht: In most situations, there are intentionally different log levels in the different environments. The default log_level is debug which works great for dev and test so it's just production which overrides this and sets it to info.

app/controllers/password_resets_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  layout "public"
+
+  def new
+  end
+
+  def create
+    os = request.operating_system
+    browser = request.browser
+    SendResetPasswordEmailJob.perform_later(params[:email], os, browser) # queue as a job to avoid timing attacks
```
krschacht, line 13: What's a timing attack, do you mean rate limiting? I just remembered seeing something in the campfire source code for rate limiting:
```
rate_limit to: 10, within: 3.minutes, only: :create, with: -> { render_rejection :too_many_requests }
```

I didn't know this was a feature in rails (and we should check if we're on the correct version). This may be worth adding to the controller.

app/controllers/password_resets_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    redirect_to '/login', notice: 'If an account with that email was found, we have sent a link to reset the password'
+  end
+
+  def edit
+    @person = find_signed_person(params[:token])
+  end
+
+  def update
+    @person = find_signed_person(params[:token])
+    password_credential = @person&.personable&.password_credential
```
krschacht, line 24: Let's use .user for readability, plus we wouldn't want to update a password for a Tombstone or other odd state of a user. Also, that first @person is guaranteed so only need the second &, right? Actually, as I look at that rescue I'm not sure why we only rescue on Invalid. Let me comment below.

```suggestion
    password_credential = @person.user&.password_credential
```

krschacht (changes requested):
Sorry I didn't get to this yesterday. I started today and got through a chunk, and I'm going to an appointment now. I'll go ahead and submit what I got since I don't know when you're working.

One general comment: let's use double quotes everywhere so we don't have the rails single-quote sometimes and double-quotes with interpolation. I've been trying to default to those. And that reminds me, there is an .editorconfig which I think you need to activate in your IDE (and I can't recall, but I think this might do the double quotes automatically, not sure).

app/controllers/password_resets_controller.rb
jlvallelonga: rate limiting would definitely slow down a timing attack, but I'm not sure it completely solves it. Timing attacks are when an attacker figures out if an email (or other) exists for a user in a system by timing how long it takes to get a response. Slower responses indicate that a user is present in the system. If an attacker knows that an email exists then they can use leaked passwords from other websites to gain access to this site since people often use the same password. There's a bunch of other stuff around this and probably much better explanations elsewhere, but that's what I understand anyway.

Probably not a bad idea to do a pass at some rate limiting in the controllers too? Or maybe it's better at the infrastructure layer, but that would mean anyone deploying HostedGPT would have to do it themselves. I wonder if we can just do some generic rate limiting in the ApplicationController and call it good enough

config/environments/production.rb
jlvallelonga: ah ok. I'm still slowly remembering rails. I was having trouble with logging, but I think it was due to something else. I'll change this back

app/controllers/password_resets_controller.rb
jlvallelonga: just added these as methods to the request object

app/controllers/password_resets_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+  end
+
+  private
+
+  def find_signed_person(token)
+    Person.find_signed!(token, purpose: Rails.application.config.password_reset_token_purpose)
+  rescue ActiveSupport::MessageVerifier::InvalidSignature
+    redirect_to '/login', alert: 'Your token has expired. Please try again'
+  end
```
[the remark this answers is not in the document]
jlvallelonga: yeah it only throws the InvalidSignature exception unless it's for the right record, but the record isn't found (I guess if it's been deleted). From the docs:

> Works like [find_signed](https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/SignedId/ClassMethods.html#method-i-find_signed), but will raise an ActiveSupport::MessageVerifier::InvalidSignature exception if the signed_id has either expired, has a purpose mismatch, is for another record, or has been tampered with. It will also raise an ActiveRecord::RecordNotFound exception if the valid signed id can’t find a record.

from: https://api.rubyonrails.org/v7.1.3.4/classes/ActiveRecord/SignedId/ClassMethods.html#method-i-find_signed-21

Figured I'd just let rails handle the RecordNotFound and return a 404 if the person record is actually missing

app/controllers/password_resets_controller.rb
krschacht: Ah, interesting. Now I get what you mean by timing attack. Security isn't a big strength of mine so that's helpful.

On the rate limit, I think we go ahead and add a line like the one below just to this controller and over time we can add them to the others that make sense. I'd hesitate to add it to all of them in one place because my hunch is that we'll need to individualize it for a few endpoints.

app/controllers/password_resets_controller.rb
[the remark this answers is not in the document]
krschacht: Oh, got it, yes that seems like a good solution if the NotFound will bubble up and abort controller execution.

Let's be sure and represent that as a test case in the controller tests for this. I like how adding test cases serves as a bit of self-documenting code (e.g. a test communications "we thought about this situation, here is how it works")

app/views/password_resets/new.html.erb
```diff
@@ -0,0 +1,23 @@
...
+
+  <%= f.submit 'Send reset email',
+        class: %|
+          text-white font-medium
+          bg-brand-blue dark:bg-gray-900
+          border border-white dark:border-gray-400
+          rounded-lg p-4 text-center
+          cursor-pointer
+          hover:opacity-95
+        |
```
krschacht, line 21: Let's do the indentation on these so that it's all referenced from the `<%`. Meaning, the way you did `email_field` above is right — you pulled everything back to indent from the `<%` whereas on this one class is indented from `f.submit` rather than pulled back to indent from `<%`. I learned this one from Rob and have been moving the app over to this style.

app/views/users/new.html.erb
```diff
@@ -1,22 +1,89 @@
...
+<% if Feature.password_authentication? %>
+  <%= form_with method: :post,
+        model: @person,
+        url: users_path,
+        class: "flex flex-col space-y-4 w-80",
+        data: {
+          controller: "transition",
+          transition_toggle_class: "max-h-0 max-h-40"
+        } do
+  |f| %>
```
krschacht, line 42: another indentation to pull back in. I won't flag all of them but it's worth a pass

docker-compose.yml
```diff
@@ -47,6 +47,16 @@ services:
...
+      - POSTMARK_MAILER_FEATURE
+      - POSTMARK_SERVER_API_TOKEN
+      - POSTMARK_FROM_EMAIL
+      - POSTMARK_PASSWORD_RESET_TEMPLATE_ALIAS
+      - ACTION_MAILER_HOST
+      - PRODUCT_NAME
+      - SUPPORT_URL
+      - COMPANY_NAME
+      - COMPANY_ADDRESS
+      - PASSWORD_RESET_TOKEN_TTL_MINUTES
```
krschacht, line 59: Let's order all of the ENV keys in this whole block so that it matches the order of the file in options.yml, e.g. POSTMARK FEATURE would be up with the other features. Feel free to reorder other ones, I haven't checked it in awhile.

Every so often I compare options.yml to this file to see if we've missed any and that comparison is easy when they're ordered identically but it's hard when they're ordered/grouped differently.

krschacht (changes requested):
I finished doing a quick pass through the whole PR today. Tag me when you think it's worth me doing another review that way I won't keep jumping in and out with each comment you add. I think it'll be more efficient for me to wait until you've finished your next round of stuff.

But do tag me / chat me if you run into something in the meantime that you want my thoughts on.

app/controllers/password_resets_controller.rb
jlvallelonga: oh I just looked into this and looks like it's not out in a stable release yet. I think it's in rails 7.2, but that's still a beta release. Unless we want to use the beta release, I'll just leave this out for now. I don't think it's imperative to have just yet

config/routes.rb
```diff
@@ -28,6 +32,11 @@
   get "/register", to: "users#new"
   get "/logout", to: "authentications#destroy"
 
+  get '/password/reset', to: 'password_resets#new'
+  post '/password/reset', to: 'password_resets#create'
+  get '/password/reset/edit', to: 'password_resets#edit'
+  patch '/password/reset/edit', to: 'password_resets#update'
```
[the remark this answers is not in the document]
jlvallelonga: ah yeah I was following some instructions and I forgot to clean that part up. I added `resources :password_resets, only: [:new, :create, :edit, :update]` and the resulting routes are:

```
                         password_resets POST   /password_resets(.:format)                                                                        password_resets#create
                      new_password_reset GET    /password_resets/new(.:format)                                                                    password_resets#new
                     edit_password_reset GET    /password_resets/:id/edit(.:format)                                                               password_resets#edit
                          password_reset PATCH  /password_resets/:id(.:format)                                                                    password_resets#update
                                         PUT    /password_resets/:id(.:format)                                                                    password_resets#update
```

so the added :id url parameters don't work with this, but I think I can add some different routes to make more sense of this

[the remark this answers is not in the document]
krschacht: @jlvallelonga just drop the 's' so "resources" becomes `resource`. I didn't catch that.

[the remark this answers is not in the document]
jlvallelonga: I ended up adding this:
```
  resources :password_resets, only: [:new, :create]
  resource :password, only: [:edit, :update]
```

which gives us the following routes:
```
                         password_resets POST   /password_resets(.:format)                                                                        password_resets#create
                      new_password_reset GET    /password_resets/new(.:format)                                                                    password_resets#new
                           edit_password GET    /password/edit(.:format)                                                                          passwords#edit
                                password PATCH  /password(.:format)                                                                               passwords#update
                                         PUT    /password(.:format)                                                                               passwords#update
```

I think it makes sense to split these into two different RESTful resources: password_resets and passwords. So I created another controller for passwords. Creating a password_reset creates the reset token and sends the email and then you can edit and update the password for the user which is a different kind of resource. It felt wrong to update a user's password by editing and updating the "password reset"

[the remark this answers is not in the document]
jlvallelonga: Also, I thought about nesting this under users/:token/password, but it felt weird to have the really long token as a url parameter instead of a query parameter. i.e. `users/<really long token>/password` was weird so I went with `/password?token=<really long token>`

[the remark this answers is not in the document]
krschacht: I agree with that. I’ve come to avoid nested resources accept in rare cases where it really adds something. That is mostly only in the case where I have one resource unnested and another version nested.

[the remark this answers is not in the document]
krschacht: I also agree it makes sense to have separate resources. Adding more resources to try and stick to CRUD always seems to be worth it.

But on password, I wonder if it should just be under person/edit but that’s a bit more complicated with the nested person > user > password credential hierarchy so it’s probably worth the dedicated password route, like you did.

Actually, I might lean towards naming the route password_credential since that’s the model it’s related to, but password seems fine too. It’s a bit like when I added “signup” for users/create just because it’s a prettier route.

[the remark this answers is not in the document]
jlvallelonga: I like the password_credential route matching the model so I made that change too

jlvallelonga:
ok updated everything I think, working on another pr against this one to modify the settings a bit, but that can come in later if we like it:https://github.com/AllYourBot/hostedgpt/pull/476

test/mailers/password_mailer_test.rb
```diff
@@ -0,0 +1,34 @@
...
+      password_reset_email: true,
+      email_sender_postmark: true
+    }
+  end
+
+  test "reset" do
+    os = "Windows"
+    browser = "Chrome"
+
+    stub_settings_and_features do
```
jlvallelonga, line 23: this didn't actually do what I wanted it to do and set the feature in time to create the password reset routes. I added it to all the tests and then realized that CI was still broken. So I'll go back and remove it, but I think maybe as more env vars are added, we'll want a better way to set up the test environment so that running tests locally and running in CI behaves the same way regardless of what env vars we have set locally

config/initializers/rails_extensions.rb
```diff
@@ -0,0 +1,4 @@
+# recursive require of all files in lib/rails_extensions
+Dir[File.join(Rails.root, "lib", "rails_extensions", "**/*.rb")].each do |path|
+  require path
+end
```
krschacht, line 4: The convention of putting this in lib/rails_extensions and then having a special initializer is new to me. Looking at the current codebase, I've either done an initializer itself. Look at `extend_active_record.rb` (that probably deserves a more specific name) and `extend_anthropic_client.rb`. And I've also just re-opened the class itself within lib/ such as `lib/string.rb`. What do you think is the best pattern?

Actually, I just looked at the campfire codebase from 37signals and they do exactly what you just did. Interesting. I think this new pattern you're introducing might be the best pattern. Which means we'll just migrate the existing rails extensions over to this approach (in a different PR) and maybe those few ruby extensions should really have been initializers or maybe we also add a ruby_extensions equivalent? Not sure. All future PR. Nothing for you to do here but I'll just keep this comment since I thought out loud about it all. :)

lib/rails_extensions/action_dispatch/request.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    private
+
+    def get_item_in_str(str, items)
+      items.each do |item|
+        if str.include?(item)
+          return item
+        end
+      end
+    end
```
krschacht, line 22: Ruby basically as this method build in already as find or it's alias `detect` which is a little more idiomatic for 

```suggestion
    def operating_system
      KNOWN_OPERATING_SYSTEMS.detect { |os| user_agent.include?(os) } || "unknown operating system"
    end

    def browser
      KNOWN_BROWSERS.detect { |browser| user_agent.include?(browser) } || "unknown browser"
    end
```

app/views/password_resets/new.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    class: %|
+      text-white font-medium
+      bg-brand-blue dark:bg-gray-900
+      border border-white dark:border-gray-400
+      rounded-lg p-4 text-center
+      cursor-pointer
+      hover:opacity-95
+    |
+  %>
+<%end%>
```
krschacht, line 23: spaces
```suggestion
<% end %>
```

app/views/password_resets/new.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  <%= f.label :email, class: "input input-bordered flex items-center justify-between" do %>
+    <span>Email <span class="">*</span></span>
+    <%= f.email_field :email,
+      autofocus: true,
+      placeholder: "Email",
+      class: "w-44 border-0 dark:text-black"
+    %>
+  <% end %>
+
+  <%= f.submit 'Send reset email',
```
krschacht, line 13: Let's title case these button labels since that's how it looks like I did the others (not super deliberate, just making it consistent):

```suggestion
  <%= f.submit "Send Reset Email',
```

app/views/password_credentials/edit.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    class: %|
+      text-white font-medium
+      bg-brand-blue dark:bg-gray-900
+      border border-white dark:border-gray-400
+      rounded-lg p-4 text-center
+      cursor-pointer
+      hover:opacity-95
+    |
+  %>
+<%end%>
```
krschacht, line 24: spaces
```suggestion
<% end %>
```

app/controllers/password_credentials_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def edit
+    @user = find_signed_user(params[:token])
+  end
+
+  def update
+    user = find_signed_user(params[:token])
+
+    if user.password_credential&.update(update_params)
+      redirect_to login_path, notice: "Your password was reset succesfully. Please sign in."
```
krschacht, line 15: Would it be trivial to just log the user in at this point? If it's not trivial, let's not bother. But it's always a slightly nicer experience when the website simply lets you in right away since we're already trusting that you are the owner of this account.

I think it may simply be:

```
      login_as(person, credential: user.password_credential)
      redirect_to root_path, notice: "Your password was reset successfully."
```

app/helpers/password_credentials_helper.rb
```diff
@@ -0,0 +1,2 @@
+module PasswordsHelper
```
krschacht, line 1:
```suggestion
module PasswordCredentialsHelper
```

test/mailers/password_mailer_test.rb
krschacht: Hmm, that's an interesting thought. I'm pretty sure it's worked every other time, but I totally get that routes may be loaded early enough that it doesn't work for routes. Asking chatgpt it suggests that we simply do:

`Rails.application.reload_routes!`

I didn't test that but maybe all we need to do is add that to stub_settings and stub_features?

test/controllers/password_credentials_controller_test.rb
```diff
@@ -0,0 +1,66 @@
...
+    credentials(:keith_password)
+
+    @settings = {
+      email_from: "teampeople@example.com",
+      product_name: "Product Name"
+    }
+    @features = {
+      password_reset_email: true,
+      email_sender_postmark: true
+    }
```
krschacht, line 18: You probably want to do the stubbing of settings and features in here, right? Because it's not currently being applied to all of the tests but we probably want it to be activated for all of the tests to ensure they're fully running as we'd expect them to.

In terms of syntax, I lean towards not having `stub_settings_and_features` read the instance var. That's a little magical to set an instance var and then know that a method is reading it. I think it's better to explicitly pass it in. My inclination would be do:

Oh, but shoot, I think the reason you didn't stub this within `setup do` is because the method requires a block.

How about this for a solution — both for this controller test and the other controller test:

1) We drop stub_settings_and_features and instead add a couple new methods to that helper (chatgpt wrote this for me, I didn't test it but it looks right):

```
def stub_features(hash)
  Feature.features_hash = nil
  @original_features = Feature.method(:raw_features)
  Feature.define_singleton_method(:raw_features) { Rails.application.config.options.features.merge(hash) }
  Rails.application.reload_routes!
end

def unstub_features
  Feature.singleton_class.send(:remove_method, :raw_features)
  Feature.define_singleton_method(:raw_features, @original_features)
  Feature.features_hash = nil
  Rails.application.reload_routes!
end

def stub_settings(hash)
  @original_settings = Setting.method(:settings)
  Setting.define_singleton_method(:settings) { Rails.application.config.options.settings.merge(hash) }
  Rails.application.reload_routes!
end

def unstub_settings
  Setting.singleton_class.send(:remove_method, :settings)
  Setting.define_singleton_method(:settings, @original_settings)
  Rails.application.reload_routes!
end
```

2) Then the before can be modified to:

```
setup do
  stub_features(feature1: true, feature2: false)
  stub_settings(setting1: 'value1', setting2: 'value2')
end

teardown do
  unstub_features
  unstub_settings
end
```

This way we are confident that every test is running with the feature fully enabled, so we're not inadvertently short-circuiting any code.

app/controllers/password_resets_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def new
+  end
+
+  def create
+    os = request.operating_system
+    browser = request.browser
+    SendResetPasswordEmailJob.perform_later(params[:email], os, browser) # queue as a job to avoid timing attacks
+
+    redirect_to login_path, notice: "If an account with that email was found, we have sent a link to reset the password"
```
krschacht, line 15: copyediting to be a little shorter for quick read and adding period (I really need to make these toast messages be persistent and not slide off so quickly):

```suggestion
    redirect_to login_path, notice: "If that email is valid, we just sent password reset email."
```

app/views/password_mailer/reset.html.erb
```diff
@@ -0,0 +1,502 @@
...
+    /* Buttons ------------------------------ */
+
+    .button {
+      background-color: #3869D4;
+      border-top: 10px solid #3869D4;
+      border-right: 18px solid #3869D4;
+      border-bottom: 10px solid #3869D4;
+      border-left: 18px solid #3869D4;
+      display: inline-block;
+      color: #FFF;
```
krschacht, line 119: My gmail client is overriding this color and rendering the button text as blue since it's a hyperlink. Adding this fixed it when I tested:

```suggestion
      color: #FFF !important;
```

app/views/password_mailer/reset.html.erb
```diff
@@ -0,0 +1,502 @@
...
+    }
+
+    .email-masthead_logo {
+      width: 94px;
+    }
+
+    .email-masthead_name {
+      font-size: 16px;
+      font-weight: bold;
+      color: #A8AAAF;
```
krschacht, line 328: this too, another link that's turning blue, this also fixed it:

```suggestion
      color: #A8AAAF !important;
```

app/views/password_mailer/reset.html.erb
```diff
@@ -0,0 +1,502 @@
...
+                        <table class="body-action" align="center" width="100%" cellpadding="0" cellspacing="0" role="presentation">
+                          <tr>
+                            <td align="center">
+                              <!-- Border based button
+           https://litmus.com/blog/a-guide-to-bulletproof-buttons-in-email-design -->
+                              <table width="100%" border="0" cellspacing="0" cellpadding="0" role="presentation">
+                                <tr>
+                                  <td align="center">
+                                    <%= link_to @edit_url, class: "f-fallback button button--green", target: "_blank" do %>
+                                      Reset your password
```
krschacht, line 471: just titlecasing button
```suggestion
                                      Reset Your Password
```

krschacht (dismissed):
I gave the code a close review and I tested everything locally, it worked great! You did an amazing job with this. I didn't anticipate how involved it would be. So much setup for emailers, multiple controllers, etc etc. But it looks really solid.

I added inline comments. It looks like a lot but the vast majority of these are really minor styling and most I did as github suggestions so you should just be able to click "accept suggestion" and it'll make the edit for you — if it looks like a good edit. There are a couple questions about logic.

config/options.yml
```diff
@@ -9,8 +9,11 @@ shared:
     http_header_authentication: <%= ENV.fetch("HTTP_HEADER_AUTHENTICATION_FEATURE", false) %>
     voice: <%= ENV.fetch("VOICE_FEATURE", false) %>
     default_to_voice: <%= ENV.fetch("DEFAULT_TO_VOICE_FEATURE", false) %>
+    password_reset_email: <%= ENV.fetch("PASSWORD_RESET_EMAIL_FEATURE", false) %>
+    email_sender_postmark: <%= ENV.fetch("EMAIL_SENDER_POSTMARK_FEATURE", false) %>
```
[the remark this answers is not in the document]
jlvallelonga: yeah that makes sense. Sender does sound like the "from" field. I can change it to provider or just leave it off. What do you think about this change instead and having this as a setting?: https://github.com/AllYourBot/hostedgpt/pull/476

lib/rails_extensions/active_support/duration.rb
```diff
@@ -0,0 +1,10 @@
+module ActiveSupport
+  class Duration
+    def as_sentence
+      # build simplifies the duration e.g. `{:seconds=>37090350}` becomes `{:years=>1, :months=>2, :days=>3, :hours=>4, :minutes=>5, :seconds=>6}`
+      Duration.build(self.seconds).parts.
+        map { |unit, v| "#{v} #{v == 1 ? unit.to_s.singularize : unit.to_s}" }.
+        to_sentence
+    end
```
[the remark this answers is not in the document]
jlvallelonga: you can actually use `.inspect` to get exactly this (or very similar) but someone somewhere said that you should never use `inspect` in production so I just added this instead

config/initializers/rails_extensions.rb
jlvallelonga: yeah I saw a few ways to do it when I was working on this. This way seemed to be the cleanest, but definitely can revisit if it starts to get messy

app/controllers/password_credentials_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  layout "public"
+
+  def edit
+    @user = find_signed_user(params[:token])
+  end
+
+  def update
+    user = find_signed_user(params[:token])
+
+    if user.password_credential&.update(update_params)
```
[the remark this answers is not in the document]
jlvallelonga: ah yeah I ran into that when I was testing early on actually. That's a good point. They wouldn't actually get to this point because I check in the job before I send the email, but maybe I shouldn't do that and send a token even if they don't have a password credential. I'll update it to create if it doesn't exist. I could definitely seeing this being confusing for users they way it currently is

.github/workflows/rubyonrails.yml
```diff
@@ -28,6 +28,12 @@ jobs:
     env:
       RAILS_ENV: test
       DATABASE_URL: "postgres://rails:password@localhost:5432/hostedgpt_test"
+      EMAIL_FEATURE: true
+      PASSWORD_RESET_EMAIL_FEATURE: true
+      EMAIL_FROM: support@the.bot
+      EMAIL_HOST: localhost:3000
+      EMAIL_PROVIDER: postmark
+      POSTMARK_SERVER_API_TOKEN: test-token
```
jlvallelonga, line 36: sadly the stub_features and stub_settings don't seem to be able to reload the code that is run in the initializers so I added this back temporarily. I'd like to create a custom Config lib or something that we can use to load settings and features into during init and then mock values of during tests. I hate to make it so complex though. Maybe there's an easier way that I haven't found yet. Will keep digging, but I feel like something like that should come in a separate PR since this has already been open for a while. Thoughts?

krschacht: That makes sense since initializers are so early in the app boot. At least the `Feature` is available in the initializer. I initially thought that was what you were suggesting and ran a quick test myself locally and Feature is definitely working with initializers, so that's good.

I think you're on the right track of basically fixing this within the test environment, but I don't think the CI script is the place to do it because we want someone to be able to pull down the repo locally, run tests, and have them work without needing to set special ENV. I think the right strategy is to somehow we need to change the options.yml defaults just for the test environment.

How about this PR I just tee'd up: https://github.com/AllYourBot/hostedgpt/pull/479

If I'm fully understanding the problem, I think this PR will allow you to simply add `test_override: "some value"` within config.yml to make the test environment work.

jlvallelonga: #479 will definitely fix the issue with the incorrect values. I think the problem that I'm running into is not being able to change the values that are dynamically loaded in the application.rb config file and the initializers after the app has started. This means we can't modify one of the features or settings during a test and check that the dependent values (`application.config.whatever`) is set and more importantly we're unable to test the logic that depends on those dependent values. For instance, some of the password reset logic is dependent on `config.password_reset_token_purpose` existing which is set dynamically based on the password reset feature being enabled. I can stub the feature so that it's enabled, but that doesn't rerun the initializer code that set the `config.password_reset_token_purpose` (and whatever else) in the first place. I might be able to just modify the config value at runtime too, but I don't think that's the real solution either because I would have to correctly duplicate the logic from the initializer code to set up the app config values the same way they would be set automatically when the feature env vars are present. Having defaults for the test env will get us set up to run the test with a specific combination of settings/features, but won't allow us to change those env vars during a test and then expect the init code to run and change our app config correctly.

Eh that's a long explanation. Ultimately I think the solution is to not do any dynamic logic in the initializers and move that to a Config class or something if we need to have dynamic logic for a config value. That way we can stub the Config methods. I'd say we can keep it simple and just avoid any dynamic logic during init, but I'm not sure we can hold that off forever given that we're aiming at letting people set this up with different service providers for email, etc.

krschacht: Yea, I follow you. Basically, now that we can set different defaults for the test environment options.yml than for the dev/production defaults we should be able to get a test suite that can pass. But you’re right that as soon as we want to have automated tests for postmark **and** some other service, now we’re back to needing to change the Feature flags within a test.

You’re right that it probably means some way that we move the conditionals outside of initializers, but the problem is that we’re setting rails.application.config values. I’m not too familiar with the inner workings of those.

But it’s also possible that when we do implement another mail provider, it may be trivially different such that there isn’t much additional to test. After all, we’ll want every mail provider to work 99% the same. 

I’m okay crossing that bridge when we come to it. We can just defer solving this problem in a deeper way.

test/models/setting_test.rb
```diff
@@ -8,6 +8,8 @@ class SettingTest < ActiveSupport::TestCase
   end
 
   test "should default to nil when feature not found" do
```
krschacht, line 10:
```suggestion
  test "should raise when feature not found" do
```

test/controllers/password_resets_controller_test.rb
```diff
@@ -0,0 +1,44 @@
+require "test_helper"
+
+class PasswordResetsControllerTest < ActionDispatch::IntegrationTest
+  include ActionDispatch::TestProcess::FixtureFile
+
+  setup do
+    @person = people(:keith_registered)
+    users(:keith)
+    credentials(:keith_password)
```
krschacht, line 9: In the setup for some of these test files you are calling a fixture (like these two lines) but not saving the value. Did you mean to save the value? If not, I don’t think that it does anything so you can remove the fixture calls that don’t do an assignment.

test/controllers/password_credentials_controller_test.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  test "should patch update" do
+    token = get_test_user_token(@user)
+
+    patch password_credential_url, params: { token: token, password: "new_password" }
+
+    assert_response :redirect
+    assert_redirected_to root_path
+  end
+
```
krschacht, line 55: In the “should patch update” test above, I think it’s worth asserting that the user’s password actually changed to “new_password”. I forget how exactly to do that since it’s encrypted, but you should be able to reference the actual authentications_controller which logs people in after verifying the password.

Also, let’s add a test for: logged in as one @another_user, but then doing patch to change password for @user using @user’s token, and asserting that the password changed for @user and not only does it redirect you to root_path but you’re now logged in as @user.

app/controllers/password_resets_controller.rb
```diff
@@ -0,0 +1,17 @@
+class PasswordResetsController < ApplicationController
+  require_unauthenticated_access
```
krschacht, line 2: you’ll need to also change this one:
```suggestion
  allow_unauthenticated_access
```

jlvallelonga:
system tests are failing with this error:
```
Error:
Authentications::PasswordTest#test_should_login_as_an_existing_user_and_logout:
Net::ReadTimeout: Net::ReadTimeout with "Net::ReadTimeout with #<TCPSocket:(closed)>"
    test/system/authentications/password_test.rb:6:in `block in <class:PasswordTest>'
```

I made another PR with just a change to the readme and it failed in the same way (https://github.com/AllYourBot/hostedgpt/pull/481 - was troubleshooting so now it has more than just changes to the readme)

Just updating about where I'm at. Will keep digging

krschacht:
Dang. I got this CI failure the other day on my PR and I thought it was an
issue with github infrastructure. I ran tests 3 times and hit the error
each time. Then in the morning I reran the failed tests and it passed.

But I just tried re-running your test suite now and it failed again. Out of
curiosity, I went to main (where tests had passed) and re-ran the test
suite there and it also failed. I’ve found multiple mentions of this issue
online and it’s not encouraging. This sounds complex without a great root
cause fix.

Rob — Do you have any ideas about this error?

This thread was a good discussion of the issue but no resolution, and it’s
many years old:
https://github.com/SeleniumHQ/docker-selenium/issues/198

This issue seemed to explain the root cause is that something during
initialization is taking more than 60 seconds:
https://github.com/teamcapybara/capybara/issues/1305#issuecomment-112535264
<https://github.com/teamcapybara/capybara/issues/1305>

This person suggests that we might want to precompile assets:
https://stackoverflow.com/questions/57546441/net-readtimeout-occurs-in-rails-systemtest

I found a few mentions online with no useful discussion.

krschacht:
Over in this PR https://github.com/AllYourBot/hostedgpt/pull/483 I tried pre-compiling assets and that didn't fix it. I also tried adding a rescue and retry around login_as, since I noticed that's where it was often failing, but the very next time I ran it it failed on a `visit ...` line.

krschacht:
Hi @jlvallelonga where do we stand on this PR? I see tests are passing so we should be really close. I just checked the diff quickly and I see a couple open comments so I didn’t look through the whole thing (e.g. reverting the CI changes and removing the README addition b/c of that other PR). Let me know if you have other to-do’s on this or if it should be done.

jlvallelonga:
> Hi @jlvallelonga where do we stand on this PR? I see tests are passing so we should be really close. I just checked the diff quickly and I see a couple open comments so I didn’t look through the whole thing (e.g. reverting the CI changes and removing the README addition b/c of that other PR). Let me know if you have other to-do’s on this or if it should be done.

Ah sorry. I missed removing that section from the readme. I removed it and the env vars. I'll take another look when I get home tonight, but I think I've got everything now. Ready for merge when you're comfortable with it

krschacht:
I gave it another once over and it all looks great. Merging in now.

PR #476, opened 2024-07-28
Rearrange email settings and add helpers

jlvallelonga:
moves the settings for email around a bit and adds helpers. Will comment further on the lines that changed


app/models/setting.rb
```diff
@@ -7,6 +7,10 @@ def settings
     end
 
     def method_missing(method_name, *arguments, &block)
+      if settings.keys.exclude?(method_name.to_sym)
+        abort "ERROR: no setting found for #{method_name}. Please check settings in options.yml"
```
jlvallelonga, line 11: Still working on getting the Rails.logger working, but maybe this is better than just logging? Should we abort when a setting is asked for that isn't listed as a key in the options.yml file? This would potentially fail during runtime at any point so it might disrupt users, but maybe we want to know sooner rather than later if we've written code against a setting that doesn't exist?

krschacht: I think failing is a good idea. The main thing I think we have to be careful of is adding a new setting which we expect to be there and it's a non-optional setting. That's the one situation where I think we'd break existing deployments, so we probably want to avoid that unless we really need to.

I was not aware of `abort` as a command.

config/initializers/email.rb
```diff
@@ -0,0 +1,31 @@
...
+      config.action_mailer.delivery_method = :postmark
+      config.action_mailer.postmark_settings = { api_token: Setting.postmark_server_api_token }
+    end
+
+    if Feature.password_reset_email?
+      config.password_reset_token_ttl = 30.minutes
+      config.password_reset_token_purpose = :password_reset
+    end
+  end
+end
```
krschacht, line 31: How about we name this file `email_feature.rb` just to better signal the connection

krschacht (approved):
This looks good! Just one suggestion on a filename, but we should land this right after merging in the other branch.

app/models/setting.rb
jlvallelonga: oh this should just check that they keys are in the options.yml file for any settings that we're using with `Setting.`. Since the options.yml file is checked in then it should always contain any keys that we're using with `Setting.` right?

krschacht: Oh, yes, that’s correct.

jlvallelonga:
ok this is based on the other branch so for simplicity's sake I'll go ahead and merge it into that branch since it's all approved and updated

PR #480, opened 2024-08-01
Update Language Models to have a boolean flag supports_tools

### LINKED ISSUE #435: Expand LanguageModel to know two more things about every LLM: context size and if it supports tool function calling
I think the function calling can just be a boolean: `supports_tools`. Inside the ai_backend's we need to make sure that a really long conversation starts skipping older messages so that the submitted payload never exceeds the context size. Maybe this attribute is simply `context_window`, that's what OpenAI and Anthropic both seems to call it.

stephan-buckmaster:
First part supports_tools. Should the migration set support_tools to false for non-OpenAI models?

Fixed #435


test/controllers/settings/language_models_controller_test.rb
```diff
@@ -41,6 +41,31 @@ class Settings::LanguageModelsControllerTest < ActionDispatch::IntegrationTest
...
 
+  test "create should have supports_tools checkbox" do
+    get new_settings_language_model_url
+    assert_response :success
+    assert_select "form" do
+      assert_select 'input[name="language_model[supports_tools]"][checked="checked"]' # Checked by default, from schema
+    end
+  end
+
+  test "edit should have supports_tools checkbox" do
```
krschacht, line 52: small copy edit

```suggestion
  test "edit should have supports_tools checkbox checked" do
```

krschacht:
@stephan-buckmaster Nice PR, this is great! I flagged one small copy edit, and I do agree with your question that we should default things. So two small additions:

1) In `registerable.rb` let's add an explicit boolean flag when we initialize and set them to true and false, appropriately.
  * OpenAI says: `gpt-4o, gpt-4-turbo, and gpt-4o-mini) have been trained to both detect when a function should be called`
  * It looks like all the Anthropic models support tools
  * All the Groq models we're initializing support tools

2) Let's update that migration so that right after adds the column it sets the OpenAI models which don't support tools to false

stephan-buckmaster:
1. Since OpenAI calls this Function Calling, should the check box include that in the label too?

* Supports tools (function calling)


2. When I look over  the OpenAI list,  [Which models support function calling?](https://platform.openai.com/docs/guides/function-calling) only these do *not*:

* gpt-3.5-turbo-instruct and  
* gpt-3.5-turbo-16k-0613

Searching by LLM name and function calling, I find these don't either: mixtral-8x7b-32768, llama3-70b-8192, llama3-8b-8192

3. I am guessing @jlvallelonga's comment above taken care of with the recent merge into main?

4. I'll see you removed the context size from this issue. So it would be complete for now.

krschacht:
1. Good idea, I think like the copy you just wrote since it works in both phrases.
2. For these open-source models it's more confusing as to whether they support function calling because even if the original model didn't inherently support it, later fine-tunings of the model sometimes add it in. In our case, all the open source models that come pre-configured are through Groq and here's their documentation where they claim that they all support it: https://console.groq.com/docs/tool-use
3. Yes, main is merged in with some fixes for system tests so it's worth pulling that into your branch. They still aren't 100% fixed, they're still flaky sometimes and I haven't totally gotten to the bottom of it, but they're better.

stephan-buckmaster:
Ok this could be ready. (Upstream was already merged) 

The UPDATE statement in the db migration could be a bit more specific (has a groq-api-service), hard to tell if users are adding their own language models. I feel it becomes a bit unwieldy as users may have more than one such API service also (with different tokens).

krschacht (approved):
Looks great! Merging in

PR #481, opened 2024-08-03
fix system tests

jlvallelonga:
actually just for the purpose of testing the system tests on a change that should not cause a failure


.github/workflows/rubyonrails.yml
```diff
@@ -70,12 +70,10 @@ jobs:
...
-      - uses: nanasess/setup-chromedriver@v2
-      - name: Start chromedriver
-        run: |
-          set -x
-          chromedriver --url-base=/wd/hub &
-          # sudo Xvfb -ac :99 -screen 0 1280x1024x24 > /dev/null 2>&1 & # optional
+      - name: Setup Chrome
+        uses: browser-actions/setup-chrome@v1.7.2
+        with:
+          install-dependencies: true
```
krschacht, line 76: Interesting, what's the difference between these two ways of doing system tests?

Also, the original way looks a little bit odd. I found the original PR in which Rob touched on this, but I don't fully understand his comment. Maybe this means something to you? https://github.com/AllYourBot/hostedgpt/pull/66

README.md
```diff
@@ -205,6 +205,10 @@ To enable Google OAuth authentication, you need to set up Google OAuth in the Go
      - `GOOGLE_AUTH_CLIENT_ID`: Your Client ID
      - `GOOGLE_AUTH_CLIENT_SECRET`: Your Client Secret
 
+##### Google OAuth Troubleshooting
+
+If you encounter an error when signing in with google on the app (potentially after deleting data from your database) related to "oauth refresh token cannot be blank ... until you remove this website as a third party connection", you can fix this by going to the [Google Third-Party Apps and Services](https://myaccount.google.com/connections) and removing the app from the list of third-party apps with access to your account.
+
```
krschacht, line 211: I too have run into this issue. I had customized the error message a tiny bit to hopefully help, but I know it wasn't really that helpful. Rather than add this to the readme, I think I want to fix this once and for all. Especially since there may be cases when an actual production user runs into it.

There were two issues with the toast: (1) the toast error messages slide on and off the screen quickly so it's hard to put a lot in them, (2) you can't put a link in the toast error messages.

I've been meaning to refactor those toasts into real alert boxes that persist on the screen. I just went ahead and did that now. Check out this PR: https://github.com/AllYourBot/hostedgpt/pull/482

What do you think of this fix? Then we don't have to add instructions in the README.

test/test_helper.rb
```diff
@@ -18,7 +18,7 @@ def exists?
 class ActionDispatch::IntegrationTest
   include Rails.application.routes.url_helpers
 
-  Capybara.default_max_wait_time = 10
+  Capybara.default_max_wait_time = 60
```
krschacht, line 21: I saw this too, but this is the wrong timeout. But I can't find any recent documentation for the right timeout. See:

https://gist.github.com/mikegehard/922296?permalink_comment_id=901298#gistcomment-901298

krschacht: I tried this and it didn't fix it. I'm not sure if it's the right way to set the timeout though since comments I found mentioning this were 6+ years old:
https://github.com/AllYourBot/hostedgpt/pull/483/files#diff-ba37813ca277c227a74a372479b7b05b7f3ff085d890ab708f80d62573efdb7a

jlvallelonga: I think I was able to reproduce the timeout last night using act and running the workflow locally. Will keep investigating today, but wanted to mention that. Not sure if you're using "act" already, but it's nice if you don't want to have to push changes everyone every time you want to test something: https://github.com/nektos/act

krschacht: That's a great find! It didn't even occur to me to try running GH actions locally.

My best thought from earlier investigation was that there probably is indeed some timeout variable deep in capybara we could set to make this go away, what I tried may simply be outdated since the comments were old. It might take inspecting the source of the right gem (copyabara, selenium, rails? not sure which).

But my one new thought  is that the application file loads some assets remotely (from cdn) and when I was on a plane last with really flaky internet the system tests seem to hang forever while trying to request those (although I didn't get that error). I wonder if there's some internal GH action network flakiness which is causing those asset requests to hang. We may need to download those files and add them to the repo.

krschacht:
Justin, two things occurred to me on this flakiness:

1) I was thinking about creating a PR in which I revert the codebase by one whole month, and go back to the smaller runners, to see if things still fail regularly. Because a month ago, there were no issues at all with system tests. They just started one day. This experiment might help isolate if there was something subtle that changed about our codebase or if it's more likely a change with github.

2) I remember seeing long ago in the Rails Guide for System Tests that there is a whole other system you can use besides capybara. I can't remember the name. Unfortuantely, it's probably a lot of work to switch to this because in testes I regularly switch between rails-native assertions and capybara specific stuff (e.g. `find("...").visible?` I think is capybara code and not rails code.

I'm not sure if these are worth doing, but just sharing these both so it's not just in my head.

jlvallelonga:
> Justin, two things occurred to me on this flakiness:
> 
> 1. I was thinking about creating a PR in which I revert the codebase by one whole month, and go back to the smaller runners, to see if things still fail regularly. Because a month ago, there were no issues at all with system tests. They just started one day. This experiment might help isolate if there was something subtle that changed about our codebase or if it's more likely a change with github.
> 2. I remember seeing long ago in the Rails Guide for System Tests that there is a whole other system you can use besides capybara. I can't remember the name. Unfortuantely, it's probably a lot of work to switch to this because in testes I regularly switch between rails-native assertions and capybara specific stuff (e.g. `find("...").visible?` I think is capybara code and not rails code.
> 
> I'm not sure if these are worth doing, but just sharing these both so it's not just in my head.

yeah definitely worth trying I think. I'd hate for us to have to run the tests multiple times for each pr like this. When I was running the tests locally I was having problems with chrome crashing during the run, but that might just be because of using the "act" tool. Seems like there have been a lot of changes lately with the chrome and chromedriver setup for selenium and the options that are required to get it to work right. Also, they charge for time using their hosted runners right? The system tests seem to be occasionally hanging for a long time when they fail so this issue might end up being costly maybe :/

krschacht:
I tried the rollback idea and things did not work. Details over here: https://github.com/AllYourBot/hostedgpt/pull/493

PR #482, opened 2024-08-04
Restyle the toast into an alert

krschacht:
Previously, success and error messages were sliding in from the top right and auto dismissing. The weren't quite prominent enough and sometimes you didn't finish reading the error before it disappeared.

Now I've restyled all of these to be alerts which come down from the top center of the screen and persist with the option of X'ing them out to hide.

Success example:
image

Error example (and smaller viewport):
image


jlvallelonga:
I don't know what's up with the system tests. I tried to fix them in the other PR by changing the chrome-driver version to the latest, but that didn't seem to fix it. Looks like the failures in this PR are different than from there though so I'm not sure what's up. I tried to open a PR with just a change to the readme to see if it was my forgot-password branch, but it still failed in the same way. https://github.com/AllYourBot/hostedgpt/pull/480 passed system tests though, but I'm not sure why the changes there would have caused a pass. I think maybe the tests are just acting inconsistently, but I've rerun the tests multiple times on my branches and it hasn't passed yet haha. I'll keep digging, but wanted to update you about what's going on.

jlvallelonga (approved):

jlvallelonga:
good fix though. I like that more than the readme section. Readmes can get way too long and hard to navigate so definintely prefer this kind of thing when possible

PR #493, opened 2024-08-11
Attempt to fix system tests by rolling back to last known working state

krschacht:


krschacht:
Well, system tests failed with the same issue. And I rolled back every commit to get back to the state where I had just merged in "Hotfix: Timeout Listener sooner & allow repeats"

If you. look at the [commit log](the latest commit was "https://github.com/AllYourBot/hostedgpt/commits/main/), tests passed when that was merged in and for the merge before and after that. So this PR should be in a known good state, yet it's failing. This means whatever changed isn't in our repo. What options does that leave? Either Github Actions server's or maybe not their servers, maybe some CI build dependency which is automatically pulled in (like one of the docker instances or even capybara itself)?

jlvallelonga:
yeah could also be the instance of chrome or the chromedriver maybe. When I was looking I saw something that was pulling in the latest chrome and the matching chromedriver. I know that chrome auto-updates unless you install "Chrome for Testing" which purposefully doesn't auto-update

PR #495, opened 2024-08-20
Add token count per message

krschacht:
This PR is picking up on #460 and just doing some final cleanup. For some reason I'm not able to push to that PR like I normally can so I needed to create a new pull request. Copying over Daniel's description:

Implement token counting per message by using the returned usage stats from the LLM API backend. Anthropic sends token stats already by default. As for OpenAI, the stream needs to be first configured to return those.

- Add migration for new Message attributes input_token_count, output_token_count
- Add concern Message::TokenCount for Message model
- Implement capturing of returned token count from Anthropic & OpenAI and set Message attributes inside the async API-specific jobs. For OpenAI, configure the stream to return usage stats
- Add basic usage dropdown-menu beneath last message that shows the number of input/output tokens when clicked on the "$" sign. This is just a quick demo for the functionality and will likely be changed to something else

This MR is related to #402 

@lumpidu All I did on this one was a little code cleanup and move the message display over to the conversation sidebar. All the update logic "just worked" so I didn't need to do anything fancy to get it to display.


krschacht:
@jlvallelonga I finished testing this and cleaning it up a bit. I'm merging it in now.

PR #498, opened 2024-08-23
adds ssh debugger for selenium test

robacarp:


krschacht:
@robacarp I suspect you'll get what you need from this running in the PR, but if you need to merge into main for testing, go right ahead

krschacht:
@robacarp another big clue in all this is that it all used to work fine, for many months, and then this all started one day. Does that suggest anything else to you or a possible solution?

I even tried rolling back to the point in our commit history where everything was working reliably but it no longer does. I wonder if a dependency has a new version rolled out but if we could pin to an older version this might fix? But rolling back should have done that with gemfile.lock so maybe it’s a dependency that’s somehow outside of that, like part of GitHub action runner config?

krschacht:
Dang, it still seems really odd that for ~6 months these tests never hit this issue and the one day it starts hitting the issue regularly. That's what makes me think that an updated chromedriver or ubuntu-latest system dependency—something outside of gemfile.lock—changed and that introduced this issue. That's the only way I can explain rolling back the repo and still getting the error.

Also, I never get this issue locally on my dev machine. I only get it in github actions. Initially, I suspected this was because my dev machine has more resources (cores, etc) so that's what prompted me to upgrade the github actions server. And notably, that did help. When I was using the default `ubuntu-latest` I ran the test suite 10 times in a row and it failed every time. I don't think I ever got it to pass successfully. But when I upgraded to the larger instances then it's now passes about 50% of the time.

In terms of working around this, would it be possible to study the stack trace and find a place within the selenium or capybara code where you could re-open the class, re-write a method to add a simple "sleep 10" or something hacky like that in the right place within the boot cycle to work around the race?

matthewbennink:
Might be worth looking at Cuprite to see if it may get around the issue. See [this article](https://janko.io/upgrading-from-selenium-to-cuprite/) and [the cuprite project](https://github.com/rubycdp/cuprite).

krschacht:
Hi @robacarp, Matthew's suggestion is an interesting one but I also got another lead. I opened an issue on Selenium and someone explained how they resolved this by pinning a chromium version:
https://github.com/teamcapybara/capybara/issues/2770#issuecomment-2313652897

krschacht:
Just to keep capture this idea, on that other Issue someone claims that he solved a bunch of issues by switching to Playwright and linked to an article: https://github.com/teamcapybara/capybara/issues/2770#issuecomment-2315118283

PR #499, opened 2024-08-23
Add token cost rollups columns to conversation

krschacht:
We already have token input & output count columns on the messages. This PR adds token input & output **cost** onto the messages and it adds aggregate columns for all four of these values on the conversation.


test/models/conversation/billable_test.rb
```diff
@@ -0,0 +1,92 @@
+require "test_helper"
+
+class Conversation::BillableTest < ActiveSupport::TestCase
```
krschacht, line 3: I should also add tests for if any of the values are **updated**

krschacht:
@jlvallelonga Take a look at this and see if it fits your mental model. I'll sleep on this and give it a fresh look tomorrow to see if I think of anything else.

app/models/concerns/rollup_cache.rb
```diff
@@ -0,0 +1,189 @@
+module RollupCache
+  extend ActiveSupport::Concern
+
+  # Rails supports counter_cache on associations, e.g.
+  #
+  # Book.rb
+  #   belongs_to :author, counter_cache: true
+  #
+  # And this keeps author.books_count updatd each time a new author is created, updated, or destroyed.
```
jlvallelonga, line 9:
```suggestion
  # And this keeps author.books_count updated each time a new author is created, updated, or destroyed.
```

krschacht:
Thanks for reviewing. I fixed one bug I discovered with the update case not working (I'm glad I added tests for that). Merging in now so this will be in main for you.

krschacht:
@jlvallelonga Thanks for reviewing. I fixed one bug I discovered with the update case not working (I'm glad I added tests for that). Merging in now so this will be in main for you.

jlvallelonga:
OK thanks. Sorry I didn't get around to looking at it closer. I've been trying to get a good pattern sorted for this stripe integration

krschacht:
Oh it was all good, it didn’t need a close review. The stripe stuff is definitely worth the focus.

PR #501, opened 2024-09-11
Add language model prices

jlvallelonga:
I'd love to refactor the code around this, but this is the quick way of accomplishing this. There might be more to it, but wanted to put this up for discussion first

### DIFF
--- a/app/models/language_model.rb
+++ b/app/models/language_model.rb
@@ -10,6 +10,18 @@ class LanguageModel < ApplicationRecord
     BEST_GROQ => "llama3-70b-8192",
   }
 
+  BEST_MODEL_INPUT_PRICES = {
+    BEST_GPT => 500,
+    BEST_CLAUDE => 300,
+    BEST_GROQ => 59,
+  }
+
+  BEST_MODEL_OUTPUT_PRICES = {
+    BEST_GPT => 1500,
+    BEST_CLAUDE => 1500,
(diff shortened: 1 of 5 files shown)


krschacht:
I think this is a good enough solution. It's the right idea to start simple.

When we update a price of a model, we have to remember to do it in two places (the registration-create place and a migration to update all the old models). There is some slightly different architecture that could have us just update it in one place, but I can't immediately think of an easier one. I'm good just running with this.

If this PR is done feel free to merge in. It looked good to me.

krschacht:
Just merged

jlvallelonga:
Thanks. I was struggling to get the system tests to pass.

krschacht:
Yes, the system tests are still an issue. Rob didn’t completely figure it out. However, he identified the root cause even though he didn’t know how to fix it. Actually this week there were a couple comments on this issue I opened with Capybara: https://github.com/teamcapybara/capybara/issues/2770  Multiple people report the same issue with multiple fixes. The simplest one seems to be pinning to a previous version of the driver, which would explain why things started failing all the sudden. It was always my hunch that something in the CI system got upgraded.

Anyway, I mention this just in case you get around to looking into it before me. The billing thread is still the higher priority, but it’ll be nice when we finally get these system tests fixed so it’s an alternative if you want to switch it up, or sometime next week I’ll try these fixes.

PR #502, opened 2024-09-14
updates precision for message costs

jlvallelonga:
updating so that we can have precision for sub-cent messages (matching precision of token costs on language models)


krschacht:
Looks good to me. If you’re done with it you can go ahead and merge.

app/models/conversation/billable.rb
```diff
@@ -1,10 +1,10 @@
...
   included do
-    attribute :input_token_total_count, :integer, default: 0
-    attribute :output_token_total_count, :integer, default: 0
+    # attribute :input_token_total_count, :integer, default: 0
+    # attribute :output_token_total_count, :integer, default: 0
 
-    attribute :input_token_total_cost, :float, precision: 10, scale: 2, default: 0.0
-    attribute :output_token_total_cost, :float, precision: 10, scale: 2, default: 0.0
+    # attribute :input_token_total_cost, :float, precision: 30, scale: 15, default: "0.0"
+    # attribute :output_token_total_cost, :float, precision: 30, scale: 15, default: "0.0"
```
jlvallelonga, line 8: these attributes are redundant with the schema right? The model should already contain these, but I feel like there was another reason you had them here

krschacht: Good catch. Agreed, I don’t think we need these. Let’s remove them and see if any tests break. 

I never add these unless there’s something unusual that I’m doing with setting a default value, so as long as the schema shows that postgres is handling default then we should be good. I think this was just b/c a contributor added these in a PR and I took over the PR and didn’t think to remove them.

PR #506, opened 2024-09-15
adds ability to add exceptions for logging filters

jlvallelonga:
before:
```
Rails.application.config.filter_parameters += [
  ...
  :token
]
```

->

```
 input_token_count: "[FILTERED]",
 output_token_count: "[FILTERED]",
 input_token_cost: "[FILTERED]",
 output_token_cost: "[FILTERED]",
```

after:

```
Rails.application.config.filter_parameters += [
  ...
  filter_regexp(:token, [:token_count, :token_cost]),
]
```
 ->

```
 input_token_count: 0,
 output_token_count: 0,
 input_token_cost: 0.0,
 output_token_cost: 0.0,
```

### DIFF
--- a/config/initializers/filter_parameter_logging.rb
+++ b/config/initializers/filter_parameter_logging.rb
@@ -1,8 +1,30 @@
 # Be sure to restart your server when you modify this file.
 
+def filter_regexp(filter_key, exemptions = [])
+  filter_key_pattern = Regexp.escape(filter_key.to_s)
+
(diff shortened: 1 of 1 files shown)


jlvallelonga:
I really wanted to use a lambda to handle this, but adding a lambda to the filter_parameters wasn't working as expected so I settled for a regexp

krschacht (approved):
oh it doesn’t like “token”, I didn’t consider that. Good solution. That’s a nice method.

PR #507, opened 2024-09-16
message cost calculations and display

jlvallelonga:
adds calculations for message costs and updates UI to display

image

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -45,6 +45,16 @@ class Message < ApplicationRecord
     !finished?
   end
 
+  def input_token_count=(count)
+    self[:input_token_cost] = assistant.language_model.input_token_cost_cents * count
+    super(count)
+  end
+
+  def output_token_count=(count)
+    self[:output_token_cost] = assistant.language_model.output_token_cost_cents * count
+    super(count)
+  end
+
   private
 
   def create_conversation
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -25,6 +25,8 @@ class Message < ApplicationRecord
   after_create :start_assistant_reply, if: :user?
   after_create :set_last_assistant_message, if: :assistant?
   after_save :update_assistant_on_conversation, if: -> { assistant.present? && conversation.present? }
+  before_save :update_input_token_cost, if: :input_token_count_changed?
+  before_save :update_output_token_cost, if: :output_token_count_changed?
 
   scope :ordered, -> { latest_version_for_conversation }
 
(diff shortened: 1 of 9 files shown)


app/models/message.rb
```diff
@@ -45,6 +45,16 @@ def not_finished?
...
+  def input_token_count=(count)
+    self[:input_token_cost] = assistant.language_model.input_token_cost_cents * count
+    super(count)
+  end
+
+  def output_token_count=(count)
+    self[:output_token_cost] = assistant.language_model.output_token_cost_cents * count
+    super(count)
+  end
+
```
krschacht, line 57: I see what you're doing here, but the more idiomatic way is:

```
before_save :check_age_change, if: :age_changed?

def check_age_change
  # Your code here
end
```

You could do it as two separate callbacks or as one, e.g.:

`before_save :update_token_costs, if: ->{ input_token_cost_changed? || output_token_cost_changed? }`

I'd probably do it as two separate callbacks with two separate helper methods:

```
before_save :update_output_token_cost, if: :output_token_cost_changed?
before_save :update_input_token_cost, if: :input_token_cost_changed?
```

It's probably a little more dangerous to override the setter like this (i.e. there may be some unintended side effects to general activerecord behavior). But my main motivation for doing it is just because it's what other rails devs would expect to see. You're used to looking for hooks at the top of a model to familiarize your self with any surprise "side effects", whereas it's a bit unexpected for the setter to have a ripple like this.

app/views/conversations/_conversation.html.erb
```diff
@@ -81,15 +81,29 @@
             <table class="w-full text-left">
               <tr>
                 <td class="pb-2 whitespace-nowrap">Conversation</td>
-                <td class="pb-2">$</td>
+                <td class="pb-2">$<%= conversation.total_cost / 100.0%></td>
```
krschacht, line 84: `number_to_currency(amount / 100.0)` and it'll make sure 0.5 gets the extra zero 0.50 and it'll do the $ and thousands separators. Likewise on the other ones within this file.

However, we'll lose fractions of a cent. What precision should we show? I think I'd propose for the conversation we use the default, which is precision: 2 if you don't specify. Maybe put a ~ in front of that so your screenshot would have looked like. And then for the input and output costs we use a precision of 5, which is what your screenshot happened to show. I think any more than that and it gets hard to great.

```
Conversation ~   $0.03
Input
Tokens           2102
Cost             $0.001051
Output
Tokens           1174
Cost             $0.001761
```
(I put the tilde at the end of "Conversation" rather than within the dollar `td` so that the number alignment still works.)

app/models/message.rb
krschacht: Let's add a couple unit test for this too. I think we want to prove that setting either of these counts properly updates the cost *and* that setting one of these without setting the other does not zero out the unchanging one.

krschacht (changes requested):
Looking good, a couple thoughts on this one

krschacht (approved):
👍 looks great, merge away

my one suggestion, but this is really nit-pick territory :) is that I would have done `...precision: 2)` so it's a named param which both lets you know when reading the caller what the "5" means and so that it mirrors _to_currency, for consistency. But it's also fine as is!

PR #508, opened 2024-09-22
use playwright instead of selenium

jlvallelonga:
hoping this will fix our system tests...

from conversation here: https://github.com/teamcapybara/capybara/issues/2770#issuecomment-2315118283

PR #509, opened 2024-09-23
upgrade to ruby 3.3.5

jlvallelonga:
trying to fix the issue with the pr that switches from selenium to playwright (#508 )

getting an error with ffi https://github.com/AllYourBot/hostedgpt/actions/runs/10982561553/job/30490969299?pr=508

specifically this
```
#11 5.518 Could not find gems matching 'ffi (= 1.17.0)' valid for all resolution platforms
#11 5.518 (aarch64-linux-musl, arm64-darwin-21, arm64-darwin-22, arm64-darwin-23,
#11 5.518 x86_64-darwin-23, x86_64-linux) in rubygems repository https://rubygems.org/ or
```

saw this comment: https://github.com/ffi/ffi/issues/1105#issuecomment-2160641685

so trying an upgrade in hopes it will fix it


Gemfile.lock
```diff
@@ -141,7 +141,10 @@ GEM
     faraday-multipart (1.0.4)
       multipart-post (~> 2)
     faraday-net_http (3.0.2)
-    ffi (1.17.0)
```
drnic, line 144: I think we still need this one; I found I had to manually add it back in when I tried to docker build. You?

krschacht: I’ve been confused for awhile why sometimes Gemfile lock ends up with multiple versions for different architectures and sometimes it doesn’t. This `ffi` gem looks like it’s still there, but the lock is removing this single one and adding 4 architecture specific ones.

But then the tiktoken_ruby gem below is doing the reverse! This confused me.

Do either of you know which is the “right” way to do it and what causes the flip flopping?

@jlvallelonga was there something you did besides the ruby version change that caused this?

jlvallelonga: oh no I've been reversing this every time I do a bundle install and I missed this one. When I try to start the containers with the different listings in the lock file for each architecture then it fails for me. I started to look it up a while back but didn't get very far. This would be a nice thing to fix for sure. I'll change this part back how it was and hopefully it will work

jlvallelonga:
this seems like the only solution that works: 

https://github.com/ffi/ffi/issues/1105#issuecomment-2356595973

at least until the issue is fixed. Keeping the ruby upgrade because why not? also upgraded tiktoken_ruby to the latest version. If any of these changes cause problems then we can probably just do the `gem "ffi", "~> 1.15.5"` change in the gemfile

Gemfile
```diff
@@ -36,11 +36,12 @@ gem "redcarpet", "~> 3.6.0"
 
 # Use Active Storage variants [https://guides.rubyonrails.org/active_storage_overview.html#transforming-images]
 gem "image_processing", "~> 1.13.0"
+gem "ffi", "~> 1.15.5" # explicitly requiring 15.5 until this is resolved: https://github.com/ffi/ffi/issues/1105
```
jlvallelonga, line 39: this is the only thing that actually fixed it. Maybe there will be another solution, but seems like our target platforms  aren't supported by the latest version of that gem so I don't think we can use it

krschacht (approved):
👍 thanks for chasing this down! It's good you left the comment so we can revisit FFI down the road. It looks like this is ready to merge in? Go ahead, if so. It looks good to me. There are a few other PRs that are throwing this same FFI error so that will help us when we merge main back into them.

PR #511, opened 2024-09-23
Fix autotitle for conversations

ceicke:
Fix autotilte for conversations for OpenAI and Anthropic


krschacht:
Thanks for tackling this! The change looks good so I’ll click merge now and I’ll do some more testing of this in production.

It’s weird that this broke because I distinctly remember confirming that Anthropic & Claude worked without the json directive. It’s supposed to be optional for both those models, it just helps the models more accurately return json. But I fully believe that commenting that out could be the source of the bug and adding this conditional is the safer path. I appreciate you tracking this down.

PR #513, opened 2024-09-29
pinning version of chrome to 127

jlvallelonga:
to fix system tests (temporarily)


jlvallelonga:
hoping a newer version will eventually work and we can remove the hard-coded version. Otherwise maybe playwright or cuprite will fix things for us

krschacht:
Yay!! 🍾 :)

PR #518, opened 2024-10-14
[draft] Support for Kamal deployment

drnic:


Gemfile.lock
```diff
@@ -142,6 +147,10 @@ GEM
       multipart-post (~> 2)
     faraday-net_http (3.0.2)
     ffi (1.17.0)
+    ffi (1.17.0-aarch64-linux-musl)
+    ffi (1.17.0-arm64-darwin)
+    ffi (1.17.0-x86_64-darwin)
+    ffi (1.17.0-x86_64-linux-gnu)
```
krschacht, line 153: Oh shoot, I don't think you mean to change these but I believe we have another PR in flight to fix. I'll make sure to get that merged in ahead of your PR.

README.md
```diff
@@ -111,6 +112,45 @@ Eligible students can apply for Heroku platform credits through [Heroku for GitH
...
+```
+
+Next, uncomment the `database:` section
+
+```yaml
+database:
+  password: some-long-string
+```
+
+Second, create a Docker Hub access token and store it as local env var `KAMAL_REGISTRY_PASSWORD`.
```
krschacht, line 132: I know github lets you host containers. I was using that for a different github project. If I were to enable github container registery for HostedGPT would that save people the trouble of needing to create a Docker Hub account? It must be that Kamal needs to pull from a docker container registery? (I need to watch the Kamal tutorial video)

config/database.yml
```diff
@@ -1,24 +1,27 @@
 default: &default
   adapter: postgresql
   encoding: unicode
-  host: localhost
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  port: <%= ENV['HOSTEDGPT_DATABASE_PORT'] || 5432 %>
-<% if RUBY_PLATFORM =~ /darwin/ %>
+  <% if ENV["HOSTEDGPT_DATABASE_HOST"] %>
```
krschacht, line 5: Rails will already, by default, read DATABASE_URL from the environment and use that as the database configuration instead of what's in database.yml. Does that make it so we don't need to declare all these new ENV variables?

I'm pretty sure other users of HostedGPT have asked for these config so maybe the DATABASE_URL behavior is just too unknown. It probably doesn't hurt to add these regardless.

config/deploy.yml
```diff
@@ -0,0 +1,107 @@
...
+    # When you start using multiple servers, you should split out job processing to a dedicated machine.
+    RUN_SOLID_QUEUE_IN_PUMA: true
+
+    # Set number of processes dedicated to Solid Queue (default: 1)
+    # JOB_CONCURRENCY: 3
+
+    # Set number of cores available to the application on each server (default: 1).
+    # WEB_CONCURRENCY: 2
+
+    HOSTEDGPT_FORCE_SSL: "false"
```
krschacht, line 49: Why disable SSL, was there some complication with Kamal? I thought I remembered hearing that Kamal would auto-configure SSL for you using Let's Authenticate.

lib/templates/rails/credentials/credentials.yml.tt
```diff
@@ -0,0 +1,10 @@
+# Used as the base secret for all MessageVerifiers in Rails, including the one protecting cookies.
+secret_key_base: <%= secret_key_base %>
+
+active_record_encryption:
+  primary_key: <%= SecureRandom.alphanumeric(32) %>
+  deterministic_key: <%= SecureRandom.alphanumeric(32) %>
+  key_derivation_salt: <%= SecureRandom.alphanumeric(32) %>
+
+# database:
+#   password: <%= SecureRandom.alphanumeric(32) %>
```
krschacht, line 10: I see that this template is used when following the instructions to execute `bin/rails credentials:edit --environment production`. But the next instruction is to uncomment these two lines. Should we instead just uncomment them within this template file?

krschacht (changes requested):
A bunch of small clarifying questions as I'm ramping up on this.

README.md
drnic: I think Kamal likes to build a new image and deploy it; whereas docker compose supports running pre-built containers or re-building them.

Specifically:

* the `servers` processes are going to run based on the `image` that is built from the local src project
* the `accessories` processes are run based on existing images.

So each person who wants to use Kamal will be building their own image to their own docker registry (docker hub, or wherever).

config/database.yml
drnic: I think if `DATABASE_URL` or `PRIMARY_DATABASE_URL` env var is provided, then rails ignores the database.yml file. So database.yml is only for users who are not using these env vars.

I thought it'd made sure all the existing behaviour/defaults existed if the new env vars weren't provided. It's definitely difficult to read YAML + ERB :)

config/deploy.yml
drnic: I don't like this either and perhaps it wasn't necessary. I just found I needed to turn off force_ssl for the app running behind kamal proxy. Hopefully we learn more over time about good Rails/Kamal practices.

drnic:
No urgency to see this PR merged - hopefully more people become adept at Kamal and can confirm this PR is "good" :)

config/deploy.yml
duckworth: I have this deployed with kamal and the FORCE_SSL issue was related to the health check. I had to uncomment the config.ssl_options in production.rb:

```ruby
# Skip http-to-https redirect for the default health check endpoint.
config.ssl_options = { redirect: { exclude: ->(request) { request.path == "/up" } } }
```

PR #523, opened 2024-10-21
upgrade rails to 7.2

jlvallelonga:


jlvallelonga:
working towards the defaults for 7.2 and then I'll remove this file: config/initializers/new_framework_defaults_7_2.rb

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_09_15_034529) do
+ActiveRecord::Schema[7.1].define(version: 2024_10_21_060860) do
   # These are extensions that must be enabled in order to support this database
```
krschacht, line 14: Notably there are migrations but the schema file didn't update. Were those migrations no-ops or do you need to run db:migrate still to get the schema updated?

The app's use of ActiveStorage is pretty hacky, since it's all uploaded image files are being stored in ActiveRecord. With these new migrations and this upgrade to 7.2, be sure to test out that functionality in your dev environment to make sure the various AIs can can still answer questions about images. (If  can't remember if I got it working on all three or not, if it doesn't work on Llama it may be that I only implemented it for GPT and Claude)

krschacht (approved):
Skimming this and it looks good. I'm really glad we have so much test coverage on the app to make upgrades like this less risky. :) I left one comment about the migrations, but once you think this PR is good to go feel free to merge in

db/schema.rb
jlvallelonga: the migrations are already in place, I think because we started on 7.1.3 but I wasn't certain if they were doing anything different with the 7.1.4 update so I just left them. I guess since the schema.rb didn't change then they can be removed. I'll check on the ActiveStorage functionality to make sure it's working

jlvallelonga:
> getting this a lot with the 7.2 update: `<ActiveRecord::ActiveRecordError: Cannot expire connection, it is not currently leased.> `

this might be a fix though: https://github.com/rails/rails/issues/53147#issuecomment-2393242969

app/models/api_service.rb
```diff
@@ -7,7 +7,7 @@ class APIService < ApplicationRecord
 
   has_many :language_models, -> { not_deleted }
 
-  enum driver: %w[ openai anthropic ].index_by(&:to_sym)
+  enum :driver, %w[ openai anthropic ].index_by(&:to_sym)
```
jlvallelonga, line 10: the old syntax is deprecated so moving to positional arguments with the enums

jlvallelonga:
ok seems to be working, but we should wait on a release of 7.2 with a fix so we don't have to point to the 7-2-stable branch

PR #525, opened 2024-10-27
Add support for gemini

papayalabs:
This is where I got stuck @krschacht. I need to change app/jobs/get_next_ai_message_job.rb for Gemini to work but that is not the idea. Sorry I do not know to go foward from this point. System instructions does not work on gemini-ai gem. ( I think is a bug ).


krschacht:
Hi @papayalabs, it’s totally fine to make changes to `get_next_ai_message_job.rb` if you needed to make this work. What’s the specific issue you ran into? Are you saying that the issue you can’t resolve is to get system message to work?

app/jobs/get_next_ai_message_job.rb
```diff
@@ -33,7 +33,7 @@ def perform(user_id, message_id, assistant_id, attempt = 1)
 
     response = Current.set(user: @user, message: @message) do
       ai_backend.new(@conversation.user, @assistant, @conversation, @message)
-      .stream_next_conversation_message do |content_chunk|
+      .stream_next_conversation_message.each do |content_chunk|
```
krschacht, line 36: This is a bug. “stream_next_conversation_message” is not an array, it’s a method call. We’re calling:

`ai_backend.new.stream_next_conversation_message` and this method call takes a block so we’re passing in `do … end` to it

app/models/api_service.rb
```diff
@@ -20,7 +21,7 @@ class APIService < ApplicationRecord
   scope :ordered, -> { order(:name) }
 
   def ai_backend
-    openai? ? AIBackend::OpenAI : AIBackend::Anthropic
+    self.url == APIService::URL_GEMINI ? AIBackend::Gemini : (openai? ? AIBackend::OpenAI : AIBackend::Anthropic)
```
krschacht, line 24:
```suggestion
    case driver
    when :openai
      AIBackend::OpenAI
    when :anthropic
      AIBackend::Anthropic
    when :gemini
      AIBackend::Gemini
    end
```

krschacht: I’m pretty sure `.driver` will return a symbol, but double check that. Also, you need to add :gemini as to to the `enum driver` line above

krschacht:
Regarding system instructions, I looked at the gem docs and it appears they think it supports it:
https://github.com/gbaptista/gemini-ai?tab=readme-ov-file#system-instructions

If you try this example in console does do system instructions work there? I’m curious if the issue with system instructions is within the gem itself or an issue getting it to work within this HostedGPT app.

app/jobs/get_next_ai_message_job.rb
papayalabs: Ahh ok now I think I understand this part! I need to write the method  stream_handler for gemini respond @krschacht thanks!

papayalabs:
> Regarding system instructions, I looked at the gem docs and it appears they think it supports it: https://github.com/gbaptista/gemini-ai?tab=readme-ov-file#system-instructions
> 
> If you try this example in console does do system instructions work there? I’m curious if the issue with system instructions is within the gem itself or an issue getting it to work within this HostedGPT app.

I have tried in the console and always got this error: `on_complete': the server responded with status 400 (Faraday::BadRequestError). I think is a problem within the gem but I did not got more deeply into it.

krschacht:
@papayalabs I spent a little while on this system instructions issue today and I couldn’t figure it out. I validated that the docs also confirm system instructions are supported, although it says it’s “in beta” and it’s “on some models”. To confirm it’s not an issue with the gem, I got the API working through curl and I can get normal replies back but I get the same 400 error when I use a system message through curl. I believe the root cause is that we’re using a model that doesn’t support it, or we’re using an API authentication method that doesn’t support it. I was access the API through a key which I got through AI Studio. I thought maybe the key must be obtained through Vertex AI instead but after about 20 minutes of digging I couldn’t figure out how to get a key through there. So anyway, I didn’t solve it but wanted to capture what I learned since there may be clues.

papayalabs:
@krschacht hey! when I first tested throught an API key I have and in the console I also checked all the models ( also the ones that are supported ) and got same result ( error 400 ). When I have more time, I'll dig a little deeper, but I haven't checked it recently.

papayalabs:
@krschacht Also fix the bug. Now the code as is works fine without system instructions. I need to add some tests and check any functionality that may be left.

krschacht:
Hi @papayalabs I'm really excited you're getting this PR done. I gave it a skim and resolved a merge issue with main and did a tiny bit of cleanup. Also, there are some other PRs related to API models that landed in main so merge that back into this branch so it stays kept up.

And I had a thought... I know the system message aren't working for gemini. I just ran into another model, OpenAI o1, which has not yet implemented support for system instructions so those also need to be disabled for that model.

The Language Model table already has booleans for image support and function calling. I think we need another boolean for `supports_system_message`. That way we can have this bit of metadata deeply associated with a language model and our three backends can conditionalize whether they pass that in with their API calls or whether they ignore it.

Would you be willing to tee up a separate PR for this? Landing that would make it easier to land this PR #525 for Gemini because then we'd just disable system message for it.

papayalabs:
Hi @krschacht! I already sync from main. So the PR is updated. I created a new (https://github.com/AllYourBot/hostedgpt/pull/542) for the new boolean in language model. Just need to create the tests.

I was thinking about why it does not work the system message in this gem and maybe is the API Key and as you mention early. Maybe this API Key does not support the gemini-pro model. Is not my API KEY, it is a client API KEY so I need to ask them what plan they have. I came to this conclusion because in Claude they give a 400 error when you do not have credits to run the model, even if you have an API KEY.

krschacht:
@papayalabs I figured it out!! Let me document it:

1. I created an API key through Google AI Studio: https://makersuite.google.com/
2. I confirmed the system prompt works through their interactive developer tool even when I set the model to `Gemini 1.5 Pro 002`. I clicked the "Get code" button and that will show me the cURL for what's entered in the tool. I tried it with curl and it works. The curl was:

```
curl \
  -X POST https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-pro-002:generateContent?key=${API_KEY} \
  -H 'Content-Type: application/json' \
  -d @<(echo '{
  "contents": [
    {
      "role": "user",
      "parts": [
        {
          "text": "Hi, What'\''s your name?"
        }
      ]
    }
  ],
  "systemInstruction": {
    "role": "user",
    "parts": [
      {
        "text": "You are a helpful assistant named Samantha. When I ask your name you should answer with that."
      }
    ]
  },
  "generationConfig": {
    "temperature": 1,
    "topK": 40,
    "topP": 0.95,
    "maxOutputTokens": 8192,
    "responseMimeType": "text/plain"
  }
}')```

3. I then tried this same thing with `gemini-ai` gem and noticed in the output that the URL was different: curl is using `.../v1beta/...` and I figured out that the Gemini gem allows this to be specified with the `version:` key. This is the code that works in rails console:

```
client = Gemini.new(
  credentials: {
    service: 'generative-language-api',
    api_key: ENV['API_KEY'],
    version: 'v1beta'
  },
  options: { model: 'gemini-1.5-pro-002', server_sent_events: true }
)

result = client.generate_content({
  contents: { role: 'user', parts: { text: 'Hi, What is your name?' } },
  systemInstruction: { role: 'user', parts: { text: 'You are a helpful assistant named Samantha. When I ask your name you should answer with that.' } },
})
```

The takeaway is: update this PR to use `version: 'v1beta'` when it calls the API!

**Comment by papayalabs** [comment]
hey @krschacht ! I make some progress. System Instructions is integrated and Tests. But did not dig into Function calls for Gemini and maybe another stuff I am missing. However it works well with System Instructions, Images and all the Tests pass.

**Comment by krschacht** [comment]
@papayalabs Nice work! I just did a pass through things and the code looks good! I think it's ready to merge in. We can always add support for function calling later. Tomorrow I will get this running locally and do a little testing on my dev machine   as one final check.

**Comment by papayalabs** [comment]
Thanks @krschacht! It was a pleasure

**Comment by krschacht** [comment]
I tried this locally and I couldn't get it to work. Initially I was having some configuration issues, which I think I resolved. In the process of doing so I realized there are a few code paths missing (I'll add some inline comments on these). But even after I resolved all of these, it's just hanging for me when I try to get Gemini to respond. I assume you have yours fully working, locally, so that you can chat with it? If so, there must be something different about how you've configured yours. I'll add comments about that too since I think we should include the configuration within this PR.

File: app/models/api_service.rb [hunk 3]
```diff
@@ -2,12 +2,13 @@ class APIService < ApplicationRecord
   URL_OPEN_AI = "https://api.openai.com/"
   URL_ANTHROPIC = "https://api.anthropic.com/"
   URL_GROQ = "https://api.groq.com/openai/v1/"
+  URL_GEMINI = "https://api.gemini.com/v1/"
```
krschacht, line 5: Is this the correct URL? When I was doing a curl test I was using: `https://generativelanguage.googleapis.com/` but this is definitely the URL you're using locally?

Let's add a migration so that everyone gets this new API service automatically created for them. I think the migration would just have an up and look like this (assuming this URL_GEMINI is correct):

**This is what I did, but I'm hoping this is actually mistaken and it's the reason why mine isn't working.**

```
    User.all.each do |user|
      user.api_services.create!(
        name: "Google Gemini",
        driver: "gemini",
        url: "https://api.gemini.com/v1/"
      )
    end
```

Be sure to run this migration in your dev environment and commit the schema change. And then also edit the `models.yml` file and add this to the top of it, assuming all these details are correct:

```
- api_name: gemini-1.5-pro-002
  name: Google Gemini 1.5 Pro (latest)
  supports_images: true
  supports_tools: false
  input_token_cost_cents: '0'
  output_token_cost_cents: '0'
  best: true
  supports_system_message: true
  api_service_name: Google Gemini
  position: 0
```

These two changes will ensure it gets added for everyone when we merge this PR in.

app/services/ai_backend/gemini.rb
```diff
@@ -0,0 +1,116 @@
+class AIBackend::Gemini < AIBackend
+  include Tools
+
```
krschacht, line 3:
```suggestion
  class ::Gemini::Errors::ConfigurationError < ::Gemini::Errors::GeminiError; end

```

app/services/ai_backend/gemini.rb
```diff
@@ -0,0 +1,116 @@
...
+      raise OpenAI::ConfigurationError
+    end
+  end
+
+  def client_method_name
+    :stream_generate_content
+  end
+
+  def configuration_error
+    ::OpenAI::ConfigurationError
```
krschacht, line 37:
```suggestion
    class ::Gemini::Errors::ConfigurationError
```

And inside get_next_ai_message_job catch this error and display a pretty output, you'll see the pattern with the others.

Try removing your Gemini API key from your HostedGPT settings and ensure it displays a pretty message. Also, try intentionally putting a bad Gemini api key into your HostedGPT settings and see if it throws a different exception. You likely need to catch that one too inside get_next_ai_job.

app/services/ai_backend/gemini.rb
```diff
@@ -0,0 +1,116 @@
...
+      ::TestClient::Gemini
+    else
+      ::Gemini
+    end
+  end
+
+  def initialize(user, assistant, conversation = nil, message = nil)
+    super(user, assistant, conversation, message)
+    begin
+      raise ::OpenAI::ConfigurationError if assistant.api_service.requires_token? && assistant.api_service.effective_token.blank?
```
krschacht, line 20:
```suggestion
      raise configuration_error if assistant.api_service.requires_token? && assistant.api_service.effective_token.blank?
```

app/services/ai_backend/gemini.rb
```diff
@@ -0,0 +1,116 @@
...
+      Rails.logger.info "Connecting to Gemini API server at #{assistant.api_service.url} with access token of length #{assistant.api_service.effective_token.to_s.length}"
+      @client = self.class.client.new(credentials: {service: "generative-language-api",
+                                                    api_key: assistant.api_service.effective_token,
+                                                    version: "v1beta"},
+                                      options: { model: assistant.language_model.api_name,
+                                      server_sent_events: true })
+    rescue ::Faraday::UnauthorizedError => e
+      raise OpenAI::ConfigurationError
```
krschacht, line 28:
```suggestion
      raise configuration_error
```

app/services/ai_backend/gemini.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  def get_oneoff_message(instructions, messages, params = {})
+    set_client_config(
+      messages: preceding_conversation_messages,
+      instructions: full_instructions,
+    )
+
+    response = @client.send(client_method_name, @client_config)
+    response.dig("candidates",0,"content","parts",0,"text")
+  end
```
krschacht, line 57: Do we need this method inside here? I think you should remove this method and just update `get_oneoff_message` inside `ai_backend`. I believe all you need to do is add another `||` with this new dig:

```
response.dig("candidates",0,"content","parts",0,"text")
```

app/services/ai_backend/gemini.rb
```diff
@@ -0,0 +1,116 @@
...
+      if Rails.env.test?
+        @client.send(client_method_name, @client_config).each do |intermediate_response|
+          content_chunk = intermediate_response.dig("candidates",0,"content","parts",0,"text")
+          yield content_chunk if content_chunk != nil
+        end
+      else
+        response = @client.send(client_method_name, @client_config) do |intermediate_response, parsed, raw|
+          content_chunk = intermediate_response.dig("candidates",0,"content","parts",0,"text")
+          yield content_chunk if content_chunk != nil
+        end
```
krschacht, line 75: I put a bunch of breakpoints in place and this is where I have issue. Look at this output from my debug. Did you run into this?

```
=> 75:       binding.pry
    76:       response = @client.send(client_method_name, @client_config) do |intermediate_response, parsed, raw|
    77:         binding.pry
    78:         content_chunk = intermediate_response.dig("candidates",0,"content","parts",0,"text")
    79:         yield content_chunk if content_chunk != nil
    80:       end
    81:     end
    82:   rescue ::Faraday::UnauthorizedError => e
    83:     puts e.message
    84:     raise OpenAI::ConfigurationError
    85:   end
    86:   return nil
    87: end

[1] pry(#<AIBackend::Gemini>)> exit
objc[38948]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.
objc[38948]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
```

krschacht (changes requested):
Here are some changes, although I'm not sure this will address whatever underlying reason is causing my Gemini not to work locally. :( I'm hoping I have configured mine incorrectly or that you recognize my error message from your own testing.

app/services/ai_backend/gemini.rb
papayalabs: Yes, I have that problem but I thought it was my pc setup. It is resolved with this line in bin/dev:

  export OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES

I got into the other stuff in this week. Thanks @krschacht !

app/models/api_service.rb
papayalabs: I don't remember where I got that URL from but it's not the correct one. However, the gem does not get the API URL from any source. It just receives the input from the service and builds the URL inside. In this case the service is used: "generative-language-api" with version: "v1beta" so the gem uses this URL inside: "https://generativelanguage.googleapis.com/v1beta/". And if the "vertex-ai-api" service is chosen it makes a more complicated configuration with credentials that are not taken into account right now.

papayalabs:
Hey @krschacht ! I resolved pretty much all the changes. I didn't put any custom Error for Gemini. Just call set_generic_error("Gemini") for now.

Also in my code I do not use ** in my @client.send(client_method_name, @client_config) like you do in AIBackend. Not sure if this will affect the get_oneoff_message method with Gemini.

The only test that does not pass is about images in Gemini but this is because you remove this method: file_base64 from document.rb -->   https://github.com/AllYourBot/hostedgpt/commit/fc7a48287067234834b265b78c106c38ca9cc676#diff-893a2b609f3fc4aade494f624a20993eb6c32598512b5e70d9ad39866b8eb51dL28

krschacht:
@drnic I just updated this PR to use the new models.yml and assistants.yml file. However, while testing this locally, immediately after the new assistant is added I get this exception:

```
No route matches [GET] "/assistants/gemini-1.5-pro-002/messages/new"
```

This relates to changing the URL from using the assistant id to the slug. I'm a bit baffled by this change because I couldn't figure out how it worked in your PR, it felt like maybe a rails internal default. So now I'm equally surprised that it doesn't work with a new assistant. Any ideas?

app/jobs/autotitle_conversation_job.rb
```diff
@@ -31,6 +31,12 @@ def generate_title_for(text)
         response_format: { type: "json_object" }  # this causes problems for Groq even though it's supported: https://console.groq.com/docs/api-reference#chat-create
       )
       return JSON.parse(response)["topic"]
+    elsif ai_backend.class == AIBackend::Gemini
+      response = ai_backend.get_oneoff_message(
+        system_message,
+        [text]
+      )
+      return response.split(":")[1].split("}")[0].gsub(/"/, "").strip
```
krschacht, line 39: Rather than parsing the JSON response manually, it looks like Gemini supports forced JSON responses:
https://github.com/gbaptista/gemini-ai#json-format-responses

Then you can do the same JSON.parse() as the other examples. Before I implemented the JSON mode for the other LLMs this method would randomly break and conversations would stop autotitling.

app/services/ai_backend/gemini.rb
```diff
@@ -0,0 +1,112 @@
...
+    @client_config =  {
+      contents: config[:messages],
+      system_instruction: ( system_message(config[:instructions]) if @assistant.language_model.supports_system_message?)
+    }.compact
+  end
+
+  def get_oneoff_message(instructions, messages, params = {})
+    response = @client.generate_content( { system_instruction: system_message(instructions), contents: { role: "user", parts: { text: messages.first }}})
+    response.dig("candidates",0,"content","parts",0,"text")
+  end
```
krschacht, line 53: Related to my other comment, this method accepts params but it isn't being passed through. That will make this get_oneoff_message() more versatile, such as supporting JSON response type, and it'll make it better match the other models.

krschacht (changes requested):
Really nice work on all the cleanup. This was working locally for me after all your fixes and there is just one last question I have about the code.

However, after I reviewed everything I merged main into this PR and corrected how new assistants and models are created for new users. But after that, this branch stopped working because of a glitch with routes. I tagged someone else who I think can help us with this. I'm a bit stumped by that issue.

papayalabs:
Hi @drnic , I merged your PR but there is a problem in the routes:

new_assistant_message_path
edit_settings_assistant_path

when you try to talk with an assistant

krschacht:
@papayalabs and @drnic that fixed it! Nic — we may need to improve the slug generation code so that it handles recognizes and skips those invalid characters.

@papayalabs — Try the app again. It's working well for me now. If you run into issues still, here is how I did a clean test:

* Switched back to main
* bin/rails db:reset
* bin/dev and signed up to create an account
* checkout this branch again
* ctrl+c on bin/dev and then re-run bin/dev
* Go back to localhost and gemini has now been created for my user. The error message worked.

I think all that is remaining is that json response format, if you're able to figure that out. But it's exciting that it's 100% working now. The only notable thing for me is that Gemini is a bit slower than OpenAI and Anthropic to respond. Once it starts streaming, it's about the same speed, but it's higher latency to start.

papayalabs:
@krschacht It worked for me after your notes. The  json response format is ready. Yes, I have noted that Gemini is more slow to start and sometimes got Faraday::TooManyRequestsError.

I have realized that the correct name to have some consistency for the driver was Google instead of Gemini because you guys use the name of the company for the api protocol and Gemini is the model name.

krschacht:
I gave this a final run through, fixed a couple issues I found, and confirmed that it all seems to be working. I'll merge in now!

@papayalabs on your last PR to rename from GEMINI to GOOGLE, I actually think GEMINI is the better name. Google is not super consistent about this, but when you look at their product page they often consider Gemini to be the umbrella name and "Flash 1.5" to be the model name. But the main reason is that we have Google throughout the code base related to authentication and related to other Google APIs (gmail, etc). We even have another Google model: Google Gemma (through the Groq service). Since all the new Google code you wrote has nothing to do with any of those, I think Gemini is actually clearer. But it was good to think this through.

papayalabs:
Great is it working!! @krschacht. Let's keep it GEMINI. Thanks for all the support!!

PR #526, opened 2024-10-29
Dalle feature

jlvallelonga:
working on what @ceicke started with Dalle


jlvallelonga:
@krschacht This seems to be working. I'm having some trouble with timeouts when referencing generated images in subsequent messages (I think because of my location), but the images are generating. I'm working on trying to test through render to see if having a server in the US helps. If it's working for you though then feel free to merge. I don't have anything else to add on this one

krschacht:
Nice! @jlvallelonga I spent some time on PRs today but I didn't get to this one yet. I will test it out soon and let you know.

ceicke:
@jlvallelonga no, this is not because of your location, I am getting errors as well. So, what works: 

- initial image generation in a new chat
- image generation in an existing chat (without a prior image)

Where it fails:
- image generation after an image has already been generated. I get the error message:

```
"Invalid 'messages[8]'. Image URLs are only allowed for messages with role 'user', but this message with role 'assistant' contains an image URL."
```

ceicke:
Additionally, I get the following errors in the Browser's console:

Bildschirmfoto 2025-02-14 um 08 58 03

ceicke:
Further testing reveals that subsequent image generation works with "gpt-4" but not with "gpt-4o".

PR #532, opened 2024-11-05
Add app_url to the config so workers can create OpenAI image URLs

krschacht:
When sending the conversation history to OpenAI API — regardless of which storage method you're using — we are still embedding all the images within the messages using base64 rather than the messages referencing an image URL.

(Note: PR #465 made it possible to store files (images) in Cloudflare R2 rather than storing them in the database.)

This PR enables the messages to reference URLs rather than requiring base64 encoding. The complication with supporting this is that workers need to generate the URL and until Rails 8.1 workers aren't able to generate URLs.


krschacht:
One thing I don't quite understand is how will this change once the Rails PRs get merged in which you referenced in this comment: https://github.com/AllYourBot/hostedgpt/pull/424#discussion_r1820064897

Will we end up removing these environment variables or changing how the rails config works?

app/models/message.rb
```diff
@@ -43,9 +43,7 @@ def finished?
       (content_text.present? || content_tool_calls.present?)
   end
 
-  def not_finished?
-    !finished?
-  end
+  def not_finished? = !finished?
```
krschacht, line 46: I really like this new feature of ruby

config/application.rb
```diff
@@ -37,6 +37,14 @@ class Application < Rails::Application
     config.time_zone = "Central Time (US & Canada)"
     config.eager_load_paths << Rails.root.join("lib")
 
+    if Setting.validate_env_vars == "1"
+      Setting.require_keys!(:app_url_protocol, :app_url_host, :app_url_port)
+    end
```
krschacht, line 42: I don't love this extra flag (validate_env_vars) which changes whether the keys are required or not. What would it take for us to always make the keys optional? For example, maybe we do this:

1. If any of these vars are set then we require that all 3 are set. But you can leave all 3 blank and the app still works.
2. Would this mean that then within `open_ai.rb` we just add a conditional so that it falls back to file_data_url() if the url_* details are not provided.
3. It would be nice to have a conditional only in a single place so can we avoid also adding a conditional inside `_message.html`? Maybe you make `document_image_url` be smart and it simply return the path if the url_* details are unknown?

config/environments/production.rb
```diff
@@ -98,6 +98,7 @@
   #   /.*\.example\.com/ # Allow requests from subdomains like `www.example.com`
   # ]
   config.hosts = ENV["PRODUCTION_HOST"].split(",").map(&:strip) if ENV["PRODUCTION_HOST"]
+  # config.hosts << Setting.app_url_host
```
krschacht, line 101: Why does development add this to `config.hosts` but production not add it?

docker-compose.yml
```diff
@@ -22,6 +22,10 @@ services:
       target: development
     environment:
       # Be sure to add environment variables to config/options.yml
+      - APP_URL_PROTOCOL
+      - APP_URL_HOST
+      - APP_URL_PORT
+      - VALIDATE_ENV_VARS
```
krschacht, line 28: I was going to say that we need to search the code base for `RUN_SOLID_QUEUE_IN_PUMA` to find all the other places we're handling env config for our production environments, but we won't need to do that if we can make all the keys successfully be optional.

test/models/message/document_image_test.rb
```diff
@@ -16,12 +16,12 @@ class Message::DocumentImageTest < ActiveSupport::TestCase
...
-    assert messages(:examine_this).document_image_path(:small).starts_with?("/rails/active_storage/representations/redirect")
+  test "document_image_url" do
+    assert messages(:examine_this).document_image_url(:small).is_a?(String)
+    assert messages(:examine_this).document_image_url(:small).starts_with?("/rails/active_storage/representations/redirect")
   end
 
-  test "document_image_path with fallback" do
-    assert_equal "", messages(:examine_this).document_image_path(:small, fallback: "")
+  test "document_image_url with fallback" do
+    assert_equal "", messages(:examine_this).document_image_url(:small, fallback: "")
```
krschacht, line 25: If we can make the keys optional, then can we easily test both configurations with automated tests? I think so by using our stubbing technique for options.yml. But if that is tricky for some reason so that we can only have automated tests for one approach, then I think we should have the automated tests for the default config (i.e. all url_* being blank)

jlvallelonga:
> One thing I don't quite understand is how will this change once the Rails PRs get merged in which you referenced in this comment: [#424 (comment)](https://github.com/AllYourBot/hostedgpt/pull/424#discussion_r1820064897)
> 
> Will we end up removing these environment variables or changing how the rails config works?

If they make the changes then we should be able to remove this code: https://github.com/AllYourBot/hostedgpt/pull/532/files#diff-f9c61ca57681712e397b045ba90ceee0d6f073f15b3c1266fa781e6634460d05R155-R159

and then yeah the config will need to change to use the new rails url instead of our custom one and we can remove our env vars too

config/environments/production.rb
jlvallelonga: ah I meant to uncomment this one. Where is PRODUCTION_HOST being created? Maybe it would be better to combine them? Should probably do the same with the DEV_HOST env too?

app/models/message.rb
jlvallelonga: yeah definitely nice to have a shorthand. reminds me of Elixir

config/application.rb
jlvallelonga: yeah it's a bit convoluted as is. I just did that optional piece to avoid it validating during the step to precompile assets. I mainly wanted to have this validation to make the app "fail fast" if the necessary values to create the url weren't present. If we fall back to the path or the base64 path then it should be ok since it still works with base64 urls. I'll update this branch

jlvallelonga:
getting a lot of "Test is missing assertions" warnings during tests. will add a commit to fix those

krschacht:
@jlvallelonga Is this one waiting on me to approve?

config/application.rb
```diff
@@ -37,6 +37,19 @@ class Application < Rails::Application
     config.time_zone = "Central Time (US & Canada)"
     config.eager_load_paths << Rails.root.join("lib")
 
+    url_settings = [:app_url_protocol, :app_url_host, :app_url_port]
+    # if !Rails.env.test? && url_settings.any?{|k| Setting.key_set?(k)}
```
jlvallelonga, line 41:
```suggestion
```

jlvallelonga:
I was going to try to get the config changes in before this one because of an error I think. But I'm not sure that's necessary at this point. I'll try to get this one into a good state today

krschacht:
@jlvallelonga After a bunch of testing, I believe this is the PR that is broke system tests. To replicate my finding:

* On `main` run `bin/rails test test/system/conversations/messages/images_test.rb:88` to see the test fail
* `git revert fc7a482` and resolve the merge conflict. I just clicked "accept current" within docker-compose and manually deleted the three APP_ lines
* Re-run the test and now it passes

The root of the issue has to do with `stimulate_image_variant_processing` at the bottom of this test file (which is a typo, it should be `simulate_....`)

This simulation calls:

```
  def stimulate_image_variant_processing(&block)
    Document.stub_any_instance(:has_file_variant_processed?, false) do
      ActiveStorage::PostgresqlController.stub_any_instance(:show, simulate_not_preprocessed) do
        yield block
      end
    end
  end
```

That very first stub of `has_file_variant_processed?` by setting it to false it used to short-circuit this old method:

```
  def document_image_path(variant, fallback: nil)
    return nil unless has_document_image?

    if documents.first.has_file_variant_processed?(variant)
      documents.first.fully_processed_url(variant)
    elsif fallback.nil?
      documents.first.redirect_to_processed_path(variant)
    else
      fallback
    end
  end
```

So that would make it return the redirect path. But now we are inside this new method:

```
  def image_url(variant, fallback: nil)
    return nil unless has_image?

    if Rails.application.config.x.app_url.blank?
      file_data_url(variant)
    elsif has_file_variant_processed?(variant)
      fully_processed_url(variant)
    elsif fallback.nil?
      redirect_to_processed_path(variant)
    else
      fallback
    end
  end
```

And even though we've stubbed `has_file_variant_processed?` it's checking the `...blank?` which is true and taking that first branch. I thought I could fix this by simply stubbing the blank? to be false, but I'm worried there is a deeper issue. Should we ever be falling back to `file_data_url` when rendering `_messages` ? I don't think we were doing that before, I think we were only using the file_data_url when constructing message history for the API, and we were using real URLs for within _messages. It might be okay but this gave me pause.

buithehien1991:
@krschacht 

I got error when upload with Cloudflare. Here is detail

image

image

image

jlvallelonga:
reverted here in a branch: https://github.com/AllYourBot/hostedgpt/pull/570

tests are failing now for a different reason. I think maybe I didn't resolve the merge conflict in the same way you did @krschacht

PR #536, opened 2024-11-06
Refresh lang models for all users

drnic:
https://github.com/AllYourBot/hostedgpt/discussions/535

In this PR, we provide a rake task `rails models:import` to create/update `LanguageModel` for all `User` from the `models.yaml` file in the root of the repo.

This `models.yaml` file can be manually modified, or can be regenerated from the first User in the local DB with `rails models:export`

* [x] revert local `ffi 1.17.0` change needed for macos dev
* [ ] update `create_initial_assistants_etc`
* [ ] going to wait until #538 merged

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,18 @@ HostedGPT requires these services to be running:
 
 Every time you pull new changes down, kill `bin/dev` and then re-run it. This will ensure your local app picks up changes to Gemfile and migrations.
 
+## Language models
+
+Each User has their own list of Language Models they can use.
+
+### Refreshing language models
+
+There is a shared list of known LLM models for OpenAI, Anthropic, and Groq in `models.yaml` and a Rake task to import them into all users:
+
+```plain
+rails models:import
+```
+
 ### Update models.yaml
 
 The `models.yaml` file in the root of the project is used by HostedGPT applications to refresh their local list of models.
--- a/README.md
+++ b/README.md
@@ -29,14 +29,14 @@ This project is led by an experienced rails developer, but I'm actively looking 
 - [Deploy the app on Heroku](#deploy-the-app-on-heroku)
 - [Deploy on your own server](#deploy-on-your-own-server)
 - [Configure optional features](#configure-optional-features)
-  - [Give assistant access to your Google apps](#configuring-google-tools)
+  - [Configuring Google Tools](#configuring-google-tools)
   - [Authentication](#authentication)
     - [Password authentication](#password-authentication)
     - [Google OAuth authentication](#google-oauth-authentication)
     - [HTTP header authentication](#http-header-authentication)
 - [Contribute as a developer](#contribute-as-a-developer)
-  - [Running locally](#Running-locally)
-    - [Alternatively, you can skip Docker:](#alternatively-you-can-set-skip-docker)
+  - [Running locally](#running-locally)
+    - [Alternatively, you can skip Docker](#alternatively-you-can-skip-docker)
   - [Running tests](#running-tests)
 - [Understanding the Docker configuration](#understanding-the-docker-configuration)
 - [Changelog](#changelog)
@@ -173,6 +173,7 @@ There are a number of optional feature flags that can be set and settings that c
 - `PASSWORD_AUTHENTICATION_FEATURE` is `true` by default, see the [Authentication](#authentication) section for more details.
 - `GOOGLE_AUTHENTICATION_FEATURE` is `false` by default, see the [Authentication](#authentication) section for more details.
 - `HTTP_HEADER_AUTHENTICATION_FEATURE` is `false` by default. If this is set to `true` it automatically disables Password and Google Authentication Features. See the [Authentication](#authentication) section for more details.
+- `LANG_MODELS_URL` overrides the URL for fetching the latest LLM models and pricing.
 
 ### Configuring Google Tools
 
--- a/README.md
+++ b/README.md
@@ -313,6 +313,22 @@ HostedGPT requires these services to be running:
 
 Every time you pull new changes down, kill `bin/dev` and then re-run it. This will ensure your local app picks up changes to Gemfile and migrations.
 
+### Update models.yaml
+
+The `models.yaml` file in the root of the project is used by HostedGPT applications to refresh their local list of models.
+
+To refresh this file using the models in local DB, run:
+
+```plain
+rails models:export
+```
+
+To export the models to another file, either json or yaml, pass in an argument:
+
+```plain
+rails models:export[tmp/models.json]
+```
+
 ### Running tests
 
 If you're set up with Docker you run `docker compose run base rails test`. Note that the system tests, which use a headless browser, are not able to run in Docker. They will be run automatically for you if you create a Pull Request against the project.
--- a/app/models/language_model.rb
+++ b/app/models/language_model.rb
@@ -67,16 +67,20 @@ class LanguageModel < ApplicationRecord
 
   def self.export_to_file(path:, models:, only: %i[api_name name api_service_name supports_images supports_tools input_token_cost_cents output_token_cost_cents])
     path = path.to_s
+    storage = {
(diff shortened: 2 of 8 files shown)


config/options.yml
```diff
@@ -51,6 +51,7 @@ shared:
     default_openai_key: <%= ENV["DEFAULT_OPENAI_KEY"] %>
     default_anthropic_key: <%= ENV["DEFAULT_ANTHROPIC_KEY"] %>
     default_groq_key: <%= ENV["DEFAULT_GROQ_KEY"] %>
+    lang_models_url: <%= ENV.fetch("LANG_MODELS_URL", "https://raw.githubusercontent.com/AllYourBot/hostedgpt/refs/heads/main/models.json") %>
```
krschacht, line 54: For all new features, I've been trying to support them within options.yml. I've really come to like the rails pattern of getting to store secrets and configs in the `bin/rails credentials:edit`. Is that too unconventional?

To continue support of it I think this line would need to be:
<%= ENV["LANG_MODELS_URL"] || default_to(:lang_models_url) || "https://raw.githubusercontent.com/AllYourBot/hostedgpt/refs/heads/main/models.json" %>

models.yaml
```diff
@@ -0,0 +1,219 @@
+---
+models:
+- api_name: gpt-best
+  name: Best OpenAI Model
+  supports_images: true
+  supports_tools: true
+  input_token_cost_cents: '0.00025'
+  output_token_cost_cents: '0.001'
```
krschacht, line 8: Just like you left the quotes off the strings, I'm pretty sure you don't need quotes around floats in a yml file

app/models/language_model.rb
```diff
@@ -47,11 +49,51 @@ def created_by_current_user?
     user == Current.user
   end
 
+  def api_service_name
+    api_service.name
+  end
```
krschacht, line 54: How about `delegate :name, to: :api_service, prefix: true` and then pull it up to the other `delegate` a few lines above

app/models/language_model.rb
```diff
@@ -47,11 +49,51 @@ def created_by_current_user?
...
+      users.each do |user|
+        lm = user.language_models.find_or_initialize_by(api_name: model[:api_name])
+        lm.api_service = user.api_services.find_by(name: model[:api_service_name]) if model[:api_service_name]
+        lm.attributes = model.except(:api_service_name)
+        lm.save!
+      rescue ActiveRecord::RecordInvalid => e
+        warn "Failed to import '#{model[:api_name]}': #{e.message} for #{model.inspect}"
+      end
+    end
+  end
```
krschacht, line 95: Can we put these two methods and the `as_json` into a concern? I think their ancillary enough that they don't need to be in the core model file. I know this model doesn't have any other concerns yet but it's just about ready for it.

lib/tasks/models.rake
```diff
@@ -0,0 +1,15 @@
...
+    models = User.first.language_models.not_deleted
+    LanguageModel.export_to_file(path: args[:path], models:)
+  end
+
+  desc "Import language models to all users from a file, defaulting to models.yaml"
+  task :import, [:path] => :environment do |t, args|
+    args.with_defaults(path: Rails.root.join("models.yaml"))
+    users = User.all
+    LanguageModel.import_from_file(path: args[:path], users:)
+  end
```
krschacht, line 14: good call, it's nice to have these helper tasks

README.md
```diff
@@ -29,14 +29,14 @@ This project is led by an experienced rails developer, but I'm actively looking
 - [Deploy the app on Heroku](#deploy-the-app-on-heroku)
 - [Deploy on your own server](#deploy-on-your-own-server)
 - [Configure optional features](#configure-optional-features)
-  - [Give assistant access to your Google apps](#configuring-google-tools)
+  - [Configuring Google Tools](#configuring-google-tools)
```
krschacht, line 32: Thanks for catching the broken anchors below. But on this change, since our readme has gotten so long, I like the more descriptive "Give assistant access to your Google apps" on this one. It gives people a vague idea of what's possible whereas I don't the more succinct "Configuring Google Tools" hints at enough.

README.md
```diff
@@ -312,6 +313,34 @@ HostedGPT requires these services to be running:
...
+
+```plain
+rails models:export
+```
+
+To export the models to another file, either json or yaml, pass in an argument:
+
+```plain
+rails models:export[tmp/models.json]
+```
```
krschacht, line 342: Thanks for adding this explanation. I am wondering though if we can add some smart defaults so that a new user can safely ignore this and still get reasonable behavior. Can we make it so that `bin/dev` also runs `models:import` on startup? It already runs migrations and bundle upon startup so I've been leaning into it.

We have a fair number of people using this app who are only mildly technical, or are technical but don't know rails, so it's been nice to keep the day-to-day running of it super simple.

But actually, now that I write all this, I don't think people will reliably restart their server. It's not like fly, render, heroku do an automatic nightly restart or anything. Now I'm on the fence... I think I still learn towards doing it on startup, since it's a bit of a catch all. But we'll also add a button to the Settings at some point. And maybe later on we could even add a nightly sync of models:import in the queue. (Scope creep for now, I'm just thinking out loud.)

app/models/language_model.rb
```diff
@@ -4,12 +4,14 @@ class LanguageModel < ApplicationRecord
   BEST_CLAUDE = "claude-best"
   BEST_GROQ = "groq-best"
 
+  # TODO: infer these from models.yaml
```
krschacht, line 7: Agreed, it's nice to eliminate this config

app/models/language_model.rb
```diff
@@ -4,12 +4,14 @@ class LanguageModel < ApplicationRecord
...
   BEST_MODELS = {
     BEST_GPT => "gpt-4o-2024-08-06",
     BEST_CLAUDE => "claude-3-5-sonnet-20240620",
     BEST_GROQ => "llama3-70b-8192",
   }
 
+  # TODO: what are these for?
```
krschacht, line 14: The app now tracks your usage $. Currently, when you click "..." on a conversation you see this:

image

Soon I want to add it so that when you click on your profile pic (button left) then it shows something equivalent for the month-to-date across all conversations.

But... (note to self) these two things should probably be moved to a separate concern of their own. They're not so important as to be so prominent in this model.

krschacht (changes requested):
I know you're still working on this, but I did a quick pass through since I had some time and shared a few thoughts. It's looking good! Thanks so much for tackling this!

models.yaml
drnic: TIL a while back - if you convert `BigDecimal` (`.to_d`) -> YAML, it stores it as a string so it can record large floats; and if its a normal `float` (`.to_f`) -> YAML, then you get a normal float value.

So these strings are perhaps because the original "floating point number" was a BigDecimal

From schema.rb:

```ruby
    t.decimal "input_token_cost_cents", precision: 30, scale: 15
    t.decimal "output_token_cost_cents", precision: 30, scale: 15
```

drnic: (note, I didn't write this YAML file, it was generated from the rake tasks)

README.md
drnic: As always, sorry if my editor's markdown formatter/linter "fixes" things in a way that aren't to be fixed. I'll try to keep any eye out for any regressions in future PRs.

app/models/language_model.rb
drnic: Sounds good.

config/options.yml
drnic: Actually, this idea of a remote URL didn't make it into this PR. Removing.

drnic:
PR being recreated off `main` in https://github.com/AllYourBot/hostedgpt/pull/556?w=1

models.yaml
krschacht: oh interesting, that’s a good lesson for me

PR #537, opened 2024-11-06
Bump rubocop TargetRubyVersion: 3.3

drnic:
Also, start upgrading syntax to use implicit variables in named parameters

Note, we could also bump rubocop related gems; but my environment cannot install ffi 1.15.5; so I'm a bit stuck touching `Gemfile.lock`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,7 +9,7 @@ require:
   - rubocop-performance
 
 AllCops:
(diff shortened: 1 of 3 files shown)


krschacht (approved):
This is great, thanks for this change! I haven't gotten used to these implicit params yet, but it's a good push. And I'm already liking the new one-line methods and `it` syntax in blocks. I love ruby. :)

PR #538, opened 2024-11-06
Dynamic "best" Language Model

drnic:
Instead of constants in code, adding a `LanguageModel#best?` field. Only one LanguageModel per APIService can be the best.

Best models are at top of table, then ordered by position:

Screenshot 2024-11-07 at 8 24 09 am

Toggle a language model to be the new "best" model for that API Service:

Screenshot 2024-11-07 at 8 16 45 am

The language model used for an assistant is now clear, rather than "Anthropic Best", etc:

Screenshot 2024-11-07 at 8 02 20 am

### DIFF
--- a/app/controllers/settings/language_models_controller.rb
+++ b/app/controllers/settings/language_models_controller.rb
@@ -3,7 +3,7 @@ class Settings::LanguageModelsController < Settings::ApplicationController
   before_action :set_system_language_model, only: [:show]
 
   def index
-    @language_models = LanguageModel.for_user(Current.user).order(updated_at: :desc)
+    @language_models = LanguageModel.for_user(Current.user).ordered
   end
 
   def edit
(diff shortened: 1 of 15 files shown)


drnic:
PR rebased against `main` and passing CI

drnic:
@krschacht bump; this PR might have gotten lost amongst my other PRs last week

krschacht:
@drnic Thanks for the bump and sorry for the delay. I just gave it a close review and it looks great. I tested it out locally and the only issue I found was that I couldn't run db:migrate from scratch. I'm not certain that it's important to support this, but sometimes I've seen that catch a dev off guard so I fixed that. But it doesn't look like I can push to your branch so I just hotfixed this into main (https://github.com/AllYourBot/hostedgpt/commit/5e967ece48b27e8ed13e1e299c9778ebab961465)

The one other thing I noticed is that it possible to go into settings and uncheck the "best" one leaving no best. I worried this might cause a problem, but my existing assistants still work and new users can still register so I think we're fine. But I'm calling this out in case it raises any red flags for you. Keep it in mind in your other PR because if that becomes a problem then we may need to add some guards for it.

But overall great work!

PR #539, opened 2024-11-10
Introduce login with Microsoft

drnic:
* Introduces a second oauth login
* README describes how to turn it on and configure it
* Access the auth token via `User#microsoft_graph_credential.oauth_token` to interact with Microsoft Graph API
* If microsoft account does not have a full name, login/signup will fail. Future work might allow a user to provide these missing details in hostedgpt.
* Future work: refactor out common code + routes with google auth

Sequence flow:

Screenshot 2024-11-10 at 1 57 10 pm
Screenshot 2024-11-10 at 1 57 25 pm
Screenshot 2024-11-10 at 2 10 54 pm
Screenshot 2024-11-10 at 2 21 17 pm


app/models/user.rb
```diff
@@ -18,6 +18,7 @@ class User < ApplicationRecord
   has_one :google_credential, -> { type_is("GoogleCredential") }, class_name: "Credential", inverse_of: :user
   has_one :gmail_credential, -> { type_is("GmailCredential") }, class_name: "Credential", inverse_of: :user
   has_one :google_tasks_credential, -> { type_is("GoogleTasksCredential") }, class_name: "Credential", inverse_of: :user
+  has_one :microsoft_graph_credential, -> { type_is("MicrosoftGraphCredential") }, class_name: "Credential", inverse_of: :user
```
krschacht, line 21: I’ve been adding simple tests for these associations. To be consistent with the current test suite, I believe that `models/user_test` should get a test for user.microsoft_graph_credential. Also, let’s add a microsoft_graph_credential_test.rb file (you can look at the google one).

config/options.yml
```diff
@@ -57,6 +58,9 @@ shared:
     cloudflare_bucket: <%= ENV["CLOUDFLARE_BUCKET"] || default_to(cloudflare: :bucket) %>
     google_auth_client_id: <%= ENV["GOOGLE_AUTH_CLIENT_ID"] || default_to(:google_auth_client_id) %>
     google_auth_client_secret: <%= ENV["GOOGLE_AUTH_CLIENT_SECRET"] || default_to(:google_auth_client_secret) %>
+    microsoft_graph_auth_client_id: <%= ENV["MICROSOFT_GRAPH_AUTH_CLIENT_ID"] || default_to(:microsoft_graph_auth_client_id) %>
+    microsoft_graph_auth_client_secret: <%= ENV["MICROSOFT_GRAPH_AUTH_CLIENT_SECRET"] || default_to(:microsoft_graph_auth_client_secret) %>
+    microsoft_graph_scope: <%= ENV["MICROSOFT_GRAPH_SCOPE"] || default_to(:microsoft_graph_scope) %>
```
krschacht, line 63: Let’s do: `microsoft: :graph_scope` and similarly for the others, then when someone does bin/rails credentials:edit they can nest all these subkeys under microsoft, e.g.:
```
…
microsoft:
  auth_client_id: xyz
  auth_client_secret: xyz
  graph_scope: xyz
```

(I wish I would have done that for the google keys… I think I added this nesting feature when I was working on cloudflare, right after the google feature. I need to refactor the google one in a way that won’t break it for anyone who has already configured it.)

config/initializers/omniauth.rb
```diff
@@ -26,4 +26,10 @@
     include_granted_scopes: true
   }
 
+  provider :microsoft_graph, Setting.microsoft_graph_auth_client_id, Setting.microsoft_graph_auth_client_secret, {
+    name: "microsoft_graph",
+    scope: Setting.microsoft_graph_scope.presence || %|
+      openid profile email offline_access user.read mailboxsettings.read
+    |
```
krschacht, line 33: Oh interesting, I like how you made the scopes part of the config. That’s a good improvement we could make to google, in the future.

app/views/authentications/new.html.erb
```diff
@@ -52,10 +73,10 @@
...
-  <% if Feature.password_reset_email? %>
-    <span class="text-sm text-center">
-      <%= link_to "Forgot your password?", new_password_reset_path, class: "underline" %>
-    </span>
-  <% end %>
+<% end %>
+<% if Feature.password_reset_email? %>
+<span class="text-sm text-center">
+  <%= link_to "Forgot your password?", new_password_reset_path, class: "underline" %>
+</span>
```
krschacht, line 80: Did you mean to remove this layer of indentation inside the conditionals? My instinct would be to keep it, but was it causing some issues?

app/views/users/new.html.erb
```diff
@@ -94,7 +113,7 @@
       |,
       data: { turbo_submits_with: "Creating..." }
     %>
-  <% end # password_authentication? %>
+<% end # password_authentication? %>
```
krschacht, line 116: same indentation question here. I don’t feel that strongly, but mostly I’d keep it just for consistency across the code base. I could easily see another contributor coming along and re-indenting all these in a future PR. :)

krschacht (changes requested):
This is a great addition. You’ve been the first person willing to dig into my credentials code since I wrote it all. :) I was pretty pleased with how this turned out when I did the initial refactor that added a bunch of different methods. I’m curious you’re take now that you’ve made sense of it all.

On this specific PR, what you did looks really good. One overall question: should we call it `microsoft_*` or `microsoft_graph_*`? My instinct would have been to just use the single word, microsoft, but googling I do see people refer to it both ways. Or maybe these are two different systems? I’m not as familiar with the Microsoft ecosystem but I do clearly see references to Microsoft Graph so I trust your context more than mine and am happy to follow your lead on this. I just wanted to raise this question for reflection since it’ll be hard to change all the references later.

The main thing I’m flagging is just a couple things on tests. I haven’t tried setting this up for myself yet on my machine. I’ll try to do that later today and I’ll give the README a close review as I’m doing this to make sure all the steps work.

config/options.yml
drnic: I'll discuss the "microsoft graph" naming here. It seems to be the name microsoft gives to their API service [1], and `microsoft_graph` is the named used by the omniauth provider implementations [2] [3] with names `microsoft_graph` and `microsoft_graph_auth` respectively [4] [5]

So we'll go with `microsoft_graph` as the full name everywhere. One name everywhere for the same thing. I had mix/match of `microsoft` and `microsoft_graph` during development and it led to confusion + failing tests.

But the login button is "Login with Microsoft" for public consumption since no human knows "Microsoft Graph" :) 

* [1] https://learn.microsoft.com/en-us/graph/overview
* [2] https://rubygems.org/gems/omniauth-microsoft_graph
* [3] https://github.com/microsoftgraph/msgraph-sample-rubyrailsapp/blob/main/graph-sample/lib/microsoft_graph_auth.rb (an alternate implementation bundled inside a microsoft sample repo
* [4] https://github.com/synth/omniauth-microsoft_graph/blob/main/lib/omniauth/strategies/microsoft_graph.rb#L10
* [5] https://github.com/synth/omniauth-microsoft_graph/blob/main/lib/omniauth/strategies/microsoft_graph.rb#L10

drnic: I'll investigate supporting the nested credentials. I didn't know we had that, so will add support for it.

drnic: I'll be aiming for:

```yaml
microsoft_graph:
  auth_client_id:
  auth_client_secret:
  scope: ...
```

drnic: Ok, I added support for google + microsoft nested credentials.

```yaml
    google_auth_client_id: <%= ENV["GOOGLE_AUTH_CLIENT_ID"] || default_to(:google, :auth_client_id) || default_to(:google_auth_client_id) %>
    google_auth_client_secret: <%= ENV["GOOGLE_AUTH_CLIENT_SECRET"] || default_to(:google, :auth_client_secret) || default_to(:google_auth_client_secret) %>
    microsoft_graph_auth_client_id: <%= ENV["MICROSOFT_GRAPH_AUTH_CLIENT_ID"] || default_to(:microsoft_graph, :auth_client_id) %>
    microsoft_graph_auth_client_secret: <%= ENV["MICROSOFT_GRAPH_AUTH_CLIENT_SECRET"] || default_to(:microsoft_graph, :auth_client_secret) %>
```

app/models/user.rb
drnic: Thanks, done.

krschacht:
@drnic I hoped to finish reviewing today but I didn't make it. I should be able do this tomorrow.

app/controllers/authentications/microsoft_graph_oauth_controller.rb
```diff
@@ -0,0 +1,96 @@
+class Authentications::MicrosoftGraphOauthController < ApplicationController
+  allow_unauthenticated_access
+
+  def create
+    if Current.user
+      Current.user.microsoft_graph_credential&.destroy
+      _, cred = add_person_credentials("MicrosoftGraphCredential")
+      cred.save! && redirect_to(edit_settings_person_path, notice: "Saved") && return
+
```
krschacht, line 9: I don't think you're ever going to hit this first conditional, are you? The Google Oauth method has this condition because it's used for two things: logging in and for giving the AI access to your google inbox, calendar, tasks. But right now Microsoft is only used for logging in.

If there is a Current.user, you'll never be able to see the login page, so you'll never be able to submit to this `create` endpoint, right?

app/controllers/authentications/microsoft_graph_oauth_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    end
+
+    if @person&.save
+      login_as(@person, credential: @person.user.reload.microsoft_graph_credential)
+      redirect_to root_path
+    else
+      @person&.errors&.delete :personable
+      msg = @person.errors.full_messages.map { |m| m.gsub(/Personable |credentials /, "") }.to_sentence.capitalize
+      if msg.downcase.include?("oauth refresh token can't be blank")
+        msg += " " + helpers.link_to("Microsoft third-party connections", "https://microsoft.com/consent", class: "underline") + " search for website, and delete all it's connections. Then try again."
```
krschacht, line 33: Are you sure this link shouldn't be https://account.microsoft.com/privacy/app-access ? That's what you put in the READMe. I'm not sure which one is the easier path for a user to take.

test/controllers/authentications/microsoft_graph_oauth_controller_test.rb
```diff
@@ -0,0 +1,182 @@
...
+class Authentications::MicrosoftGraphOauthControllerTest < ActionDispatch::IntegrationTest
+  setup do
+    OmniAuth.config.test_mode = true
+  end
+
+  test "cancelling a microsoft oauth flow redirects to the settings page" do
+    login_as users(:keith)
+    get "/auth/failure"
+    assert_redirected_to edit_settings_person_path
+  end
```
krschacht, line 12: We should definitely remove the first of these two tests (logged in yet hitting /auth/failure). That is related to the same `if Current.user` comment I made within the controller. It's a flow which is unique to google.

test/controllers/authentications/microsoft_graph_oauth_controller_test.rb
```diff
@@ -0,0 +1,182 @@
...
+
+  test "cancelling a microsoft oauth flow redirects to the login" do
+    get "/auth/failure"
+    assert_redirected_to login_path
+  end
```
krschacht, line 17: Does Microsoft ever redirect here? If you look at the first block `omniauth.rb`, that's where google auth configures the failure path. I don't see an equivalent configuration of failure path for Microsoft.

If you start logging in or signing up with Microsoft oauth and then you abort on the Microsoft confirmation step, what happens? I don't think it'll redirect here unless we tell it, but I'm curious where it goes by default.

test/controllers/authentications/microsoft_graph_oauth_controller_test.rb
```diff
@@ -0,0 +1,182 @@
...
+    }
+    OmniAuth.config.add_mock(:microsoft_graph, details)
+
+    refute user.microsoft_graph_credential.present?
+    get microsoft_graph_oauth_path
+
+    assert_redirected_to edit_settings_person_path
+    assert_equal "Saved", flash[:notice]
+    assert_credential_matches_details(user.reload.microsoft_graph_credential, details)
+  end
```
krschacht, line 40: This is the same `if Current.user` settings-page case that we're not supporting so we should remove this test.

test/controllers/authentications/microsoft_graph_oauth_controller_test.rb
```diff
@@ -0,0 +1,182 @@
...
+
+  test "should log you in for a user that exists" do
+    OmniAuth.config.add_mock(:microsoft_graph, { uid: credentials(:keith_microsoft_graph).oauth_id }) #, info: { email: "gmail@address.com", first_name: "John", last_name: "Doe" }})
```
krschacht, line 43:
```suggestion
    OmniAuth.config.add_mock(:microsoft_graph, { uid: credentials(:keith_microsoft_graph).oauth_id })
```

krschacht (changes requested):
I finished giving it a careful review and flagged a few final cleanup items but they're all minor.

Unfortunately, I had trouble getting into my Azure account. I haven't used it for a long time. So, I didn't actually try initializing the whole thing on my account but I'm trusting that you tried this out. :) Aside from that, I looked closely at all the code and it's all sane.

Take a look at my final comments and then I think we should be ready to merge.

README.md
```diff
@@ -196,6 +197,7 @@ HostedGPT supports multiple authentication methods:
 
 - [Password authentication](#password-authentication)
 - [Google OAuth authentication](#google-oauth-authentication)
+- [Microsoft OAuth authentication](#microsoft-oauth-authentication)
```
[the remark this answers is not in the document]
drnic: Ooh nested TOCs

app/controllers/authentications/microsoft_graph_oauth_controller.rb
drnic: "Never" is strong. Someone might have the Login page open on two tabs (say after being automatically logged out overnight or similar), and come back and click Login on each one.

I'll update this to at least bail out of the `create` if the user is already logged in:

```ruby
    if Current.user
      redirect_to(edit_settings_person_path, notice: "Already logged in") && return
```

app/controllers/authentications/microsoft_graph_oauth_controller.rb
drnic: For anyone interested here is where the two links go, similar but different.

1. https://account.microsoft.com/privacy/app-access
2. https://microsoft.com/consent

Screenshot 2024-11-14 at 6 52 15 am

The suggested former one looks better; will switch.

test/controllers/authentications/microsoft_graph_oauth_controller_test.rb
drnic: As above, I believe this is a useful flow for an existing User adding a connection/credentials to Microsoft.

test/controllers/authentications/microsoft_graph_oauth_controller_test.rb
drnic: Good question.

Cancelling login/auth looks to redirect to: `/auth/microsoft_graph/callback?error=access_denied&error_description=The%20user%20has%20denied%20access%20to%20the%20scope%20requested%20by%20the%20client%20application.&state=<state>`.

Adding/fixing tests.

drnic: Done

app/controllers/authentications/microsoft_graph_oauth_controller.rb
krschacht: Oh dang, great catch. And very true about “never” :)

krschacht:
Thanks for chasing down all these details! It looks great, merging now.

PR #540, opened 2024-11-10
Ruby {key: key} -> {key:} simplification

drnic:


lib/active_storage/service/postgresql_service.rb
```diff
@@ -7,22 +7,22 @@ def initialize(public: false, **options)
...
           ActiveStorage::Postgresql::File.open(key) do |file|
             while data = file.read(5.megabytes)
               yield data
             end
           end
         end
       else
-        instrument :download, key: key do
+        instrument :download, key: do
           ActiveStorage::Postgresql::File.open(key) do |file|
```
krschacht, line 26: wow, it actually surprises me the ruby parser can parse this properly. i would have assumed this would need parens

krschacht (approved):
This is great. It’s times like this when I really appreciate having lots of test coverage. :) Now I just need to automatize writing this new syntax. :)

PR #541, opened 2024-11-10
Fix 'Test is missing assertions' warnings

drnic:


krschacht (approved):
Great cleanup, thank you

PR #544, opened 2024-11-13
User can only have 2 initials in UI

voodoo:
Addresses:  "You should create a method that takes user.name.initials as its argument. We are already properly extracting initials, we just want fewer of them. So make a helper like you’ve done that takes in all initials and returns a max of two. If there are more than two, keep the first and last initial and drop any in the middle."

I think this kind of works - but I've not done this type of process in a while so I expect it is not perfect ... yet.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -437,6 +437,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-23
+  arm64-darwin-24
   x86_64-linux
 
 DEPENDENCIES
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -437,7 +437,6 @@ GEM
 
 PLATFORMS
   arm64-darwin-23
-  arm64-darwin-24
   x86_64-linux
 
 DEPENDENCIES
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,21 @@
 module ApplicationHelper
+
+  def only_user_initials(initials, limit: 2)
+    words = initials.to_s.strip.split(/\s+/)
+    case words.length
+    when 0
+      ""
+    when 1
+      words.first[0].to_s.upcase
+    else
+      if words.length <= limit
+        words.map(&:first).join.upcase
+      else
+        [words.first[0], words.last[0]].join.upcase
+      end
+    end
+  end
+
   def spinner(opts = {})
     html = <<~HTML
       <svg class="animate-spin -ml-1 mr-3 h-#{opts[:size]} w-#{opts[:size]} #{opts[:class]}" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24">
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,17 +1,17 @@
 module ApplicationHelper
 
   def only_user_initials(initials, limit: 2)
-    letters = initials.to_s.strip.split("")
+    letters = initials.to_s.downcase.strip.split("")
     case letters.length
     when 0
       ""
     when 1
-      letters.first.to_s.upcase
+      letters.first
     else
       if letters.length <= limit
-        words.map(&:first).join.upcase
+        letters.map(&:first).join
       else
-        [letters.first, letters.last].join.upcase
+        [letters.first, letters.last].join
       end
     end
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,7 +2,7 @@ module ApplicationHelper
 
   def at_most_two_initials(initials)
     return initials if initials.nil? || initials.length <= 2
-    initials[0] + initials[-1]  
+    initials[0] + initials[-1]
   end
 
   def spinner(opts = {})
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,17 +1,17 @@
 module ApplicationHelper
 
   def only_user_initials(initials, limit: 2)
-    words = initials.to_s.strip.split(/\s+/)
-    case words.length
+    letters = initials.to_s.strip.split("")
+    case letters.length
     when 0
       ""
     when 1
-      words.first[0].to_s.upcase
+      letters.first.to_s.upcase
     else
-      if words.length <= limit
+      if letters.length <= limit
         words.map(&:first).join.upcase
       else
-        [words.first[0], words.last[0]].join.upcase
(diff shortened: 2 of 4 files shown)


voodoo:
I'm looking to make a 'good' PR. So go over with a fine tuned comb and shout. I already see a couple things I would like to have done better.

Gemfile.lock
```diff
@@ -437,6 +437,7 @@ GEM
 
 PLATFORMS
   arm64-darwin-23
+  arm64-darwin-24
```
krschacht, line 440: I don’t know what this line is doing, do you? I’m not opposed to adding it if it’s helping something, but if it’s just a stray line we should remove it.

app/views/layouts/_user_avatar.erb
```diff
@@ -11,5 +11,5 @@
         <%= classes %>
         "
 >
-  <%= user.name.initials&.upcase %>
+  <%= only_user_initials(user.name.initials).upcase %>
```
krschacht, line 14: The method name is a bit confusing. I find myself thinking, “but isn’t name.initialis” only the user initials? I don’t expect that to return something other than initials.

A better name might be: “only_two_initials”

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,21 @@
 module ApplicationHelper
+
+  def only_user_initials(initials, limit: 2)
+    letters = initials.to_s.downcase.strip.split("")
```
krschacht, line 4: I suspect this whole method can be a lot simpler. Probably this will work:

def …
  return initials if initials.nil? || initials.length <= 2
  initials[0] + initials[-1]
end

I haven’t tested that so I may be overlooking something, but I think that’s a simple way to grab the first letter and the last letter.

app/views/layouts/_user_avatar.erb
krschacht, line 14: Also, I think I’d keep the &.upcase rather than removing &, at least the method the way I wrote it returns a nil if it’s passed a nil, which seems like reasonable behavior

test/helpers/application_helper_test.rb
```diff
@@ -0,0 +1,27 @@
+require "test_helper"
+
+class ApplicationHelperTest < ActionView::TestCase
+  
+test "only at most 2 initials with default limit" do
+    assert_equal "jz", only_user_initials("jdz")
+    assert_equal "pl", only_user_initials("Plllll")
```
krschacht, line 7: It’s a bit unexpected that this method would change case. The way I wrote it preserves whatever case was passed in. I think that’s what most users would expect.

Also, this second test is bad because the fact that it returns “pl” does not confirm that it’s returning the first letter and the last letter. There are a lot of “l”s

test/helpers/application_helper_test.rb
```diff
@@ -0,0 +1,27 @@
...
+  end    
+
+  test "can have numbers" do
+    assert_equal "p2", only_user_initials("Pvv2")
+  end  
+
+  test "can have spaces" do
+    assert_equal "pq", only_user_initials("P vv qQ")
+  end  
+
```
krschacht, line 26: It’s worth adding a test for the nil case

krschacht (changes requested):
Alright, you asked for it so I gave it a full review. :)

Gemfile.lock
voodoo: That was added automatically. I just assumed cause I'm on a mac m1. I can take it out ... maybe :) I'll look into git-ing it out.

voodoo:
Ok, I added those awesome/needed/helpful changes. I appreciate you doing that.

voodoo:
DAMN! Did no mean to CLOSE the PR.

voodoo:
These code versioning tools are awesome but I  feel like I'm running with scissors.  I might have done something wrong - but I learn. Really, make any other observations.

app/views/layouts/_user_avatar.erb
voodoo: only_two_initials makes great sense

voodoo: &.upcase on the method.

test/helpers/application_helper_test.rb
voodoo: No longer changing case. Took out a few l's.

app/views/layouts/_user_avatar.erb
voodoo: Will do.

test/helpers/application_helper_test.rb
voodoo: True, does not change case now.

voodoo:
I appreciate your patience and your coding skills. You can delete. I'll do better on the next one.

Gemfile.lock
voodoo: I removed the arm64-darwin-24. I think lol. Thanks for your patience.

voodoo: Running the test, it tries to stuff it back in there.

app/views/layouts/_user_avatar.erb
voodoo: only_two_initials was an even better name but I went with at_most_two_initials to cover the case for one initial

voodoo:
Hopefully you have not ghosted or unsubscribed or something.  I've not done a pull request in about a decade. That does not explain my carelessness - but I do care!

krschacht:
Hey! A busy couple days at my day job, but I’ll definitely circle back soon. Sorry for the delay!

voodoo:
No worries Keith  - I appreciate you letting me know.

Gemfile.lock
krschacht: Some quick googling suggests that you may be on a newer version of the Mac operating system and that’s why it’s added, but I didn’t dig into it much. It’s probably fine to leave it but it’s also fine to take it out.

krschacht (approved):
All your changes were good ones. I did some final cleanup and as soon as the test suite passes I’ll merge this in. Congrats on your first PR!

You can look at the commit log to see my final changes. Notable takeaways:

1) Your code editor is letting you save files with trailing whitespace. In the root of this project is an .editorconfig file. You should configure your editor to use this file and then it’ll always use the proper editor settings for whatever project you’re working on.

2) When you finish making changes locally, you can run both `bin/rails test:all` and also `bin/rubocop`. That latter one checks all the code formatting to catch issues like whitespace.

3) When you cleaned up the tests, there was one test you removed which was important to keep: a test of greater than 2 initials to ensure it only returns 2. :) I added that one back in.

PR #547, opened 2024-11-14
Improve database config

jlvallelonga:
This PR does two things:

1. DATABASE_URL looked like it was supported in docker compose but it wasn't actually working. We've officially decided NOT to support DATABASE_URL within docker compose and indicated that in comments
2. Add additional ENV variables to support overwriting of all the pieces of database config for running dev in plain Rails and for running dev from Docker compose

- [x] Remove the files we don't need
- [x] Test both dev modes from scratch (as if a new user is running this branch)
- [x] Test both dev modes by first initializing from `main` and then switching to branch (to ensure we don't break)
- [x] Update README to explain the supported ENV variables for database config


config/database.yml
```diff
@@ -1,27 +1,32 @@
 default: &default
   adapter: postgresql
+  # url format: postgres://username:password@localhost:5432/database_name
+  url: <%= ENV["DATABASE_URL"] %>
+  # It's best to set DATABASE_URL to configure database details. That overrides the following settings
+  username: <%= ENV["HOSTEDGPT_DB_USERNAME"] || "hostedgpt_user" %>
```
jlvallelonga, line 6: defaults here for when running rails outside of docker

config/database.yml
```diff
@@ -1,27 +1,32 @@
...
+  password: <%= ENV["HOSTEDGPT_DB_PASSWORD"] %>
+  host: <%= ENV["HOSTEDGPT_DB_HOST"] %>
+  port: <%= ENV["HOSTEDGPT_DB_PORT"] %>
+  database: <%= ENV["HOSTEDGPT_DB_NAME"] || "hostedgpt" %>
   encoding: unicode
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
-  port: <%= ENV['HOSTEDGPT_DATABASE_PORT'] || 5432 %>
```
jlvallelonga, line 5: changing the names of a lot of these env vars for consistency. This would be a breaking change for anyone with an existing setup. I can have it look at the old env vars too so that it wont break anyone's existing setup or we can make people update and provide instructions for how to do so.

docker-compose.yml
```diff
@@ -1,15 +1,18 @@
...
     image: postgres:16
+    container_name: hostedgpt_postgres
     restart: always
     environment:
-      POSTGRES_USER: app
-      POSTGRES_DB: app_development
-      POSTGRES_PASSWORD: secret
+      # NOTE: If you use DATABASE_URL, it needs to match these env vars
+      #  so you can either have it match the default values for these env vars
+      #  or you can set these env vars to match the user and password from the url
```
jlvallelonga, line 9: this is the issue with using a url and running with docker. If you end up changing the username or password then you have to change it in two places. Pretty minor issue, but still not ideal

docker-compose.yml
```diff
@@ -1,15 +1,18 @@
...
+      - POSTGRES_USER=${HOSTEDGPT_DB_USERNAME:-hostedgpt_user}
+      - POSTGRES_PASSWORD=${HOSTEDGPT_DB_PASSWORD:-secret}
     volumes:
       - hostedgpt_pgdata:/var/lib/postgresql/data
     healthcheck:
-      test: ["CMD", "pg_isready", "-U", "app", "-d", "app_development"]
```
jlvallelonga, line 12: doesn't seem necessary to have the healthcheck try to connect to the database specifically. Just the database server seems like enough. This way we don't have to put the database name here and have another place to update it

docker-compose.yml
```diff
@@ -1,15 +1,18 @@
 services:
   postgres:
     image: postgres:16
+    container_name: hostedgpt_postgres
     restart: always
     environment:
-      POSTGRES_USER: app
-      POSTGRES_DB: app_development
```
jlvallelonga, line 7: rails will create the database so we don't need to make the postgres init process do it

docker-compose.yml
```diff
@@ -22,7 +25,12 @@ services:
       target: development
     environment:
       # Be sure to add environment variables to config/options.yml
-      - DATABASE_URL=postgres://app:secret@postgres/app_development
+      - HOSTEDGPT_DB_USERNAME=${HOSTEDGPT_DB_USERNAME:-hostedgpt_user}
+      - HOSTEDGPT_DB_PASSWORD=${HOSTEDGPT_DB_PASSWORD:-secret}
+      - HOSTEDGPT_DB_HOST=${HOSTEDGPT_DB_HOST:-postgres}
+      - HOSTEDGPT_DB_PORT=${HOSTEDGPT_DB_PORT:-5432}
+      - HOSTEDGPT_DB_NAME=${HOSTEDGPT_DB_NAME:-hostedgpt}
+      - DATABASE_URL
```
jlvallelonga, line 33: if DATABASE_URL is set then database.yml will use that as the url option (which will override other settings for the db there), otherwise we use the env vars here (which have default values now)

config/database.yml
```diff
@@ -1,27 +1,32 @@
...
 <% if RUBY_PLATFORM =~ /darwin/ %>
   gssencmode: disable
 <% end %>
-  url: <%= ENV['DATABASE_URL'] %>
-  # url format: postgres://username:password@localhost:5432/database_name
 
 development:
   <<: *default
-  database: <%= ENV['HOSTEDGPT_DEV_DB'] || "hostedgpt_development" %>
```
jlvallelonga, line 14: removed this in favor of just using the default in development

config/database.yml
```diff
@@ -1,27 +1,53 @@
...
+
+  def deprecated(env, message)
+    puts_red "DEPRECATED: #{env} is deprecated, #{message}" if ENV[env]
+  end
+
+
+  deprecated("HOSTEDGPT_DATABASE_PORT", "use HOSTEDGPT_DB_PORT instead")
+  deprecated("HOSTEDGPT_DEV_DB", "use HOSTEDGPT_DB_NAME instead")
+  deprecated("HOSTEDGPT_TEST_DB", "provide HOSTEDGPT_DB_NAME and test database name is generated")
+  deprecated("HOSTEDGPT_DATABASE_PASSWORD", "use HOSTEDGPT_DB_PASSWORD instead")
```
jlvallelonga, line 19: I wish there was a helper file for this. I guess I could put this in lib and include it manually. Maybe this could be an initializer too, but I wanted to keep this close to where the env vars are used. Or if we remove other envs in the future then maybe it's better in an initializer

config/database.yml
```diff
@@ -1,27 +1,53 @@
...
 default: &default
   adapter: postgresql
+  # url format: postgres://username:password@localhost:5432/database_name
+  url: <%= ENV["DATABASE_URL"] %>
+  # It's best to set DATABASE_URL to configure database details. That overrides the following settings
+  username: <%= ENV["HOSTEDGPT_DB_USERNAME"] || "hostedgpt_user" %>
+  password: <%= first_of("HOSTEDGPT_DATABASE_PASSWORD", "HOSTEDGPT_DB_PASSWORD") %>
+  host: <%= ENV["HOSTEDGPT_DB_HOST"] %>
+  port: <%= first_of("HOSTEDGPT_DATABASE_PORT", "HOSTEDGPT_DB_PORT") || 5432 %>
+  database: <%= first_of("HOSTEDGPT_DEV_DB", "HOSTEDGPT_DB_NAME") || "hostedgpt" %>
```
jlvallelonga, line 31: looking for existing env vars first to avoid breaking existing setups. A message is now printed to the console if these deprecated envs are being used.

config/database.yml
```diff
@@ -1,27 +1,53 @@
...
 development:
   <<: *default
-  database: <%= ENV['HOSTEDGPT_DEV_DB'] || "hostedgpt_development" %>
-  # It's best to set DATABASE_URL to configure database details. That overrides this "database" name. Note: Docker sets DATABASE_URL inside docker-compose.yml
 
 test:
   <<: *default
-  database: <%= ENV['HOSTEDGPT_TEST_DB'] || "hostedgpt_test" %>
+  # regex to match groups in the url: /(before db name)(db name)(after db name)/
+  url: <%= "#{ENV['DATABASE_URL']}".gsub(/(.*\/\/[^\/]+\/)([^?]+)(\??.*)/, '\1\2_test\3') %> # add "_test" to the database name if it's specified
```
jlvallelonga, line 44: we could have them set the test database name manually or set a test database url, but the combinations there get complicated and feels like it adds a lot of work to the user just to allow testing. This way there are less env vars to deal with too

docker-compose.yml
```diff
@@ -22,7 +25,16 @@ services:
...
+      - HOSTEDGPT_DATABASE_PORT # DEPRECATED: use HOSTEDGPT_DB_PORT
+      - HOSTEDGPT_DEV_DB # DEPRECATED: use HOSTEDGPT_DB_NAME
+      - HOSTEDGPT_TEST_DB # DEPRECATED: provide HOSTEDGPT_DB_NAME and test database name is generated
+      - HOSTEDGPT_DATABASE_PASSWORD # DEPRECATED: use HOSTEDGPT_DB_PASSWORD
```
jlvallelonga, line 37: leaving these in and marking as deprecated because if they're provided then we still want to use them to avoid breaking existing hostedgpt deployments

jlvallelonga:
maybe it's extra complexity, but I think this adds some useful functionality and it feels valuable to have consistently named env vars

.github/workflows/rubyonrails.yml
```diff
@@ -27,7 +26,7 @@ jobs:
           --health-retries 5
     env:
       RAILS_ENV: test
-      DATABASE_URL: "postgres://rails:password@localhost:5432/hostedgpt_test"
+      DATABASE_URL: "postgres://rails:password@localhost:5432/hostedgpt"
```
krschacht, line 29: What motivated the change of database name? I would have thought that using `hostedgpt` for the test environment would collide with the other environments?

docker-compose.yml
krschacht: Ah, I definitely see that we left off the HOSTEDGPT_DB_* env names from this compose file and needed to add those. That’s a good catch.

But why not continue to provide the defaults through the DATABASE_URL one? I think it effectively works the same, but I leaned slightly towards using DATABASE_URL because if we do that we’re subtly teaching people that this is what they should change locally if they want to configure things differently. Whereas by moving the defaults to all the individual env variables, we’re suggesting to people that they should set these variables locally if they want to change things.

Also, if I was running docker locally successfully yesterday, and then I pull down main and restart docker, will it blow up because the database name and username just changed in this commit? I like the default values yo used better than what was there, but I’m just a little worried that changing this will break it for active users.

krschacht: Oh wait, I just read in another comment that if we use DATABASE_URL and want to change the username and password, then we end up needing to change it in two places? I thought DATABASE_URL became the default so we just had to change it in there, but I may not fully understand how these configurations are interacting.

docker-compose.yml
krschacht: Oops, I asked about this in a different thread so I’ll resolve this one

.github/workflows/rubyonrails.yml
krschacht: Ohhh, I see your regexing within database.yml. Got it.

docker-compose.yml
krschacht: Okay, maybe what you’re doing is that if someone was using HOSTEDGPT_DEV_DB then if we switch our config to use DATABASE_URL it would override their configuration. So you changed docker config to use the individual pieces so as to respect people who have configured their local app using those.

.github/workflows/rubyonrails.yml
```diff
@@ -17,7 +17,6 @@ jobs:
         ports:
           - "5432:5432"
         env:
-          POSTGRES_DB: hostedgpt_test
           POSTGRES_USER: rails
           POSTGRES_PASSWORD: password
```
krschacht, line 21: In the docker compose, you removed the db, user, and password from the postgres container since it all comes from rails. That feels like a good change. But here within this github CI you only removed the db but kept user and password. Can’t we remove those too since they will also come from the rails container?

jlvallelonga: oh no in the docker compose the user and password are still there. Rails needs a user to connect to the database server in order to create the database so we have to at least set password. https://hub.docker.com/_/postgres the "Environment Variables" section explains these for the postgres image

docker-compose.yml
jlvallelonga: yeah the database default name change will break the existing deployments until the new env vars are set up for them. We can go back to the old values if that's a concern. I was thinking it would be less confusing for someone with a local postgres setup to see the "hostedgpt_..." named values

jlvallelonga:
just to communicate this since I kind of scattered the explanation around a bunch of comments. The DATABASE_URL works great for anyone not using the docker compose setup. However, if you are using docker compose, if you set DATABASE_URL then you also have to either set HOSTEDGPT_DB_USERNAME and HOSTEDGPT_DB_PASSWORD or you have to make sure the username and password in your DATABASE_URL match the defaults set in the docker compose:

```
      - POSTGRES_USER=${HOSTEDGPT_DB_USERNAME:-hostedgpt_user}
      - POSTGRES_PASSWORD=${HOSTEDGPT_DB_PASSWORD:-secret}
```

So since that's inconvenient for the docker users, I added the ability to just set the different database parameters as different settings.

There was also an issue with the tests using the development database when we just had the `url` setting in place so I added the regex to just set that `database` automatically. This way you only need to set either the DATABASE_URL or the separate db parameters and you don't have to set any specific test url or test parameters.

Honestly having the DATABASE_URL seems like it adds a lot of complexity that could be solved by just having the separate database parameters, BUT rails automatically looks for a DATABASE_URL env var and will use that as the `url` setting so there's still an opportunity for confusion there when running locally in the case that someone has that env var already set for some other project. So it seems better to follow the recommendation and explicitly use DATABASE_URL as the `url` so that it's more obvious that that's what is being used.

Lastly I updated the database parameters with a "hostedgpt_" prefix so that anyone running outside of docker can know that this database is for hostedgpt in the case that they have multiple projects using the same postgres server.

krschacht:
Thanks for that context! On this key point:

> if you are using docker compose, if you set DATABASE_URL then you also have to either set HOSTEDGPT_DB_USERNAME and HOSTEDGPT_DB_PASSWORD or you have to make sure the username and password in your DATABASE_URL match the defaults set in docker-compose.yml:

Is this still true, as in this is something inherent in the docker setup? Or is this the issue that you’ve now fixed so we can just use DATABASE_URL even for Docker?

If it’s still true, do you know why this is an issue with Docker? I’m guessing it’s because **Rails** reads DATABASE_URL and uses it to override any other settings, but Docker does not? So Docker reads the explicit username & password?

Is there a solution in here where we configure the docker postgres container to work with no username and password, so then DATABASE_URL can be all you need to do for docker usage? Or do we just need to accept that DATABASE_URL isn’t really supported by our Docker setup and we’ll just direct users to explicitly set the username and password?

jlvallelonga:
> Thanks for that context! On this key point:
> 
> > if you are using docker compose, if you set DATABASE_URL then you also have to either set HOSTEDGPT_DB_USERNAME and HOSTEDGPT_DB_PASSWORD or you have to make sure the username and password in your DATABASE_URL match the defaults set in docker-compose.yml:
> 
> Is this still true, as in this is something inherent in the docker setup? Or is this the issue that you’ve now fixed so we can just use DATABASE_URL even for Docker?
> 
> If it’s still true, do you know why this is an issue with Docker? I’m guessing it’s because **Rails** reads DATABASE_URL and uses it to override any other settings, but Docker does not? So Docker reads the explicit username & password?
> 
> Is there a solution in here where we configure the docker postgres container to work with no username and password, so then DATABASE_URL can be all you need to do for docker usage? Or do we just need to accept that DATABASE_URL isn’t really supported by our Docker setup and we’ll just direct users to explicitly set the username and password?

No problem! Yeah that's still true. It's because the postgres image that we use in the docker compose will use the POSTGRES_USER and POSTGRES_PASSWORD values to set up the default user with a password. If we use the DATABASE_URL then I don't know of a way to make the postgres image automatically pick up the username and password from that env var. I guess we could have some intermediate image that uses a script to extract and set those values for the postgres image. We'd have to make a custom dockerfile for that though I think. Idk maybe there's an easier way

jlvallelonga:
ok I played with the idea of modifying the postgres image to create the rails user. Turns out there's a way to hook into the init process of the postgres server and run shell scripts. I added steps to create an init script to extract out the user and password from the DATABASE_URL so that it can be created before rails starts. So no more need to make the username and password match the DATABASE_URL and docker users can just use that env var if they want

Dockerfile
```diff
@@ -86,6 +86,34 @@ EXPOSE 3000
...
+RUN echo "fi" >> 01-create-rails-user.sh
+
+RUN echo "echo \"creating user \$HOSTEDGPT_DB_USERNAME\"" >> 01-create-rails-user.sh
+RUN echo "psql -v ON_ERROR_STOP=1 --username \"\$POSTGRES_USER\" <<-EOSQL" >> 01-create-rails-user.sh
+RUN echo "	CREATE USER \$HOSTEDGPT_DB_USERNAME WITH SUPERUSER PASSWORD '\$HOSTEDGPT_DB_PASSWORD';" >> 01-create-rails-user.sh
+RUN echo "EOSQL" >> 01-create-rails-user.sh
+
+WORKDIR /
+
+#### END of POSTGRES ####
```
jlvallelonga, line 115: this section can go in a Dockerfile.postgres file if we want to split it out. I can probably combine those echo statements too or copy this script from a file in the repo (that's probably best)

docker-compose.yml
```diff
@@ -1,15 +1,20 @@
...
     restart: always
     environment:
-      POSTGRES_USER: app
-      POSTGRES_DB: app_development
-      POSTGRES_PASSWORD: secret
+      - POSTGRES_PASSWORD=${POSTGRES_PASSWORD:-secret}
+      # NOTE: If DATABASE_URL is set it will override values set in HOSTEDGPT_DB_USERNAME and HOSTEDGPT_DB_PASSWORD
+      - DATABASE_URL
+      - HOSTEDGPT_DB_USERNAME=${HOSTEDGPT_DB_USERNAME:-hostedgpt_user}
+      - HOSTEDGPT_DB_PASSWORD=${HOSTEDGPT_DB_PASSWORD:-secret}
```
jlvallelonga, line 13: these are now the values for the rails user instead of the default postgres user

krschacht:
Sorry I was a little slow in reviewing your last comments, but I just saw you made more updates to this. I’ll be sure to finish going through this today.

.github/workflows/rubyonrails.yml
krschacht: What about now that you removed the USER and PASSWORD from the dockerfile, do we remove them from here?

docker-compose.yml
```diff
@@ -22,7 +29,16 @@ services:
       target: development
     environment:
       # Be sure to add environment variables to config/options.yml
-      - DATABASE_URL=postgres://app:secret@postgres/app_development
+      - HOSTEDGPT_DB_USERNAME=${HOSTEDGPT_DB_USERNAME:-app}
+      - HOSTEDGPT_DB_PASSWORD=${HOSTEDGPT_DB_PASSWORD:-secret}
+      - HOSTEDGPT_DB_HOST=${HOSTEDGPT_DB_HOST:-postgres}
+      - HOSTEDGPT_DB_PORT=${HOSTEDGPT_DB_PORT:-5432}
+      - HOSTEDGPT_DB_NAME=${HOSTEDGPT_DB_NAME:-app_development}
+      - DATABASE_URL
```
krschacht, line 37: When I look at this docker-compose, it still looks like we're setting things twice. the username "app" and password "secret" are being set on HOSTED_DB_USERNAME and also POSTGRES_USER, but isn't that what you were eliminating?

I know this whole change ended up being a bigger thing than you intended. Maintaining support for multiple docker configs & the non-docker config is really quite complicated.

config/database.yml
```diff
@@ -1,27 +1,52 @@
...
+  deprecated("HOSTEDGPT_DATABASE_PASSWORD", "use HOSTEDGPT_DB_PASSWORD instead")
+%>
+
 default: &default
   adapter: postgresql
+  # url format: postgres://username:password@localhost:5432/database_name
+  url: <%= ENV["DATABASE_URL"] %>
+  # It's best to set DATABASE_URL to configure database details. That overrides the following settings
+  username: <%= ENV["HOSTEDGPT_DB_USERNAME"] || "app" %>
+  password: <%= first_of("HOSTEDGPT_DATABASE_PASSWORD", "HOSTEDGPT_DB_PASSWORD") %>
```
krschacht, line 27: Is first_of any different than `ENV[..] || ENV[..]`? It's fine to leave it in, I just couldn't tell if there was an edge case you were running into with that approach which first_of was solving for

docker-compose.yml
jlvallelonga: yeah support for the database url really added a lot. The postgres user is created in the init before our new init script is run. So we have to have that set as well and I'm not sure there's a way to take it from the database_url since the built-in init script runs before we can run any custom code

config/database.yml
jlvallelonga: yeah it should be the same. I can remove it to avoid the extra code. I thought it would be more helpful than it ended up being

docker-compose.yml
```diff
@@ -1,15 +1,22 @@
...
+      - POSTGRES_DB=${POSTGRES_DB:-app_development}
+      # NOTE: If DATABASE_URL is set it will override values set in HOSTEDGPT_DB_USERNAME and HOSTEDGPT_DB_PASSWORD
+      - DATABASE_URL
+      - HOSTEDGPT_DB_USERNAME=${HOSTEDGPT_DB_USERNAME:-app}
+      - HOSTEDGPT_DB_PASSWORD=${HOSTEDGPT_DB_PASSWORD:-secret}
     volumes:
       - hostedgpt_pgdata:/var/lib/postgresql/data
     healthcheck:
-      test: ["CMD", "pg_isready", "-U", "app", "-d", "app_development"]
+      test: pg_isready -U $$POSTGRES_USER -d $$POSTGRES_DB
```
jlvallelonga, line 19: had to go back to using postgres_db because if the database already exists then the init process doesn't run and pg_isready was trying to then connect to the "app" database (matching the "app" user) which doesn't exist if you have an existing deployment. Existing deployments will have a db named app_development.

.github/workflows/rubyonrails.yml
jlvallelonga: Maybe I made changes after you made this comment. I'm not sure where I removed those. Do you mean the docker-compose.yml file?

jlvallelonga:
ok I tested every way I could think of and it's working. Had to make some other minor changes for docker to allow the env vars to be used, but it's pretty simple now. Hopefully this will serve us well until we can remove the deprecated env vars

PR #549, opened 2024-11-15
replaces puts with logger now that logger works

jlvallelonga:
now that the logger is working for the workers, this replaces usage of "puts" with "Rails.logger.info"


krschacht (approved):
👍 on the cleanup!

PR #550, opened 2024-11-16
truncate long names so nav-container does not grow

voodoo:
1) Long names make the nav-container grow
2) Truncates the name to 20 chars
3) Rubocop worked :) 373 files inspected, 1 offense detected, 1 offense autocorrectable
4) Hopefully, no glass chewing

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -3,7 +3,6 @@ module ApplicationHelper
   # If the name on the nav-container is long truncate it to 20 characters
   # So it does not grow the left menu
   def truncate_long_name(name)
-    return name if name.nil? || name.length <= 20
     truncate(name, length: 20)
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,7 +1,5 @@
 module ApplicationHelper
 
-  # If the name on the nav-container is long truncate it to 20 characters
-  # So it does not grow the left menu
   def truncate_long_name(name)
     truncate(name, length: 20)
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,5 +1,12 @@
 module ApplicationHelper
 
+  # If the name on the nav-container is long truncate it to 20 characters
+  # So it does not grow the left menu
+  def truncate_long_name(name)
+    return name if name.nil? || name.length <= 20
+    truncate(name, length: 20)
+  end
+
   def at_most_two_initials(initials)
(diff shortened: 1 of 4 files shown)


krschacht (approved):
This looks great! I’ll merge it in right now.

A couple FYI comments:

When teeing up PRs that make a visual change, it’s helpful to have a before and after screenshot on the PR. In this case we talked about it on another thread and I know what it looks like so I’m good without it. But that’s a good rule of thumb.

And second, there is a whole other way to solve this problem which is in CSS rather than Ruby. Generally, I think that’s slightly preferred because there is a certain elegance to this simply being a “display problem” rather than a “code problem” and, in theory, it’s faster for the user because it’s pushing processing onto the browser engine (super efficient and each user has a whole one running) rather than the server engine (shared by all users and adds overhead to how long it takes the user to receive the new web page). If you’re curious about the CSS solution you can inspect the CSS around the conversation’s title in the same sidebar. I believe those use the CSS truncate option.

All that said, for this situation, this solution is perfectly fine and we don’t need to re-do it. Just sharing for the future.

krschacht:
Oh, and I just removed your comment on the helper method. When you make code self-descriptive than you don’t need comments to describe it. That’s always best. When you start adding comments, it’s actually just another thing to maintain. Someone may change 20 to 30 and then forget to update the comment which also includes 20.

voodoo:
- truncate in Tailwind did not seem to work - you can't limit the chars.
- Will include the screenshots next time.

This is a great project, with a great framework working on something relevant and useful and emerging (AI). 

Awesome, thanks again for your patience and passing on your experience - perhaps I will tackle something non-trivial next.

krschacht:
You’re welcome! I appreciate you diving in so it’s a win-win. :) CSS truncate relies on width rather than characters. If we rely on characters then it’s somewhat depend on font size. For example, a user may Cmd + on their keyboard to zoom in the text size for accessibility reasons and then the same number of characters might overflow a container where it previously wasn’t.

Quick searched pulled up this, as an example:
https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

But I vaguely remember there are other options besides ellipsis. I think even just the combination of “width: xxx; overflow: hidden” could be viewed as a form of truncation which prevents the container from expanding too wide.

voodoo:
Well look at that! That's a much better fix and gets rid of server code.  That's just much better man. 

One tiny w-20 and take out the ruby and tests. Yea.

Who knew you could delete a PR - you did!

krschacht:
I already merged this PR in but if you’re motivated you can tee up a new PR that does it the other way :)

you don’t have to, I’m good either way

PR #552, opened 2024-11-17
This commit just takes out the old truncate code on server/tests.

voodoo:
I'll push the change of w-20 on a new pull


voodoo:
Tests fail for the w-20? The universe is conspiring against me.

PR #553, opened 2024-11-17
Fix truncate with css. Take out truncate from server/tests.

voodoo:
Tests are passing locally. Same with rubocop.


voodoo:
This is simply removing a former PR. I took out the w-20 change to isolate it (the w-20 is on another branch) and iterate.

krschacht:
I'll close out in favor of #554

PR #554, opened 2024-11-17
Truncate profile name using CSS instead of rails helper

voodoo:
This was generated with a system test. I'm not seeing how we can tests the value in that field. It does not see the ... from truncate.

failures_test_should_truncate_the_name

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,9 +1,5 @@
 module ApplicationHelper
 
-  def truncate_long_name(name)
-    truncate(name, length: 20)
-  end
-
   def at_most_two_initials(initials)
     return initials if initials.nil? || initials.length <= 2
     initials[0] + initials[-1]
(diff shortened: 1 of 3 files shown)


voodoo:
password_resets_controller_test.rb failed? I don't get it.

I dont 'think' we will be able to test truncate.

https://stackoverflow.com/questions/51630998/capybara-and-text-overflow-ellipsis

> Capybara depends on the text being returned by the drivers. The drivers are trying to meet the WebDriver spec definition for what text to return, which is currently defined at https://w3c.github.io/webdriver/#get-element-text . That spec states that it is intended to return the text "as rendered", however the bot.dom.getVisibleText given as the standard doesn't currently take text-overflow into account. This means that as of now there's not much Capybara can do about it, and there really isn't a way to directly test it.


https://stackoverflow.com/questions/18090678/is-it-possible-to-call-truncate-method-in-rspec-feature-test

https://medium.com/@dave_go/testing-text-truncation-effective-strategies-2a589841a22b

krschacht:
@voodoo I saw you tee'd up a separate PR simply to remove the old approach, but this whole PR is small enough I think it's fine to combine them. Also, I agree that this aspect of display cannot be easily tested with automated testing. I think it's fine that it's not covered by our test suite.

I saw a failed system test, but that appears to be a flappy test. It's really hard to make system tests 100% reliable so I simply clicked "re-run failed test" within github CI and it passed the second time. All looks good on this PR so I'll merge in!

PR #555, opened 2024-11-17
MicrosoftGraphCredential#token is always a refreshed token

drnic:
This approach to requesting an access `token` and ensuring it is always refreshed is different from your Google code. I've left the code in the `MicrosoftGraphCredential` for now; but it can be moved into `Credential` or a concern if you wanted other subclasses to use it too, in a future PR.


krschacht (approved):
:+1 change looks good, but it looks like there is some new unreliability in our system tests. I've spent so many cycles on it thought it's been solved for months. Anyway, that was the only reason for the rest failure.

PR #556, opened 2024-11-17
Import LanguageModels from models.yaml

drnic:
* [x] Update app/models/user/registerable.rb to import models
* [x] Clean out the migrations of explicit data creation
* [x] Add some newer models for testing of `rails db:prepare` upgrade path


drnic:
@krschacht this PR replaces #536

test/models/language_model_test.rb
```diff
@@ -81,6 +81,72 @@ class LanguageModelTest < ActiveSupport::TestCase
     assert list.include?("alpaca:medium")
   end
 
+  # Functionality in LanguageModel::Export concern
```
krschacht, line 84: These tests look good, but I've been putting tests for a concern into the corresponding test file so that would be: `test/models/language_model/export_test.rb`. Then you don't need this comment

app/models/concerns/language_model/export.rb
```diff
@@ -0,0 +1,55 @@
+module LanguageModel::Export
+  extend ActiveSupport::Concern
+
+  def as_json(options = {})
+    options = options.with_indifferent_access
+    attrs = super(options)
+    attrs["api_service_name"] = api_service_name if options[:only].include?(:api_service_name)
+    attrs
+  end
```
krschacht, line 9: I got an exception in rails console when I did `LanguageModel.last.as_json` because I left off only. We could just add `&` before, but it technically won't work if we do `...as_json(only: :api_service_name)`

We don't necessarily need to support all those different edge cases, but out of curiosity I was asking chatgpt if there was a more idiomatic way to support delegate attributes with as_json since this is a common pattern. I don't think there is, but ChatGPT wrote this code as a suggestion (I didn't test this):

```
  def as_json(options = {})
    super(options).tap do |hash|
      hash['api_service_name'] = api_service_name if include_api_service_name?(options)
    end
  end

  private

  def include_api_service_name?(options)
    options = options.with_indifferent_access
    return true unless options[:only] || options[:except]

    only = Array(options[:only])
    except = Array(options[:except])

    (only.empty? || only.include?('api_service_name')) && !except.include?('api_service_name')
  end
```

app/models/language_model.rb
```diff
@@ -1,5 +1,7 @@
 # We don"t care about large or not
 class LanguageModel < ApplicationRecord
+  include LanguageModel::Export
```
krschacht, line 3: When you move the concern file, this can be updated to `include Export`

models.yaml
```diff
@@ -0,0 +1,254 @@
+---
+models:
+- api_name: gpt-4o
+  name: GPT-4o (latest)
+  supports_images: true
+  supports_tools: true
+  input_token_cost_cents: '0.00025'
+  output_token_cost_cents: '0.001'
```
krschacht, line 8: Just like with strings, you don't need to put the floats inside quotes. But it's also okay to leave them.

models.yaml
```diff
@@ -0,0 +1,254 @@
+---
```
krschacht, line 1: rails convention seems to be `.yml` rather than `.yaml`

krschacht (changes requested):
I pulled it locally, ran the import script, and all works. I did a little poking around console and reviewed the code closely, just a few small comments.

The one thing I'm still a bit unsure of is the auto-updating. With the old approach of adding a migration for each new model is annoying, but one bonus is that everyone who forks off hostedgpt gets their branch auto-updated automatically each night (provided they haven't diverged from main). When their fork's main updates with the new migration, heroku and render automatically re-deploy (I'm not sure about Fly). This means that a lot of people have been noticing new models & assistants automatically appearing within their apps without doing anything.

It's not critical that we continue to support that, but is there an easy way? All 3 environments are configured to have a release step which runs `bin/rails db:prepare` so can we connect the `models:import` so it auto-runs after db:prepare is run?

README.md
```diff
@@ -368,6 +368,46 @@ HostedGPT requires these services to be running:
...
+If you want to export the models in the local DB to another file, either `.json` or `.yaml`, pass in an argument:
+
+```plain
+rails models:export[tmp/models.json]
+```
+
+To import from another file, similarly provide the path as an argument:
+
+```plain
+rails models:import[tmp/models.json]
```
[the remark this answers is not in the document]
drnic: So, good news - JSON is a subset of YAML, so `YAML.load_file("test.json")` "just works".

app/models/concerns/language_model/export.rb
```diff
@@ -0,0 +1,55 @@
+module LanguageModel::Export
```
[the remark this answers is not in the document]
drnic: Done

app/models/concerns/language_model/export.rb
drnic: Ooh, a little research (surely this is easier than this?) and it's now just:

```ruby
  def as_json(options = {})
    super(options).merge("api_service_name" => api_service_name)
  end
```

drnic: More research (because, why is this so complicated?) and discovered I can override `attributes` method to add new custom attributes.

```ruby
  def attributes
    super.merge("api_service_name" => api_service_name)
  end
```

However, for reasons I cannot debug, but I've moved on with my life, we also need to re-define this method. Dumb, but whatever.

```ruby
  # Unsure why this needs to re-defined, but the original ActiveModel::Serialization
  # implementation is ignoring the #attributes method above.
  def attribute_names_for_serialization
    attributes.keys
  end
```

models.yaml
drnic: Done

drnic:
Ok, `rails db:prepare` now also runs `rails models:import`.

I've also added some newer models into `models.yml` to allow testing of the upgrade from `main` to this PR.

README.md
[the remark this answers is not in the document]
krschacht: TIL :)

app/models/concerns/language_model/export.rb
krschacht: Very nice, that's really simple.

PR #557, opened 2024-11-18
Add 'forget' on individual memory

mattlindsey:
Add's trash icon like on ChatGPT.

### DIFF
--- a/app/views/settings/memories/_memory.html.erb
+++ b/app/views/settings/memories/_memory.html.erb
@@ -1,7 +1,7 @@
 <tr class="py-1 <%= cycle('bg-gray-100 dark:bg-gray-600', 'bg-white dark:bg-gray-900') %>">
   <td>
     <%= memory.detail %>
-    <%= link_to_source(memory).html_safe %>
+    <%= link_to_source(memory).html_safe if memory.message.present? %>
   </td>
   <td>
     <%= button_to settings_memory_path(memory), method: :delete, class: "inline-block cursor-pointer py-1 pt-0 align-middle" do %>
--- a/app/views/settings/memories/_memory.html.erb
+++ b/app/views/settings/memories/_memory.html.erb
@@ -1,4 +1,20 @@
-<li class="list-disc py-1">
-  <%= memory.detail %>
-  <%= link_to_source(memory).html_safe %>
-</li>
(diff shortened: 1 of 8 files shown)


app/views/settings/memories/_memory.html.erb
```diff
@@ -1,4 +1,20 @@
-<li class="list-disc py-1">
-  <%= memory.detail %>
-  <%= link_to_source(memory).html_safe %>
-</li>
+<tr class="py-1 <%= cycle('bg-gray-100 dark:bg-gray-600', 'bg-white dark:bg-gray-900') %>">
+  <td>
+    <%= memory.detail %>
+    <%= link_to_source(memory).html_safe if memory.message.present? %>
```
krschacht, line 4: Looking at the model, I don't think the message can ever be blank but it doesn't hurt to be cautious

test/controllers/settings/memories_controller_test.rb
```diff
@@ -27,4 +27,14 @@ class Settings::MemoriesControllerTest < ActionDispatch::IntegrationTest
...
+
+  test "delete a single user's memory" do
+    login_as users(:keith)
+    assert_difference "Memory.count", -1 do
+      delete settings_memory_url(memories(:pet))
+    end
+
+    assert_redirected_to settings_memories_url
+    assert flash[:notice].present?
+  end
```
krschacht, line 39: I added a quick test so we have some coverage for the new code path

app/views/settings/memories/_memory.html.erb
mattlindsey: If you delete it in the UI I think the memory remains.

krschacht: Ohhh, hmm, it looks like message is missing a has_one (or many?) memories so there is no delete propagation. Ack.

BTW, I finally fixed the test that I added and kept breaking. :) This PR looks great, thanks so much for teeing it up! If you want to do another PR on the memories association that would be great, but no obligation at all. I'll take whatever you're feeling motivated to work on.

mattlindsey: I think it would be annoying if the delete propagated.  Everything works great now and ChatGPT doesn't delete memories automatically either. 
I might do the thing about showing the tool call in the UI so the user knows what's happening.  I was going to do 'web search' but that seems pretty hard!

PR #560, opened 2024-11-20
Default assistants from file + models from file

drnic:
This PR includes #536 

Instead of new Assistants being defined in code, they are defined in `assistants.yml`.

The PR defines the same Assistants as before, plus a new "GPT-4o Mini" assistant.

Assistants have a unique (per user) slug which is used in urls `/assistants/claude-sonnet/messages/new` and used to match updates from `assistants.yml` to existing Assistant.

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -73,7 +73,7 @@ class MessagesController < ApplicationController
   end
 
   def set_assistant
-    @assistant = Current.user.assistants_including_deleted.find_by(id: params[:assistant_id])
+    @assistant = Current.user.assistants_including_deleted.find_by(slug: params[:assistant_id])
     @assistant ||= @conversation.latest_message_for_version(@version).assistant
   end
 
--- a/app/models/assistant/export.rb
+++ b/app/models/assistant/export.rb
@@ -3,10 +3,9 @@ module Assistant::Export
 
   DEFAULT_EXPORT_ONLY = %i[
     name
+    slug
     description
     instructions
-    tools
-    external_id
     language_model_api_name
   ]
 
--- a/app/models/assistant/export.rb
+++ b/app/models/assistant/export.rb
@@ -42,8 +42,8 @@ module Assistant::Export
       assistants.each do |assistant|
         assistant = assistant.with_indifferent_access
         users.each do |user|
-          asst = user.assistants.find_or_create_by(name: assistant["name"])
-          asst.assign_attributes(assistant.except("name"))
+          asst = user.assistants.find_or_create_by(slug: assistant["slug"])
+          asst.assign_attributes(assistant.except("slug"))
           asst.save!
         end
       end
--- a/app/models/assistant/export.rb
+++ b/app/models/assistant/export.rb
@@ -1,0 +1,53 @@
+module Assistant::Export
+  extend ActiveSupport::Concern
+
+  DEFAULT_EXPORT_ONLY = %i[
+    name
+    description
+    instructions
+    tools
+    external_id
+    language_model_api_name
+  ]
+
+  DEFAULT_ASSISTANT_FILE = "assistants.yml"
+
+  def attributes
+    super.merge("language_model_api_name" => language_model_api_name)
+  end
+
+  # Unsure why this needs to re-defined, but the original ActiveModel::Serialization
+  # implementation is ignoring the #attributes method above.
+  def attribute_names_for_serialization
+    attributes.keys
+  end
+
+  class_methods do
+    def export_to_file(path: Rails.root.join(DEFAULT_ASSISTANT_FILE), assistants:, only: DEFAULT_EXPORT_ONLY)
+      path = path.to_s
+      storage = {
+        "assistants" => assistants.as_json(only:).map(&:compact)
+      }
+      if path.ends_with?(".json")
+        File.write(path, storage.to_json)
+      else
+        File.write(path, storage.to_yaml)
+      end
+    end
+
+    def import_from_file(path: Rails.root.join(DEFAULT_ASSISTANT_FILE), users: User.all)
+      users = Array.wrap(users)
+
+      storage = YAML.load_file(path)
+      assistants = storage["assistants"]
+      assistants.each do |assistant|
+        assistant = assistant.with_indifferent_access
+        users.each do |user|
+          asst = user.assistants.find_or_create_by(name: assistant["name"])
+          asst.assign_attributes(assistant.except("name"))
+          asst.save!
+        end
+      end
+    end
+  end
+end
--- a/test/controllers/password_resets_controller_test.rb
+++ b/test/controllers/password_resets_controller_test.rb
@@ -31,7 +31,10 @@ class PasswordResetsControllerTest < ActionDispatch::IntegrationTest
 
     email = @person.email
 
-      # set the user agent in the request headers
+    # Test can flap on CI if we don't clear the queue
+    clear_enqueued_jobs
+
+    # set the user agent in the request headers
     ActionDispatch::Request.stub_any_instance(:user_agent, "#{browser} on #{operating_system}") do
       post password_resets_url, params: { email: email }
     end
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,6 @@ gem "turbo-rails", "~> 2.0.5"
 gem "stimulus-rails", "~> 1.3.3"
 gem "tailwindcss-rails", "~> 2.7.2"
 gem "rack-cors"
-gem "ostruct"
 
 # Build JSON APIs with ease [https://github.com/rails/jbuilder]
 # gem "jbuilder"
--- a/app/controllers/settings/assistants_controller.rb
+++ b/app/controllers/settings/assistants_controller.rb
@@ -49,6 +49,6 @@ class Settings::AssistantsController < Settings::ApplicationController
   end
 
   def assistant_params
-    params.require(:assistant).permit(:name, :description, :instructions, :language_model_id)
+    params.require(:assistant).permit(:name, :slug, :description, :instructions, :language_model_id)
   end
 end
--- a/app/controllers/settings/assistants_controller.rb
+++ b/app/controllers/settings/assistants_controller.rb
@@ -38,7 +38,7 @@ class Settings::AssistantsController < Settings::ApplicationController
   private
 
   def set_assistant
-    @assistant = Current.user.assistants.find_by(id: params[:id])
+    @assistant = Current.user.assistants.find_by(slug: params[:id])
     if @assistant.nil?
       redirect_to new_settings_assistant_url, notice: "The assistant was deleted", status: :see_other
     end
--- a/app/models/assistant.rb
+++ b/app/models/assistant.rb
@@ -1,4 +1,6 @@
 class Assistant < ApplicationRecord
+  include Export
+
   MAX_LIST_DISPLAY = 5
 
   belongs_to :user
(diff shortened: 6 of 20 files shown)


krschacht:
@drnic I merged in your other PR and it created some merge conflicts. I took a quick look but I'm not 100% sure on the right way to resolve so I don't want to break it.

drnic:
Ok, rebuilt the PR off `main`.

Question: in `Assistant` what is `Assistant#tools` used for? I don't think its currently being used to select or filter the available Toolbox classes. If its unused, I can drop it from the `DEFAULT_EXPORT_ONLY` list until its used in future.

drnic:
Also, perhaps `Assistant#instructions` should not be exported to `assistants.yml` as I think this is where each user's personal "memory" is stored?

drnic:
And, are we using `Assistant#external_id` yet? What's its future use?

krschacht:
@drnic Ah, currently our Assistants class is a local-only model. But you know how I mentioned that OpenAI has this new API where they persist Assistants: at one point in time I was preparing to implement that API so I added the missing fields. Those fields are to match this schema:

https://platform.openai.com/docs/api-reference/assistants/createAssistant

It's probably worth dropping those columns since they're not actually implemented. I don't like keeping around half-implemented code and we can always add them back if ever deciding to implement the Assistant API.

krschacht:
It looks like this branch is behind main. I'll wait to review until you flag it for me.

drnic:
Ok ci green

krschacht:
Ohhh, sorry when I went to diff I saw `include Export` and other code that was so familiar from the other PR — which was merged into main — I thought it was mistakenly showing me parts of the diff that were incorrect. I see now this is all new code. :)

I just did a pass through the diff and it looks good. Are there any other changes you had planned for it or is it ready to be merged in, from your view?

drnic:
I think it's good as it is. There's always other ideas. But this is handy and obvious where assistants + their associated LLM models come from.

drnic:
I'd do a PR for dropping the unused fields later.

app/models/assistant/export.rb
```diff
@@ -0,0 +1,53 @@
...
+      assistants = storage["assistants"]
+      assistants.each do |assistant|
+        assistant = assistant.with_indifferent_access
+        users.each do |user|
+          asst = user.assistants.find_or_create_by(name: assistant["name"])
+          asst.assign_attributes(assistant.except("name"))
+          asst.save!
+        end
+      end
+    end
```
krschacht, line 51: I tested this locally and I'm thinking through a common workflow:

Say OpenAI comes out with a new variation of gpt-4o tomorrow called gpt-4.5o. We'll tee up a PR that updates the `models.yml` with this new model. It'll have a distinct `api_name` and we'd name it `GPT 4.5o (latest)` and `best: true`. Then we would update the existing gpt-4o model in our to change `best: false` and remove `(latest)` from it's name or maybe changing that parenthetical to the date of it's original date.

All of this seems doable since `language_models/export.rb` does find_or_initialize_by(api_name). Good call on that.

Then, I'd normally write the migrations to remove GPT-4o from everyone's assistant and instead add a new GPT-4.5o as their assistant. But this new assistants.yml workflow doesn't support that.

Should it? Maybe not. If we were to merge this in as-is, we'd (a) change the `GPT-4o` entry within it to `GPT-4.5o` and we could write a migration to remove the existing one for everyone.

What if we wanted to do a more minor edit such as changing `GPT-4o` to `GPT 4o` and changing `Meta Llama` to `Meta Llama 70b`. If we made those changes within assistants.yml then it would create new duplicate ones.

**I think we should consider this update logic to (1) grab the language_model record based on `language_model_api_name` and then (2) for each user do a `find_or_initialize_by(language_model_id: )`. That would permit us to rename assistants for people.**

So that would enable our YML files to be useful for adding models & assistants, changing details of models & assistants, but not removing either. We use migrations to remove things. That feels right. Do you agree?

krschacht (changes requested):
Left one comment inline

app/models/assistant/export.rb
drnic: > I think we should consider this update logic to (1) grab the language_model record based on language_model_api_name and then (2) for each user do a find_or_initialize_by(language_model_id: ). That would permit us to rename assistants for people.

So I think my own uses of Assistants will have many that share the same LLM (e.g. `gpt-4o` latest) but have different system prompts. I think this might be called an Agent to some people.

Therefore neither find/update by `name` (original PR) nor `language_model_api_name` is appropriate.

Couple ideas:

1. We introduce a slug for each Assistant that is permanent, even if the name is changed. E.g. "GPT 4o" could have slug `gpt-4o`. Now, many Assistants in the system would have duplicate `gpt-4o` but we'd only allow one per user. We could then use `User.assistants.find_or_initialize_by(slug:)` for the update.
2. We never update, only create Assistants for new users.

krschacht: Oh, good point. I think what you’re doing with multiple assistants is important to support.

I like the idea of a slug as the way to resolve it. That’s a good idea. I think it’s worth us having a small set of default assistants for users which is basically the “best” assistant for each service, so the slug could be something like “openai-default” or something like that, which we can then continually update as these LLMs evolve.

drnic: Ok, PR now has `Assistant#slug` support; including migration that tries hard to map any existing Assistant's to the "best" assistants' slugs in assistants.yml; and generates unique slugs for all other `Assistant` based on `name`.

drnic:
The failing spec is same as on `main`

drnic:
Please double check the ideas in new migration -- is this a useful help to existing users' to generate slugs that match to the assistants.yml; or is it better to match based on the name only -- if a previous user didn't change the name, then they get the matching slug.

mattlindsey:
> Please double check the ideas in new migration 
Currently the migration leaves Assistants that I created to `slug: nil` so they disappear from the UI.  Is your idea to create a slug from the Name?
Or is exporting them a required step before running db:migrate?

drnic:
>  Currently the migration leaves Assistants that I created to `slug: nil` so they disappear from the UI.  Is your idea to create a slug from the Name?
> Or is exporting them a required step before running db:migrate?

In the PR it will add the new slug column and then generate a slug for every assistant based on its name. Except first it tries to find the original Assistants and give them the same slugs that are now in the assistants.yml file.

mattlindsey:
This might be a wierd case. The Language Model for my Assistant is based on an API Service that I had added, so that may be a case you need to handle.  Let me know if you need more info.

drnic:
> This might be a wierd case. The Language Model for my Assistant is based on an API Service that I had added, so that may be a case you need to handle. Let me know if you need more info.

Some people might have a simple 1:1 between an assistant and a language model api_name.

Others might have 10 assistants for gpt-4o, each with different system prompts, ready to behave differently from each other; but using the same language model.

drnic:
So what the migration is doing is finding the oldest assistant for each of the primary language models (gpt-4o, Claude, llama-3.1-70b) and assign them their assistants.yml slug. 

The rest get a generated slug from their name.

However, the “original” assistants will have their name overridden by assistants.yml. And that might not be what we want.

mattlindsey:
> Maybe assistants.yml should only update the language model/api name for the assistants; but not change the assistant name.

Cool.  I was just saying currently in this PR the slug for my assistant gets nil, but when your done with this PR it should get one based on the Name. Right?  I'm a little unclear about your issue regarding names, but hopefully people won't mind too much if a name gets changed.

drnic:
> Cool. I was just saying currently in this PR the slug for my assistant gets nil, 

Ah, that’s a bug. The migration should run save! on every assistant which will generate the slug. I’ll try to reproduce.

> (as long as you don't keep changing them back to a default when they want a different name).

Yeah that’s the issue - the PR currently will keep renaming the assistants each time you db:prepare; so that’s annoying. I’ll update that.

test/controllers/password_resets_controller_test.rb
```diff
@@ -31,7 +31,10 @@ class PasswordResetsControllerTest < ActionDispatch::IntegrationTest
 
     email = @person.email
 
-      # set the user agent in the request headers
+    # Test can flap on CI if we don't clear the queue
+    clear_enqueued_jobs
+
+    # set the user agent in the request headers
```
krschacht, line 37: thank you!! I saw this flappy test a couple times in dev but wasn't able to figure it out yet

mattlindsey:
> @mattlindsey what's the name of your assistant which got a nil slug?

I just checked again and the slug is not nil.  Sorry about that.  
There must be another reason it's not showing up in the UI.
Below is from the console.  Maybe if I send the model and api records too, you can figure out why.

```
#<Assistant:0x0000000127c56048
  id: 4,
  user_id: 1,
  name: "Ollama",
  description: "Meta llama7b on ollama",
  instructions: "",
  tools: [],
  created_at: "2024-11-25 10:29:16.222762000 -0600",
  updated_at: "2024-11-27 06:08:57.487161000 -0600",
  language_model_id: 34,
  deleted_at: "2024-11-26 10:00:19.987117000 -0600",
  external_id: nil,
  slug: "ollama">]

**Comment by mattlindsey** [comment]
```
 #<LanguageModel:0x0000000136d15b50
  id: 34,
  position: 34,
  api_name: "llama3.1:8b",
  name: "Meta Llama 3.1 8b",
  supports_images: false,
  created_at: "2024-11-25 10:34:39.007164000 -0600",
  updated_at: "2024-11-25 11:16:27.141688000 -0600",
  deleted_at: nil,
  user_id: 1,
  api_service_id: 4,
  supports_tools: false,
  input_token_cost_cents: 0.5e-5,
  output_token_cost_cents: 0.8e-5,
  best: false,
  supports_system_message: true>,
  
   #<APIService:0x0000000134a92f88
  id: 4,
  user_id: 1,
  name: "Ollama",
  driver: "openai",
  url: "http://localhost:11434/v1/",
  token: [FILTERED],
  deleted_at: nil,
  created_at: "2024-11-25 10:33:26.337918000 -0600",
  updated_at: "2024-11-25 11:07:46.365756000 -0600">]

mattlindsey:
@drnic Damn it.  It is there fine.  Sorry for the wild goose chase!

app/controllers/messages_controller.rb
```diff
@@ -73,7 +73,7 @@ def set_conversation
   end
 
   def set_assistant
-    @assistant = Current.user.assistants_including_deleted.find_by(id: params[:assistant_id])
+    @assistant = Current.user.assistants_including_deleted.find_by(slug: params[:assistant_id])
```
krschacht, line 76: I was a bit surprised that this worked because I did not see you change a corresponding link_to in order to generate a url that contains this slug. Is that some default rails helper? I tried asking gpt4 but it claimed not so I don't get how the URLs are now using slug rather than id.

krschacht:
@mattlindsey no problem at all, it's great that you were previewing the PR and trying it out.

@drnic I just finished reviewing and testing locally and I think it's good to go! I'm going to merge in now.

sorry about this flappy system test. I think Justin might take a look at it this week when he wraps up his PR.

app/controllers/messages_controller.rb
drnic: See `Assistant#to_param` which is magically used by rails routes

krschacht: Oh, clever! But now I can't figure out why a newly added assistant is not working? Based on that implementation I would expect it to recognize any slug. Let's move over to this thread: https://github.com/AllYourBot/hostedgpt/pull/525#issuecomment-2505145760

PR #562, opened 2024-11-21
Display memory and weather tool call

mattlindsey:
I'm hoping this is a good start, but may need work:
- Too much logic in the view?
- Needs test?

--------------

image

### DIFF
--- a/app/helpers/messages_helper.rb
+++ b/app/helpers/messages_helper.rb
@@ -29,11 +29,11 @@ module MessagesHelper
 
   def format_for_display(message, append_inside_tag: nil)
     if memory_updated?(message)
-      return link_to message.content_text,
+      return link_to JSON.parse(message.content_text)["message_to_user"],
         settings_memories_path,
         {data: { turbo_frame: "_top" }, class: "ml-1 text-gray-600 dark:text-gray-300 no-underline"}
-    elsif weather_fetched?(message)
-      return "Good summary: #{JSON.parse(message.content_text)["good_summary"]}"
+    elsif message_to_user_from_tool_call?(message)
+      return JSON.parse(message.content_text)["message_to_user"]
     else
       escaped_text = html_escape(message.content_text)
 
@@ -69,12 +69,14 @@ module MessagesHelper
     )
   end
 
-  def memory_updated?(message)
-    message.tool? && message.content_tool_calls.dig(:function, :name) == "memory_remember_detail_about_user"
+  def message_to_user_from_tool_call?(message)
+    JSON.parse(message.content_text)["message_to_user"].present? if message.content_text.present?
+  rescue JSON::ParserError
+    false
   end
 
-  def weather_fetched?(message)
-    message.tool? && message.content_tool_calls.dig(:function, :name) == "openmeteo_get_current_and_todays_weather"
+  def memory_updated?(message)
+    message.tool? && message.content_tool_calls.dig(:function, :name) == "memory_remember_detail_about_user"
   end
 
   private
--- a/app/helpers/messages_helper.rb
+++ b/app/helpers/messages_helper.rb
@@ -30,9 +30,9 @@ module MessagesHelper
     if memory_updated?(message)
       return link_to JSON.parse(message.content_text)["message_to_user"],
         settings_memories_path,
-        {data: { turbo_frame: "_top" }, class: "ml-1 text-gray-600 dark:text-gray-300 no-underline"}
+        { data: { turbo_frame: "_top" }, class: "text-gray-400 dark:!text-gray-500 font-normal no-underline" }
     elsif message_to_user_from_tool_call?(message)
-      return JSON.parse(message.content_text)["message_to_user"]
+      return content_tag(:span, JSON.parse(message.content_text)["message_to_user"], class: "text-gray-400 dark:!text-gray-500")
     else
       escaped_text = html_escape(message.content_text)
 
--- a/app/helpers/messages_helper.rb
+++ b/app/helpers/messages_helper.rb
@@ -32,7 +32,8 @@ module MessagesHelper
       return link_to message.content_text,
         settings_memories_path,
         {data: { turbo_frame: "_top" }, class: "ml-1 text-gray-600 dark:text-gray-300 no-underline"}
-
+    elsif weather_fetched?(message)
+      return "Good summary: #{JSON.parse(message.content_text)["good_summary"]}"
     else
       escaped_text = html_escape(message.content_text)
 
@@ -70,6 +71,10 @@ module MessagesHelper
 
   def memory_updated?(message)
     message.tool? && message.content_tool_calls.dig(:function, :name) == "memory_remember_detail_about_user"
+  end
+
+  def weather_fetched?(message)
+    message.tool? && message.content_tool_calls.dig(:function, :name) == "openmeteo_get_current_and_todays_weather"
   end
 
   private
--- a/app/helpers/messages_helper.rb
+++ b/app/helpers/messages_helper.rb
@@ -15,8 +15,7 @@ module MessagesHelper
     case message.role
       when "user" then "You"
       when "assistant" then message.assistant.name
-      when "tool" then "Tool"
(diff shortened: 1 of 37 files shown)


mattlindsey:
Hi @krschacht.  I think this PR looks good now, unless you think it needs a system test or other change.

krschacht:
Great, I'll give it a close look now and also test it locally!

app/views/messages/_message.html.erb
```diff
@@ -13,11 +13,31 @@ end %>
 <% scroll_down = scroll_down || initial_page_load %>
 <% thinking ||= last_message && message.content_text == "" && message.not_cancelled? %>
 <% done_thinking = !thinking && message.content_text.present? %>
-<% debug_tool_messages = false %>
```
krschacht, line 16: Let's keep this debug in place, I've found it quite helpful when working on new tools

app/views/messages/_message.html.erb
```diff
@@ -13,11 +13,31 @@ end %>
...
+    %>
+    <div class="ml-1">
+      <%= format_for_display(message.content_text) %>
+    </div>
+    <%= link_to "(manage memories)\n", settings_memories_path,
+      data: { turbo_frame: "_top" },
+      class: "ml-1 text-gray-600 dark:text-gray-300"
+    %>
+  </div>
+<% end %>
```
krschacht, line 36: Instead of creating a whole new div structure, let's lean into the existing message structure below. This will help to make future design changes easier and ensure things like "data-role" and "data-subrole" are all in place.

Also, I think it's a bit brittle to do string matching of the tool's response, but I see why you did that because there isn't an easy way to tell what this tool message response is from. That motivated me to just land PR to help you out:
https://github.com/AllYourBot/hostedgpt/pull/563

Here's how I think you can refactor this:

1) Turn "memory_updated?" into a message_helper method so you can use `memory_updated?(message)` wherever you need to within here. I believe it'll simply be this (and I should have done that with some of these extra lines within here):

```
def memory_created?(message) = message.tool? && message.content_tool_calls.dig(:function, :name) == "memory_remember_detail_about_user"
```

2) Update `render_avatar_helper` method to include an elsif for `message.tool?`. Can we make it a wrench-screwdriver? I just played with it locally and this looked nice and visually lined up (although I didn't check dark mode so account for that too):

```
      icon "wrench-screwdriver",
        variant: :micro,
        class: %|
          mt-[5px]
          ml-[2px]
          text-gray-400
        |
```

3) Update `format_for_display` helper to **not** have a "Tool" one since we just want to render this thing as a single line

4) Update the `format_for_display` helper. I think you'll need to pass message into it and then at the very top of it you can parse the JSON and turn it into a link (basically, let's make the whole thing clickable but without an underline and then delete the parenthetical after it). This looked nice in my quick test (again, I didn't look at dark mode):

```
    text = JSON.parse(text) rescue text

    text = link_to(text, settings_memories_path,
      data: { turbo_frame: "_top" },
      class: "text-gray-400 no-underline"
    ) if message
```

app/views/messages/_message.html.erb
```diff
@@ -13,11 +13,31 @@ end %>
...
 
 <div
   id="<%= dom_id message %>"
-  class="mb-5 ml-9 sm:ml-3 md:ml-8 group <%= message.only_tool_response? && !debug_tool_messages && "hidden" %>"
+  class="mb-5 ml-9 sm:ml-3 md:ml-8 group <%= message.only_tool_response? && "hidden" %>"
```
krschacht, line 40: reverting the debug and using your helper, this would become:

```suggestion
  class="mb-5 ml-9 sm:ml-3 md:ml-8 group <%= message.only_tool_response? && !memory_updated?(message) && !debug_tool_message && "hidden" %>"
```

krschacht (changes requested):
This is on the right track, but I'd like to use the existing HTML structure. I think some minor modifications to the helpers can make the existing structure work. I put some comments inline

krschacht:
Let's also add one system test for this. Within system/conversations/messages I've been creating a separate file for each "feautre", so probably create a new file like memory_response_test.rb and there is an existing fixture which already contains a conversation which made a call to memory. This should get you started:

```
 setup do
    @user = users(:keith)
    login_as @user
    @conversation = conversations(:weather)
  end

  test "ensure a visible link is displayed to user that the memory was updated" do
    visit_and_scroll_wait conversation_messages_path(@conversation)
    binding.pry
  end
end

```

it's probably just that single test. poke around at the other features iwthin this test folder and you'll see various helpers I use to parse the messages

mattlindsey:
Added display of weather tool call because I wanted to be sure we would handle tool calls in a consistent manner.
I had been thinking the memory tool should return JSON like `{ message_to_user: "memory updated" }` and any tool with a message could return a `message_to_user` too which would be automatically displayed.  But maybe that's overkill and would confuse the llm.

Also tried adding a system test but may need help!  I'm unsure how to get the tool_call in the conversation fixture.

mattlindsey:
I added an option to add an optional `message_to_user` in the JSON returned by a tool.  It seems to work well, so I hope it's OK.  If the tool returns it, it will be displayed next to a Tool/wrench icon.  Example usage:
```ruby
{
  message_to_user: "This super tool worked great.",
  tool_data: "Here is some info for the llm",
}
```

I'm still having problems with the tests and could use help, if possible.

app/helpers/messages_helper.rb
```diff
@@ -25,17 +27,25 @@ def format_for_speaking(text)
     ::MarkdownRenderer.render_for_speaking(text)
   end
 
-  def format_for_display(text, append_inside_tag: nil)
```
krschacht, line 28: a few lines above this is “from_name_for()” and I think we should remove the “when tool” line from that so that the tool calls are just a single line message rather than two lines

app/views/layouts/_tool_avatar.erb
```diff
@@ -0,0 +1,10 @@
+<%= 
+  icon "wrench-screwdriver",
+    variant: :micro,
+    size: 22,
+    class: %|
+      ml-1
+      mr-1
+      text-gray-400
```
krschacht, line 8: check dark mode to see if the color should be different there

krschacht:
I looked quickly at the code and it looks good. I don’t think you necessarily needed to create a whole new conversation fixture, but there is no harm in doing it and we can use that one for future function calling.

Regarding the failing system test, I don’t see an obvious cause to it. I’m not at a computer but the way I debug is:

1) edit “application system test” file and change headless_chrome to just chrome
2) execute that single test and now you’ll see a browser open where it actually clicks and does things. put a binding.pry on the line right before the failure
3) in binding you can execute specific lines of the test and watch it remote control the browser so this way I execute different parts of the failing test and I get to see both what it’s doing in the browser and what console is getting back

krschacht:
Yes, some of these tests have recently started being really flaky. Sorry about that.

It’s great to hear you got it working! I’ll try to give this a run through on my computer by end of day Wednesday. I might be able to do later today, but I’m traveling until Tuesday PM so it may not be until I’m back.

mattlindsey:
I was playing with testing today.  Is it worth adding `test/helpers/messages_helper_test.rb` with the following? (They do pass.)

```ruby
require "test_helper"

class MessagesHelperTest < ActionView::TestCase

  test "recognizes memory updated" do
    message = messages(:keep_memory_tool_result)
    assert memory_updated?(message)
  end

  test "recognizes message to user from tool call" do
    message = messages(:keep_memory_tool_result)
    assert message_to_user_from_tool_call?(message)
  end
end

mattlindsey:
@krschacht Great. You can throw those tests in too if you are tweaking the PR.

krschacht:
The main changes I made were (1) making the messages to user be a bit less visually prominent, and (2) removing the message_to_user from the actual response that's visible to the LLM just so it doesn't cause any surprises. I also went ahead and removed the open_meteo message to user since it was a bit redundant in practice, but it was useful for us testing things. Overall, this PR is a great addition thanks for tackling it!

PR #563, opened 2024-11-23
Store the function call with the function response

krschacht:
With OpenAI, when functions are called, there is a response for each function call. We store these as messages in the `conversation.messages` history. When we do this, one thing that is tricky about parsing the response is knowing which function this response is from.

OpenAI gives us back a `tool_call_id` so if you look at the preceding message in the conversation history you can use this to match it up and see which tool/function that was. But this is a little tricky within the context of a single message.

So, this PR improves matters by populating the `content_tool_calls` attribute on response messages.

To summarize:

* OpenAI gives us back a tool_calls JSON block which indicates what tools it wants to call. This is an array since it can be multiple calls.
* We call those tools and create! a new message for each response from the tool calls, and **we now populate the tool_calls attribute for each of these responses so that it's easy to see what call this is a response to.**

### DIFF
--- a/app/jobs/get_next_ai_message_job.rb
+++ b/app/jobs/get_next_ai_message_job.rb
@@ -105,7 +105,7 @@ class GetNextAIMessageJob < ApplicationJob
       else
         error_text = nil
         begin
-          error_text = e&.response&.dig(:body, "error", "message")
+          error_text = e&.response&.dig(:body, "error", "message") rescue e&.response&.dig(:body)
         end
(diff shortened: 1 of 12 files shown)


drnic:
Thoughts on why this PR is causing the system test failure on `main`?

```
bin/rails test test/system/conversations/messages/images_test.rb:88
```

krschacht:
@drnic It's actually not that PR, I did a bunch of git reverts and re-applying and I narrowed it down to a different PR that @jlvallelonga was working on #532. I'm going to start a discussion on that one.

PR #565, opened 2024-11-27
Register test requires password_authentication; test login pages

drnic:
I discovered these specs might fail when I had password auth disabled locally. Fleshed out more tests for `login_url` too.


krschacht:
Looks great, I'm merging now. I also spent time digging into that broken system test and hopefully have gotten to the root of the problem, but I'm asking Justin how best to fix it. Hopefully we'll have that addressed soon.

PR #566, opened 2024-11-27
Assistants redirect to settings if no token yet

drnic:


krschacht:
I'm not sure about this change. The lock doesn't signal to me mis-configured, and when you click the assistant in order to see why it's locked then you land on that page and it's not immediately obvious why you're staring at this settings page.

What about this... immediately after signing up, I think it's probably clearer for the user to land on the new conversation view since this is really familiar. It's orienting to take a second and take in this interface that you should recognize. Soon you try typing a message and then are informed of the issue with the API key. However, the instructions are not the easiest to follow. Their multiple steps and you can easily go to the Settings page and end up clicking **the assistant in the left sidebar** and be really confused since you don't see an API key field there.

How about we keep the user error message which is displayed, but we improve that error message so that it actually includes a direct link to the appropriate API Services page? The error message could show:

```
(You need to enter a valid API key for Anthropic to use Claude. Click to visit the **Settings > API Services > Anthropic** and enter your token. You'll also find instructions for how to get it.)
```

But that bold part should be a clickable hyperlink.

PR #567, opened 2024-11-29
Add google_search tool

mattlindsey:
This works, but I'm sure it needs work and am hoping for feedback.
 
Test it by asking something like: `What is the latest headline on the Boston Globe?`.  That worked!

I added a `get_body` method because I wasn't sure how the other code worked.  Does it look correct?


mattlindsey:
@krschacht So this works pretty well but I need to know if the new `get_body` method is ok.

And adding `webmock` for testing broke a couple of tests in `open_meteo_test.rb` and I was unable to fix them.  Maybe there was a way to test`google_search` without adding webmock?  Anyway, let me know what you think.

mattlindsey:
I fixed the tests in `open_meteo_test.rb` by disabling webmock for them.  I don't think the actual requests are being made.

app/services/toolbox/google_search.rb
```diff
@@ -0,0 +1,32 @@
...
+  describe :google_search, <<~S
+    Search Google for the indicated query.
+    Use this to answer questions about current events, look up information, or find answers to questions.
+    Try to use this sparingly; prefer to use the user's memories and the tools you have available to answer questions.
+    When you do use this, try to use exact queries for which you expect to get a definitive answer.
+    When you respond to the user, try to include an answer to the question rather than just a link.
+  S
+  def google_search(query_s:)
+    encoded_query = URI.encode_www_form_component(query_s)
+    response_body = get("https://www.google.com/search?q=#{encoded_query}").get_body
```
krschacht, line 12: Instead of get_body, the SDK library is expecting you to pass .params() or .no_params(). Those are the methods that return the body. So try:

```
get(“https://www.google.com/search”).params(q: query_s)
```

I believe that should handle the encoding for you as well and it should return the body. Notably, I wrote this generic SDK because I got tired of reading through API docs for all these different services, and reading through SDK docs wrapping each of the APIs. REST APIs are easy enough these days I just wanted a generic library that let me interact with them. I did not intend it for this specific use-case since google isn’t returning a JSON response, but I think it’ll probably work.

The `params` method ends by calling a helper method called `handle`. You can see it here:
https://github.com/AllYourBot/hostedgpt/blob/b8c9c52dd4916697cdf0332a9cd181946d629834/app/services/sdk/verb.rb#L30

It looks like I wrote it so that if the response is not JSON data then it simply returns the `response` object, so you may actually just want to call it like this:

```
get(“https://www.google.com/search”).params(q: query_s).body
```

krschacht:
I’m so excited that you’ve managed to get this working! I wasn’t sure if this v1 would play out in a useful way, but it sounds like you’ve figured out the parsing and hopefully there’s consistently enough info from scraping the google body to get useful stuff! I have some ideas about an easy way to make it decide which pages to visit, but that’s out of scope for a v1. We’ll get to live with this for awhile and see how often it’s able to glean useful results.

I reviewed the code and left one comment on the SDK usage.

> I fixed the tests in `open_meteo_test.rb` by disabling webmock for them. I don't think the actual requests are being made, are they?

It was a little unusual, but I actually left that live API call within the test. :) Maybe I shouldn’t, but I figured that if I mocked every API then there are some complex code paths deep within SDK that would never actually be hit so it was nice to exercise more of the code. Plus, it was only 2 API calls per CI run (I think) which didn’t seems crazy. And open metro is a completely open API, no keys are required, so it actually worked in CI and would work on anyone’s machine with no special configuration. I’m not set on leaving it, but that was my original rationale for not mocking that one tool.

app/services/toolbox/google_search.rb
mattlindsey: I originally tried calling it like that, and just tried again but it doesn't work and I think something about the `handle` method or the `OpenData` class has to change.  There's maybe something wrong or something I have to do to the `body` text before I can parse it.
It looks cool but I don't quite understand it yet.  I'll fool around with them a bit.

mattlindsey: The following works.  Does it look OK?  I'll fool around with it a little more.

```ruby
  def handle(response)
    raise ResponseError.new(response) if !response.status.in? @expected_statuses

    if response.status.between?(200, 299)
      begin
        json = JSON.parse(response.body)
      rescue JSON::ParserError, TypeError => e
        return response
      end
      response.body.presence && OpenData.for(json)
    else
      response
    end
  end
  ```

krschacht:
Hi @mattlindsey sorry for the delay on this, but I just returned from some travel and circled back. The code looks good and I tried it out and it worked great for me. I tried asking it an open ended question with no suggestion for it to look up anything (is moana 2 in theaters yet?) and it knew to do a search and got the right answer. And I'm really liking that "respond_to_user" option in the tool response. It's very slick.

So I think it's ready to merge. Was there anything more you were doing with it or can I go ahead?

krschacht (approved):

mattlindsey:
Great. That’s all I was going to do for now.

PR #569, opened 2024-11-30
Fix system tests

jlvallelonga:
fixes mocks so that selenium can run

fixes images_test system test


test/system/conversations/messages/images_test.rb
```diff
@@ -103,15 +103,14 @@ class ConversationMessagesImagesTest < ApplicationSystemTestCase
       refute img.visible?
 
       image_btn.click
-
       2.times do
         sleep 0.1
         sleep 0.5 if !modal_loader.visible?
         sleep 0.1
         image_btn.click if !modal_loader.visible?
       end # TODO: sometimes modal has not popped up after clicking, why?? Try 2x times before failing the test.
```
krschacht, line 111: @jlvallelonga I think we had 3 distinct test failures. You fixed one of them with the update to `simulate_image_variant_processing`. I tried to fix a sporadic `Net::ReadTimeout` with a timeout. However, there remains one more test failure on this line. I think it might be a legitimate bug.

When I turn off headless chrome and run just this specific test, I do not see the spinners in place of the images. I spent a little while investigating it and couldn't figure out why not. The thing I did not try was checking out an old commit right before the refactor to image attachments, and that would confirm for sure that we broke it recently.

Could you do a little more investigating of this one? What you should see when running `bin/rails test test/system/conversations/messages/images_test.rb:88` in normal chrome mode is the spinners appearing for 8 seconds both inline in the conversation messages and also when you click on a spinner you see a spinner within the modal that opens. Then after 8 seconds (of faked delay) then the spinners are swapped out for the images.

jlvallelonga: ok nice! yeah I'll dig into this last failure tomorrow

krschacht:
@jlvallelonga Let me know if you any time to look into this failure. Or if things are busy on the home front, let me know and I can try to dig into this.

jlvallelonga:
Yeah busy week at home. I'll take a look at this now

jlvallelonga:
@krschacht I'm seeing errors like this when running system tests locally and on the github runners:

```
Error:
ConversationMessagesImagesTest#test_ensure_images_display_a_spinner_initially_if_they_get_a_404_and_then_eventually_get_replaced_with_the_image:
WebMock::NetConnectNotAllowedError: Real HTTP connections are disabled. Unregistered request: POST http://127.0.0.1:9515/session with body '{"capabilities":{"alwaysMatch":{"browserName":"chrome","goog:chromeOptions":{"args":["--headless"],"binary":"/usr/bin/google-chrome"}}}}' with headers {'Accept'=>'application/json', 'Accept-Encoding'=>'gzip;q=1.0,deflate;q=0.6,identity;q=0.3', 'Content-Length'=>'136', 'Content-Type'=>'application/json; charset=UTF-8', 'User-Agent'=>'selenium/4.25.0 (ruby linux)'}
```

Are you doing something to get around that?

krschacht:
> @krschacht I'm seeing errors like this when running system tests locally and on the github runners:

Dang, I didn’t notice those. I get in this habit of once the system tests are broken I stop looking at them closely so I didn’t notice we had made them worse. I think it’s this change:
https://github.com/AllYourBot/hostedgpt/pull/567/files

Webmock was added along with some config for it. I don’t think we need webmock. I was able to mock up HTTP requests previously without using webmock so we should probably unwind that addition.

krschacht:
@jlvallelonga We can close this PR now that we merged in #586 right?

PR #570, opened 2024-11-30
reverts app_url changes to fix system tests

jlvallelonga:


jlvallelonga:
the system tests are actually failing for a different reason so closing this

PR #581, opened 2024-12-10
Fix conversation search and list to morph on refresh

mattlindsey:
I think this fixes problem with last PR that added searching, but I'm a little confused so better check it out.

### DIFF
--- a/app/javascript/stimulus/search_controller.js
+++ b/app/javascript/stimulus/search_controller.js
@@ -5,6 +5,10 @@ export default class extends Controller {
 
   inputTargetConnected() {
     this.setSearchClearIcon()
+    // Ensure the cursor is at the end of the input
+    this.inputTarget.selectionStart =
+      this.inputTarget.selectionEnd =
+        this.inputTarget.value.length
   }
 
   disconnect() {
--- a/app/views/conversations/index.html.erb
+++ b/app/views/conversations/index.html.erb
@@ -17,7 +17,9 @@
             w-full
             p-2 py-1
             border border-gray-200 rounded-lg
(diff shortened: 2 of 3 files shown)


mattlindsey:
I just rechecked this PR, and it does fix an extra refresh that was happening after a new conversation is auto-titled.

It's not too important, but feel free to reopen and apply this PR if desired.

krschacht:
I like the idea of getting this fixed so I’ll probably merge it in. I noticed yesterday “main” is broken for my personal instance of hostedgpt. It’s so easy for bugs to slip in and I feel bad when it breaks for other people, so I just need to get a block of time where I can try this and the other pending PRs out and be careful about merging in. But I think this PR is a good idea so I’ll re-open just so it stays there on the list and I don’t forget about it. Sorry I’m slow on some of these. This holiday time-of-year is not helping my already busy schedule. :)

mattlindsey:
This PR does fix the extra refresh, but now causes the cursor to jump to the beginning of the search input after every search, which is more annoying.  I changed this to 'draft' and will look into it.

mattlindsey:
Fixed cursor jumping, but now the X icon appears after every message.  I'll look for better fix. Sorry for all the noise.

buithehien1991:
image

I think we also have a small bug at dark mode. Search box still is white

buithehien1991:
@mattlindsey could you please fix it and commit for accept merge this pull?

krschacht:
@buithehien1991 sorry this PR was blocked on me! thanks for flagging this, I've just fixed the dark mode issue and am merging it in.

Thanks @mattlindsey !

PR #582, opened 2024-12-11
Set OpenAI ApiVersion to empty on initialization

papayalabs:
Set OpenAI ApiVersion to be empty during initialization; otherwise the Ruby-Openai gem will force v1 in the URL


mattlindsey:
Hi @papayalabs.  This worked fine for my setup with OpenAI, Anthropic, Ollama, and Groq.  
Although I had to manually add `v1/` to the OpenAI API Services setting.  I think you could safely add a migration to update any record that matches `https://api.openai.com/` exactly.

Issue #436, opened 2024-06-25
Add a "check connection" button for API Services


krschacht:

mattlindsey:
I started working on this and think it makes more sense to put the `Test` button on the page for a Language Model.

Or we could have it on the API Service page and get the `best` model (or first one if no best) and try that, if you want.

johndoe929:
Oh interesting, I see what you're saying. I forget who suggested this feature and exactly what context it came up, but looking at these pages it feels like these are both places where having a `Test` button could be helpful.

My suggestion would be to style it like the "How do I get?" text. So basically:

On the API Service editing screen, to the right of the "URL" text label would be blue underlined text that said "Test URL". This text would probably need to be greyed out or completely hidden until something had been entered into the text input since it doesn't make sense until text is entered, plus someone's attention will be drawn to it the moment it appears. Clicking this blue link could add additional default colored text afterwards which is simply "..." and then as soon as the test succeeds or fails the response appears afterwards.

On the Language Model screen this could be on the "API name" line.

So a completed test could look like this:

<img width="673" alt="image" src="https://github.com/user-attachments/assets/8c608f70-4c40-4636-9173-4a64c073cca6">

krschacht:
Oh interesting, I see what you're saying. I forget who suggested this feature and exactly what context it came up, but looking at these pages it feels like these are both places where having a Test button could be helpful.

My suggestion would be to style it like the "How do I get?" text. So basically:

On the API Service editing screen, to the right of the "URL" text label would be blue underlined text that said "Test URL". This text would probably need to be greyed out or completely hidden until something had been entered into the text input since it doesn't make sense until text is entered, plus someone's attention will be drawn to it the moment it appears. Clicking this blue link could add additional default colored text afterwards which is simply "..." and then as soon as the test succeeds or fails the response appears afterwards.

On the Language Model screen this could be on the "API name" line.

So a completed test could look like this:

<img width="703" alt="image" src="https://github.com/user-attachments/assets/2d2f8cf3-c9d8-4f55-9be5-09b57912725e">

PR #584, opened 2024-12-11
Add test on language model settings page

mattlindsey:
Adds test link and green message on success, red on error.
If this looks ok, I can also add a test link on the API Service settings page as shown here: #436

I tested this PR on my setup with Anthropic, OpenAI, Gemini, Groq, and Ollama.

I think I need to add tests for the new `test` controller action.  And probably the model and service too.


krschacht (changes requested):
The code for this looks great. I tested this out locally and made some small tweaks the aesthetic of it, but functionally it's working well. Two questions:

1) When I removed the API key for a service (pretending like I left it off) and then I clicked "test" I got an exception. Can we fail more gracefully so the user sees a nice red error message next to the "Test" link?
2) Should we also add a Test link when editing the API Services or is that tricky because we don't know a model name?

krschacht:
Nice improvements! I'm merging this in now.

PR #613, opened 2025-02-03
Update Groq Models

lumpidu:
The `llama-3.1-70b-versatile` model has been removed from the Groq API. With deepseek-r1's rising popularity, testing distilled versions on Groq's infrastructure could be worthwhile.

- add deepseek-r1-distill-llama-70B
- replace llama-3.1-70b-versatile with llama-3.3-70b-versatile

Note that Groq doesn't support the max. context lengths of 128K on the free tier. At least I get an API error if I exceed ~6000 tokens:

```
 Request too large for model deepseek-r1-distill-llama-70b in organization XYZ service tier on_demand
 on tokens per minute (TPM): Limit 6000, Requested 16955, please reduce your message size and try again.
 Visit https://console.groq.com/docs/rate-limits for more information.
```

### DIFF
--- a/assistants.yml
+++ b/assistants.yml
@@ -8,7 +8,10 @@ assistants:
   language_model_api_name: claude-3-5-sonnet-20241022
 - name: Meta Llama
   slug: llama-3-70b
-  language_model_api_name: llama-3.1-70b-versatile
+  language_model_api_name: llama-3.3-70b-versatile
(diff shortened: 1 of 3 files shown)


krschacht:
@lumpidu I think the issue is that you still had an assistant that was referencing the llama model you removed. I updated the assistants file.

Can you confirm this now works for you?

PR #633, opened 2025-04-22
Update to use the latest version of the old Anthropic gem.

zackgilbert:
The [Anthropic gem that Alex wrote](https://github.com/alexrudall/ruby-anthropic) will be deprecated eventually in favor of the [new official gem](https://github.com/anthropics/anthropic-sdk-ruby), so in order to get Claude working, we need to update it to the latest version. This actually simplifies things a bit and we no longer need to include the initializer.


krschacht:
@zackgilbert Thanks for this! I enabled Ci and it looks like there is one failing test. I think that’s a legit failure and not a flappy test. Could you take a look?

krschacht:
This seems like a fine solution especially if this gem won't be long lived. When you removed the initializer, it was doing some copying of those attributes so you manually needed to copy them in the config. I don't recall off hand why this hack was needed with the gem but I remember going back and forth with Alex about some issues with the gem related to streaming so it may have been related to that. Regardless, merging since it is a clear step forward.

PR #634, opened 2025-04-23
Update dependencies, including nokogiri 1.18.8.

zackgilbert:
Do a bundle update to update dependencies.

Resolves https://github.com/AllYourBot/hostedgpt/pull/632 (and maybe some others).

### DIFF
--- a/test/controllers/authentications/password_test.rb
+++ b/test/controllers/authentications/password_test.rb
@@ -30,7 +30,7 @@ class AuthenticationsController::PasswordTest < ActionDispatch::IntegrationTest
       email: people(:keith_registered).email,
       password: "wrong"
     }
-    assert_response :unprocessable_entity
+    assert_response :see_other
     assert_match(/Invalid email or password/, flash.alert)
   end
 
@@ -39,7 +39,7 @@ class AuthenticationsController::PasswordTest < ActionDispatch::IntegrationTest
       email: "wrong@email.com",
       password: "wrong"
     }
-    assert_response :unprocessable_entity
+    assert_response :see_other
     assert_match(/Invalid email or password/, flash.alert)
   end
 end


krschacht:
@zackgilbert I got your other PR merged in and turned my attention to this one. I resolved the merge conflict now that first PR was in main but tests aren’t passing. It looks like there is an issue with the new nokogiri gem version on one of the architectures.

zackgilbert:
@krschacht Strange. I see it in Github/CI, but I'm not seeing that on my end. 

I've been trying to run the tests locally by doing:
```
docker compose up --build     
docker compose run base rails test 
```
I thought Docker was supposed to help standardize and resolve this issue? hehe. Any suggestions on what I should try to resolve this?

zackgilbert:
@krschacht Okay. Tried a bunch of things to get this to work (monthly via the help of Claude), but unfortunately, I wasn't able to get around the fact that some of the platforms just don't appear to have the required builds (specifically darwin-22, darwin-23). Not familiar enough to know if we should expect them or if we need to find a different compatible version of nokogiri to try to lock to?

I'm assuming we don't want to adjust the platforms we are building for, correct?

It'll just prevent us being able to upgrade things until we sort it out.

krschacht:
@zackgilbert To be honest, I’ve never really understood what those different platforms are and why these dependency issues. My vague understanding though is that we need to support two platforms—x86 and arm-macos—so that the app can run successfully run outside of a docker image and inside of the docker image. These errors are part of the reason the gems are old as I was never able to resolve them. :(

But I just took this discussion and the error output and dropped it all into the new chatgpt-o4 and i got the clearest explanation I’ve ever read, and it suggests the solution:

Bundler is choking because your Gemfile.lock claims to support a bunch of platforms (arm64-darwin-23, x86_64-darwin-22, etc.) for which Rubygems doesn’t have a precompiled nokogiri-1.18.8—and in your Alpine container it also sees x86_64-linux-musl. It then tries to find one gem that’s valid “for all resolution platforms” and gives up.

You can fix it two ways:

Add only the Linux/Musl platform and drop the macOS ones from your lock:

bundle lock --add-platform x86_64-linux-musl
bundle lock --remove-platform x86_64-darwin-22 x86_64-darwin-23 arm64-darwin-23 arm64-darwin-24

commit the updated Gemfile.lock, rebuild and the Alpine container will find the matching nokogiri-1.18.8-x86_64-linux-musl.

Or force the source gem to build so Bundler ignores platform‐specific gems:

RUN bundle config force_ruby_platform true \
 && bundle install

That tells Bundler “always use the generic (source) gem” and compile it inside your container, sidestepping the missing prebuilt binaries.

ME: if i drop the macos ones that means this docker container will not be able to be run on mac, right?

No. Removing the darwin platforms from your lockfile only tells Bundler “don’t try to install Mac‐specific gems” during the image build. Docker Desktop on macOS still runs your container in a Linux VM (amd64-linux), so the image will run fine. You’d only lose the ability to do a host-side bundle install on your Mac against those darwin platforms.

ME: explain that last sentence you just said

When I say “host-side bundle install on your Mac,” I mean running bundle install directly in your macOS environment (outside of Docker). If you drop the darwin platforms from your lockfile, Bundler won’t know to fetch or compile the macOS-specific gem versions when you do that local install, so you’d get errors or end up only with the Linux gems. Inside the Docker container (which always uses Linux), nothing changes.

—-

Conclusion: 

Or force the source gem to build so Bundler ignores platform‐specific gems:

RUN bundle config force_ruby_platform true \
 && bundle install

Can you try that? It would be amazing if this resolves it because it would also suggest a generalized solution to this problem.

zackgilbert:
@krschacht Welp, was able to resolve the build issues by changing the platforms it's building for. When I tried the `RUN bundle config force_ruby_platform true && bundle install` route, it ended up causing errors with Tailwind that I couldn't easily get around.

No clue why there are now system_tests failing. I tried it twice, so might be legitimately an issue? Not sure if it's related to the platform adjustments I made? Also, no clue if those changes affect others, but it seems like the darwin22-24 builds were just replaced by more generic darwin builds.

krschacht:
@zackgilbert I pulled this down and spent a little while debugging. It seemed like a legit issue with autofocus, but I couldn't figure out what broke it so I just changed the approach to something a bit more reliable. Tests now pass so I'll merge this in.

Thanks for handling the upgrade and wrangling that docker issue.

Issue #474, opened 2024-07-24
Figure out why single quotes stream in escaped and remove the hack


krschacht:

matthewbennink:
I ran into this same issue, stumbling upon the same turbo-rails issue, and tracked it down to the use of `Nokogiri` [here](https://github.com/AllYourBot/hostedgpt/blob/main/app/jobs/get_next_ai_message_job.rb#L133-L134).

If I broadcast the original `html` value, it works fine, but if I broadcast the `inner_html` value, then I get duplicate single quotes. The main difference between the two is that the original `html` value uses `&#39;` and Nokogiri replaces those with a regular apostrophe character, which then appears to get escaped by some later process. I tried tacking on `.html_safe` to the new `html` value, but that didn't seem to have any effect.

For example (text courtesy of ChatGPT) -
```
html content before Nokogiri (works): &quot;Jessica&#39;s cat loves to nap in the sun.&quot;
html content after Nokogiri (doesn't work):  "Jessica's cat loves to nap in the sun."
```

Given the `messages/_message.html` file is already returning what's being parsed by Nokogiri, a simple fix is to pass `layout: false` as another option to the `render` call, and then the top-level tag of the rendered HTML will be what you're looking for. I can submit a PR for that if you like, or feel free to use the diff below.

```diff
diff --git a/app/jobs/get_next_ai_message_job.rb b/app/jobs/get_next_ai_message_job.rb
index 076d427..da943a3 100644
--- a/app/jobs/get_next_ai_message_job.rb
+++ b/app/jobs/get_next_ai_message_job.rb
@@ -128,10 +128,9 @@ class GetNextAIMessageJob < ApplicationJob
         only_scroll_down_if_was_bottom: true,
         streamed: true,
         message_counter: message.index
-      }.merge(locals)
+      }.merge(locals),
+      layout: false
     )
-    dom = Nokogiri::HTML.fragment(html)
-    html = dom.at_id(dom_id message).inner_html
     message.broadcast_update_to message.conversation, target: message, html: html
   end
```

krschacht:
Dang @matthewbennink good tracking this one down! This one baffled me. I was convinced it was a bug deeper within turbo but the team wanted a simple repro from me and I never created that.

Could you tee up a PR? I'd really appreciate that.

[...]

PR #637, opened 2025-04-24
Fix issue where single quotes are doubled up in Turbo streams

matthewbennink:
See https://github.com/AllYourBot/hostedgpt/issues/474.

### DIFF
--- a/app/jobs/get_next_ai_message_job.rb
+++ b/app/jobs/get_next_ai_message_job.rb
@@ -128,10 +128,9 @@ class GetNextAIMessageJob < ApplicationJob
         only_scroll_down_if_was_bottom: true,
         streamed: true,
         message_counter: message.index
-      }.merge(locals)
+      }.merge(locals),
+      layout: false
     )
-    dom = Nokogiri::HTML.fragment(html)
-    html = dom.at_id(dom_id message).inner_html
     message.broadcast_update_to message.conversation, target: message, html: html
   end
 


krschacht:
@matthewbennink any idea about that failing test? I re-ran it once just to make sure it wasn’t a flappy test, and it seems to be a legit failure.

matthewbennink:
> @matthewbennink any idea about that failing test? I re-ran it once just to make sure it wasn’t a flappy test, and it seems to be a legit failure.

Well, I see the issue at least. It's important to only send the **inner** HTML to the client from the background job, which is what the Nokogiri logic was taking care of. I missed that the first time around because our own logic didn't depend on not sending the outer div. Thankfully your system tests caught it!

I'm trying to rework this to avoid Nokogiri and the best I've come up with thus far is to extract yet another partial so that we can just render the inner HTML content when needed. Does that seem like a reasonable solution to you?

(It's early and I'm not used to doing development within Docker, plus not being able to run the system tests locally has me doing CI-driven dev, but hopefully I'll have a good PR in place in the next week or so.)

krschacht:
@matthewbennink now that you tracked down the issue to the &#39; substitution, we could just go back to nokogiri but tack on a replace to convert the single quote back to &#39; ?

mattlindsey (changes requested):
See comment by Keith. Thanks!

PR #677, opened 2025-09-05
Create Images Feature

mattlindsey:
This seems to fix problems with #526

Updated line app/models/mesage/document_image.rb from:
  `documents.first.image_url(variant, fallback: fallback)`
to:
   `documents.last.image_url(variant, fallback: fallback)`
   
 I will do some more testing, but comments welcome.


ceicke:
I am so happy to see this! Thanks!

krschacht:
Hey @ceicke ! I'm curious to ask: are you (or your daughter) still using HostedGPT regularly? If so, how do you decide to use HostedGPT rather than switching to ChatGPT? This is the core product strategy question that I'm still grappling with.

ceicke:
@krschacht yes, we are totally still using it! For my personal needs, this is my main window into LLMs. In the workplace I use the desktop app of Claude, but mostly because it integrates nicely with MCP servers. TBH, I don't know why I still use it that way. Maybe it's a habit, but I still like that I am able to self-host it.

PR #700, opened 2025-09-14
Add share conversation feature

### LINKED ISSUE #34: Ability to share a conversation

mattlindsey:
Fixes #34 

This is a test of using the `claude-on-rails` gem.  It was expensive in tokens but did fairly well.  I did have to fix a few things, like the tests.

The 'share conversation' is a nice feature if we want to merge this.  I think we should take out all of the `claude-on-rails` files since it might not be desired in the long run.


mattlindsey:
This PR might mess up the app on phones.  I changed the setting in Chrome and the icons seem to disappear. I'll look into that.

PR #713, opened 2025-09-27
Add simple logger (logcraft)

mattlindsey:
Make rails log easier to read.  Update gems.


mattlindsey:
@krschacht Do you think this might end up causing problems for Cursor or Claude Code by changing/simplifying the development logging output?

PR #719, opened 2025-09-30
Add token cost editing/adding via the UI

ceicke:
Add the possibility to add/edit the input/output token cost per model via the UI.


mattlindsey:
@ceicke See issue https://github.com/AllYourBot/hostedgpt/pull/719 too for reasoning.

ceicke:
Ah, interesting. I wasn‘t aware. But is that really true about different pricing per token per tier/region? I only looked at the docs of Claude amd OpenAI and only find one price per input/output per mTokens. But I only looked at it with that assumption and never challenged it. So could very well be that its different for different users. But even then, should we not keep it? If the users have different rates per mTokens, then that can be entered too…

ceicke:
Yeah, you are both right. Especially with these cached tokens. I will close the PR. Thanks for the good discussion!

PR #738, opened 2026-07-10
Add in rubyllm ai

strivedi183:


strivedi183:
I was just experimenting trying opencode, I meant to open it on my own fork. Apologies!

strivedi183:
I have been using RubyLLM in my projects and I was just experimenting to see what a potential replacement in HostedGPT would look like

strivedi183:
Just in case anyone is interested, I used opencode with the compound engineering plugin.

mattlindsey:
This seems like a good idea if you have any interest in reopening it.

strivedi183:
Thanks @mattlindsey , really appreciate it. 

I would be interested in trying to get RubyLLM in, but I can you see with the AI spike, it is pretty massive.

Do you think maybe there is a gradual way to get it in? I'm not sure if you have any ideas. Maybe behind a flag? Or maybe it is used to replace one service at a time? I am not sure as of yet.

Or I can re-open this PR, but it affects so much of the app, I would hope there could be a piecemeal way, but maybe not and just do it as a large PR.

I'm open to any thoughts or suggestions, thank you for following up!

mattlindsey:
@strivedi183   I think it would be great to do it as one big swoop and test it well. And it would be nice and clean to remove all unnecessary code at the same time. @keithschacht who created this might have a different opinion although he's busy on a newer project

keithschacht:
It's not a crazy idea to switch to RubyLLM, although it might be worth being clear on what it gets you over the existing SDK. I suspect there are some important things, it's just generally worth stating those and confirming they're missing and RubyLLM has them before investing the effort.

When it comes to implementing, this kind of refactor is typically done best as a parallel path strategy. Based on some kind of toggle (i.e. an ENV variable: USE_RUBYLLM=true), this would trigger the app to branch all the controllers to a new set of hits (or branch in the model or wherever is the easiest place to branch).

That way, you keep the old code paths and SDK alive until the RubyLLM branch has feature parity. This will probably make your life easier while implementing it because you when you run into issues with all this new code you can ask yourself, "Was this working in the old SDK code path (i.e. I broke it) or has it been broken for awhile?" Being able to flip the ENV back and forth let's you test old behavior vs new behavior.

mattlindsey:
@strivedi183 I think better support of adding new vendors and models would be an advantage of RubyLlm. It seems designed for a project like this. Can you think of other advantages?
Let me know if you want to work on this with a toggle and maybe break it into steps. I'm happy to comment and review them.

strivedi183:
@mattlindsey One of the advantages for using RubyLLM would be new vendors models for sure and I think there are a few other advantages.

But @keithschacht brought up a good point to compare it to the existing SDKs. The one that is not really supported is OpenAI Realtime or any speech to speech model. I am not sure if HostedGPT is planning on speech to speech but RubyLLM doesn't seem to support that yet.

I'll work on this. I think ENV toggle is a solid plan and I'll look to breaking it down into steps. I'll most likely open a separate draft PR.

And I'll do a deeper dive between RubyLLM and the existing SDKs in case there are other features that are unsupported by RubyLLM.

mattlindsey:
@strivedi183.  Cool. I created issue #740.

Issue #740, opened 2026-07-21
Migrate to RubyLLM


mattlindsey:
Spike has been done and step-by-step migration is planned (and already being worked on) with a toggle to go back and forth from current implementation to new RubyLLM for the current HostedGPT features.  
Probably first step should be text conversations with the currently supported sources.  Comments welcome!

mattlindsey:
Some thoughts/ideas:
- Toggle should be easily accessible in the UI, probably in Settings. Not flag requiring app restart. Have a note that switching to 'new' might not have all features yet.
- While developing, no current tests should be skipped or removed without good reason, and running bin/ci should test both old and new code. Maybe have duplicate test:system tests (or use newer method of controller tests, see #729)

mattlindsey:
Some thoughts on a phased plan, informed by why #738 stalled (it replaced everything in one shot, deleting all provider backends/tests):

**1. Toggle mechanism — reuse `Feature`, not a new mechanism**
`Feature.<name>?` already supports a global default in `config/options.yml`, overridable per-user via preferences with no restart (same pattern as the `voice` feature). Add e.g. `use_rubyllm` there and surface it as a Settings toggle, with a note that the new backend may not have all features yet.

**2. New backend lives *beside* the old ones**
Add `AIBackend::RubyLlm` in `app/services/ai_backend/`, implementing the same interface the existing classes do (`self.client`, `self.test_execute`, `set_client_config`, `client_method_name`, `stream_handler`, `preceding_conversation_messages`). `APIService#ai_backend` picks the class, checking the feature flag before falling back to the existing driver dispatch. `AIBackend::OpenAI/Anthropic/Gemini` stay untouched.

**3. Add a `TestClient::RubyLlm` stub**
Following the existing `test/support/test_client/` pattern, so tests stay deterministic without hitting real APIs.

**4. Phase the work into small, separately reviewable PRs**
1. Scaffolding only: gem, initializer, feature flag + Settings toggle, empty `AIBackend::RubyLlm` shell, CI green.
2. Plain text chat, OpenAI only — no images, no tools. Streaming + token counting parity. New test file alongside the existing OpenAI tests (untouched).
3. Anthropic + Gemini/Groq text chat, same shape.
4. Image/PDF attachment parity.
5. Tool/function calling parity via `Toolbox` — likely the riskiest phase since RubyLLM's tool-call streaming shape differs from each provider's raw API.
[...]

PR #742, opened 2026-07-24
Support attaching multiple images to a single chat message

### LINKED ISSUE #78: Add support for multiple images attached to a single chat
https://platform.openai.com/docs/guides/vision/multiple-image-inputs

mattlindsey:
The message/document data layer already supported multiple documents per message (has_many :documents, all three AIBackend providers already looped over them), but the composer UI, upload controller, and message display only ever handled one.

- Message::DocumentImage: replace .first/.last helpers with aggregate has_document_image?/has_document_pdf?/has_documents? checks; the view now reads per-document data straight off Document.
- Composer: file picker accepts multiple files, preview box becomes a cloneable template instead of one static slot.
- image_upload_controller.js: track an ordered list of attached files. Uses a separate, unnamed <input multiple> purely to trigger the OS file dialog, keeping the real Rails-nested-attributes inputs single-file always — a multiple file input left empty submits a blank value instead of being omitted, which would 422 every send.
- Message display: loop over documents, giving each image its own modal controller/dialog instead of sharing one per message.
- Add Gemini multi-image payload test to match existing OpenAI/ Anthropic coverage.

Closes #78

### DIFF
--- a/app/javascript/stimulus/image_upload_controller.js
+++ b/app/javascript/stimulus/image_upload_controller.js
@@ -1,7 +1,7 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
-  static targets = [ "file", "content", "preview" ]
+  static targets = [ "file", "picker", "content", "previewList", "previewTemplate", "fileInputs" ]
 
   connect() {
     if (!this.hasFileTarget || !this.hasContentTarget) {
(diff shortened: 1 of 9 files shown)


mattlindsey:
This was done entirely by Claude Code and works for OpenAI but it seems like Anthropic vision might not be working (even before this PR), so probably that should be investigated first.  Also I couldn't manually verify Gemini (api key permission problems).

And I probably shouldn't be adding a new major ai feature before we do #740.

mattlindsey:
Actually this PR works for OpenAI and Anthropic (user error with key).  So, for example, you can add 2 images in 1 message and ask it to compare them.

Still need to figure out key problem with Gemini.  The new key process is complicated.

P.S.  This works now for OpenAI, Anthropic, and Gemini.  So I'm taking off of Draft.

PR #744, opened 2026-07-26
Add RubyLLM as alternative AI backend behind feature flag.

strivedi183:
Adds ruby_llm gem (~> 1.16) as an alternative backend that routes all three providers (OpenAI, Anthropic, Gemini) through RubyLLM's unified Chat API when RUBYLLM_FEATURE is enabled. When disabled (default), the app is byte-for-byte identical — 715 tests pass unchanged, no existing code modified.

This is a proposed solution for issue #740 and supersedes the PR #738 

In order to test this, it would require importing the RubyLLM models.json

`RUBYLLM_FEATURE=true bin/rails models:import_rubyllm`

We are mapping the models.json to LanguageModel for the 3 providers but we can expand it to more after we verify that this has 1-to-1 parity with the existing paths with the provider SDKs.

@fluxgravity can you take a look at this before we ask for a review?

This most likely requires a lot of changes and is more of a starting point for a discussion on using a Feature flag and making RubyLLM an adapter, similar to the existing provider SDKs.


mattlindsey:
Quick notes:
- models.json is huge.  Does it have all 800+ models from RubyLLM?  Do we want to do that?
- Image tool is updated to use RubyLLM paint.  That is beyond scope of this PR, but I guess if it's a simple update it's OK. Right?

fluxgravity:
@strivedi183 Thanks, I think this is a good spike on how RubyLLM would work alongside the SDKs, showing that the feature flag path is possible. We'll probably want to break it up into smaller chunks, especially for posterity. I'm going to put together a phased plan we can work from and use to guide coding agents, which can reference this work and PR #738. I think @mattlindsey's suggestion in issue #740 is a good starting point, so I'll flesh it out further. 

@mattlindsey While this PR shows the feature flag path, as recommended by @keithschacht. Do you think we should still pursue a user-facing toggle?

Rather than an app-wide setting per user table, I'm curious whether we can leverage the User's `preference` attribute to track the feature. We can still keep the feature flag in `config/options.yml` and use it to set a default on a User, but the backend would likely need a different implementation than the app-wide `Feature.rubyllm?`. Just looking for a gut check before putting energy toward it.

mattlindsey:
@fluxgravity I like having new phased plan.  I hate to add too many new features with this effort, but a think a UI "General" settings screen (like ChatGPT has) where a User's preferences (such as current vs. RubyLLM) is a good idea.  I'm not sure if the `Users.preferences` field is even used currently or has a place to set/change actual preferences.

mattlindsey:
P.S.  Looks like there's a 'dark mode' preference.  If there's a place to set that, use it.  Otherwise a new page.
I have a gut feeling the new UI stuff will be useful.

Also using a feature flag in settings.yml as default is as intended, I think.

strivedi183:
This PR was a spike to see what it would look like to use the Feature flag with an ENV variable. I think the Feature flag was really useful, but I think that this spike showed some issues that could be worth thinking through. In order to avoid breaking any existing functionality, I tried only to add code, and minimize changing ore deleting code.
 
@mattlindsey You bring up some good points with two earlier points:

> Quick notes:
> 
> * models.json is huge.  Does it have all 800+ models from RubyLLM?  Do we want to do that?
> * Image tool is updated to use RubyLLM paint.  That is beyond scope of this PR, but I guess if it's a simple update it's OK. Right?

So for the 1st one, agreed, models.json is way too big to be similar to the existing models.yml

RubyLLM has a refresh feature that saves to the database [Model.refresh!](https://rubyllm.com/rails/#with-model-registry-default-for-new-apps) but it requires having their database schema with their Rails integration, which is beyond the scope of this PR.

It could be work looking at adapting or adding that method to LanguageModel so it is directly brought in and saved to the database.

For the 2nd one with RubyLLM paint replacing the Image tool, I agree that it is beyond the scope of this PR, it was more just to see if it was possible using the Feature flag, but I think a follow up separate PR would be a better place.

But after doing this spike, I think that having RubyLLM as its own backend adapter without removing the other SDK adapters might be the way to integrate and it is definitely should be done with a Feature flag with an ENV variable. If current SDKs get new features that RubyLLM doesn't support, they can just switch back to the provider SDK.

The user preferences is interesting next step after we get the Feature flag working. 

But this PR as is, should not be merged, and I think a new PR would be better than trying to get this one cleaned up.

mattlindsey:
@strivedi183  Now I'm thinking that the current model import rake job should be updated to get all of the models from RubyLLM, since then the user wouldn't have to enter any of them.  And the UI appears to only show relevant models (from defined services).

PR #753, opened 2026-08-09
Add Brave as a new API service driver

mattlindsey:
## Summary
Adds a "brave" driver to `APIService` for the Brave Search API, following the same env-var-fallback pattern used by OpenAI/Anthropic/Groq: if `BRAVE_API_KEY` is set as an environment variable it's used as the default token; if a token is entered in the Settings form, that takes precedence.

- `APIService`: new `URL_BRAVE` constant, `brave` driver, `requires_token?`, and an `effective_token` fallback to `Setting.default_brave_key` (sourced from `BRAVE_API_KEY`) that applies unconditionally — unlike the LLM default keys, which are gated behind `Feature.default_llm_keys?`.
- Settings UI: Brave gets setup instructions like the other official providers, but no "Test" link since it has no `AIBackend`/chat backend (`test_api_service` now returns a friendly error instead of raising when `ai_backend` is nil).
- New users get a blank Brave `APIService` seeded on registration (`User::Registerable`), and a migration backfills one for existing users.

This is infrastructure only — no `Toolbox::BraveSearch` tool is added in this PR; a future search tool can consume the key via `api_service.effective_token`.

## Test plan
- `bin/rails test` — 679 runs, 0 failures
- `bundle exec rubocop` on touched files — no offenses

### DIFF
--- a/app/helpers/settings/api_services_helper.rb
+++ b/app/helpers/settings/api_services_helper.rb
@@ -1,7 +1,7 @@
 module Settings
   module APIServicesHelper
     def official?(model)
-      openai?(model) || anthropic?(model) || groq?(model) || gemini?(model)
+      openai?(model) || anthropic?(model) || groq?(model) || gemini?(model) || brave?(model)
     end
 
     def openai?(api_service)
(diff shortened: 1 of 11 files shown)


mattlindsey:
Fixes #751

PR #754, opened 2026-08-10
Add Toolbox::BraveSearch and fix retired Claude model ids and broken Anthropic test button

mattlindsey:
## Summary
- Anthropic retired the `claude-3-5-sonnet-20241022` / `claude-3-5-haiku-20241022` snapshots that `models.yml` listed as the "(latest)" Claude 3.5 entries. Any assistant on one of those started failing every message with a 404 `not_found_error`. Swapped them for the current Claude 5 family ids (`claude-sonnet-5`, `claude-opus-5`, `claude-haiku-4-5-20251001`).
- Fixed `AIBackend::Anthropic.test_execute` (the Settings "Test" button). `ruby-anthropic`'s `Client#messages` only reads `args[:parameters]` and silently drops any other top-level keys, so passing `model:`, `messages:`, and `system:` alongside `parameters: { max_tokens: 1000 }` sent Anthropic a request with no model and no messages, returning a 400. Nested everything under `parameters:`, matching how the real streaming code (`set_client_config`) already builds its request.

## Test plan
- Verified the new model ids directly against the live Anthropic API.
- Reproduced the original bug locally: two existing `LanguageModel` rows pinned assistants to the dead `claude-3-5-sonnet-20241022` id; repointed them at the new `claude-sonnet-5` model and confirmed `stream_next_conversation_message`'s code path resolves the corrected id.
- Ran `LanguageModel#test` (the Settings "Test" button path) against the live API post-fix and got back a real assistant reply instead of a 400.


mattlindsey:
Accidental added addition of `Toolbox::BraveSearch < Toolbox` to this PR.

mattlindsey:
Fixes #752

PR #780, opened 2026-08-26
Move image generation to aibackend

strivedi183:
## Move image generation into the AI backend layer

Groundwork for [#740](https://github.com/AllYourBot/hostedgpt/issues/740) so it gives the RubyLLM migration a clean seam for images ([#775](https://github.com/AllYourBot/hostedgpt/pull/775) Phase 4).

## Summary

`Toolbox::Image` built its own `OpenAI::Client`, resolved the OpenAI `APIService` itself, and carried the comment *"For all backends, use OpenAI client for image generation"* - provider policy living as prose inside a tool. The provider call now lives in the backend layer:

- `AIBackend::OpenAI.generate_image` owns the call: service lookup, key validation, request parameters, and a raw `{ b64_json:, model: }` result. The tool payload strings stay in the toolbox.
- `AIBackend` defines a delegation default so **every backend answers the image question** - today's "always OpenAI" policy stated once as executable code. Backends with native capability override it: when Anthropic or Gemini ships image generation, or RubyLLM's `.paint` arrives, the swap is one method override in one file, and the toolbox never notices.
- `Toolbox::Image` now asks the current assistant's backend and names no provider. Behavior is preserved: images still always use OpenAI (`gpt-image-1`, same parameters, same payload, same error text).
- `TestClient::OpenAI` gains `images.generate` following its existing class-level-stubbable pattern, and the "OpenAI API key not found" error message is asserted for the first time.

## Follow-ups

- Routing images by per-user backend choices and the `RubyLLM.paint` implementation arrive with #775 Phase 4 - this PR builds the slot they plug into. Until then, images remain OpenAI-only.

PR #781, opened 2026-08-27
Autotitle json intent refactor

strivedi183:
## Make conversation titles provider-blind

Groundwork for [#740](https://github.com/AllYourBot/hostedgpt/issues/740) — gives the RubyLLM migration a clean seam for titles ([#775](https://github.com/AllYourBot/hostedgpt/pull/775) Phase 2+), the same treatment images got in #780.

## Summary

`AutotitleConversationJob` picked request parameters by comparing `ai_backend.class ==` against three concrete classes and appended an Anthropic-only prompt suffix by string-comparing the driver, provider policy living as conditionals inside a job. Titles now flow through one seam:

- `get_oneoff_message` gains a trailing `json:` keyword, threaded like `streaming:`. The job makes one call; the class branches, per-provider param literals, driver checks, and the dead regex-scrape fallback are deleted.
- `AIBackend::OpenAI` folds `response_format: { type: "json_object" }` into its parameters when intent is set, before the caller-params merge. Explicit parameters still win last. Any backend extending OpenAI inherits this with zero extra code, pinned by a direct-subclass test.
- `AIBackend::Gemini` injects `generation_config` inside its existing override; `AIBackend::Anthropic` appends a mechanism-only "ONLY valid JSON" coercion to its system slots internally, sends no request param, and leaves the reply schema to the caller: the title prompt already demonstrates it, and OpenAI/Gemini never carried a suffix.
- Duplicate-spend guard: a conversation that already has a title exits before any provider call. Rapid consecutive touches no longer double-bill.
- Graceful failure: a reply that can't yield a non-blank-String topic (blank, prose-wrapped, malformed, wrong shape, safety-blocked) warns with the conversation id and keeps the prior title. Provider network failures and stalls (one-off calls bounded at 30s) degrade the same way. Titles are cosmetic — they should never be queue failures.
- `TestClient` doubles now record received arguments and answer dig-able envelopes, so the Anthropic and Gemini one-off paths, previously untestable, are covered — including a first Anthropic one-off test.
- Behavior is preserved: the refactor commits are wire-identical per provider, with request shapes pinned by recorded arguments. The one deliberate wire change, in the final commit, trims Anthropic's suffix to mechanism-only, matching OpenAI and Gemini, which never carried a schema suffix.
- Every commit is green on its own (full suite verified at each), so any improvement commit can be reverted independently.

## Follow-ups

- Routing titles through per-user backend choices and the RubyLLM implementation arrive with #775 — this PR builds the slot they plug into: its one-off override should adopt the trailing `json:` keyword when that phase lands (heads-up posted on the PR).
- New backends extending the OpenAI dialect inherit the json handling automatically; the subclass test pins that guarantee.

_NOTE: The reason there are a ton of commits is to make it easier for me to review because AI loves giant commits that make a bunch of changes, and give a vague description._

### DIFF
--- a/app/jobs/autotitle_conversation_job.rb
+++ b/app/jobs/autotitle_conversation_job.rb
@@ -37,22 +37,27 @@ class AutotitleConversationJob < ApplicationJob
   end
 
   def usable_topic(response)
-    return nil if response.blank?
+    if response.blank?
+      Rails.logger.warn("[AutotitleConversationJob] Unusable reply for conversation #{@conversation.id}")
+      return nil
+    end
 
-    topic = JSON.parse(response)["topic"]
+    document = JSON.parse(response)
+    topic = document.is_a?(Hash) ? document["topic"] : nil
+
     if topic.is_a?(String) && topic.present?
       topic
     else
       Rails.logger.warn("[AutotitleConversationJob] Unusable reply for conversation #{@conversation.id}")
       nil
     end
-  rescue JSON::ParserError, TypeError => e
+  rescue JSON::ParserError, TypeError, NoMethodError => e
     Rails.logger.warn("[AutotitleConversationJob] Unusable reply for conversation #{@conversation.id}: #{e.class}")
     nil
   end
 
   def system_message
-    base_message = <<~END
+    <<~END
       You extract a 2-4 word topic from text. I will give the text of a chat. You reply with the topic of this chat,
       but summarize the topic in 2-4 words. Even though it's not a complete sentence, capitalize the first letter of
       the first word unless it's some odd anomaly like "iPhone". Make sure that your answer matches the language of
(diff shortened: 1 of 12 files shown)


strivedi183:
Context for anyone connecting this to the RubyLLM work:

- @fluxgravity - heads-up since #775 is affected here: Phase 2's `AIBackend::RubyLLM#get_oneoff_message` override should accept the new trailing `json:` keyword (default off) and forward it however RubyLLM wants. Your existing `with_params` passthrough covers the OpenAI-style cases. Nothing breaks on main before then (RubyLLM routing never reaches title requests until your routing lands), but titles in RubyLLM mode will want it from day one. Happy to pair on the rebase.
- One note on the removed regex fallback: it looked like a safety net but was unreachable (exactly three backend classes exist), so it's deleted rather than ported.

mattlindsey:
Looks good. Thanks!

strivedi183:
Thanks you @mattlindsey ! It's ready to merge.

PR #787, opened 2026-08-30
Groq backend error contract

strivedi183:
## Give Groq its own backend

Groundwork for [#740](https://github.com/AllYourBot/hostedgpt/issues/740). Continues the provider-policy-out-of-jobs work from #780 and the title refactor, in the same shape.

## Summary

Groq services ride the OpenAI driver, so they got the OpenAI backend everywhere, and provider-specific behavior lived in other files as string comparisons against the service *name*, which users can edit. Backends now own their error facts, and Groq owns a backend:

- `AIBackend::Groq` subclasses `AIBackend::OpenAI` and carries Groq's own error contract: Llama-flavored key-error copy and `console.groq.com/keys`. This fixes the live bug where Groq users hitting quota errors (routine on Groq's free tier) were pointed at OpenAI's billing page.
- Each backend owns its error facts as class-level methods (`key_error_message`, `billing_url`) with a generic fallback on the base. `GetNextAIMessageJob` rescues one unified `AIBackend::ConfigurationError` and renders what the current backend declares; the job's per-provider copy methods, name-string branching, billing-URL case map, and the Gemini error monkey patch are deleted. Provider-blind failures (blank response, connection error) stay in the job.
- Backend selection checks the driver and canonical URL pair: a service on Groq's URL gets `AIBackend::Groq`; OpenAI services and openai-dialect custom URLs (local model servers etc.) resolve exactly as before.
- `URL_GROQ` joins the `requires_token?` allowlist, so a blank Groq key now fails fast with the friendly error at chat, at title generation, and at the settings Test button instead of a doomed provider call.
- Tool policy moves out of `LanguageModel`: the model keeps its per-model `supports_tools` attribute and asks its backend for the provider policy. Groq's backend pins the answer to denied until Groq tool use is debugged. The class comment records the two deliberately untaken future paths (a dedicated Groq SDK, with the community gems and their trade-offs, and a first-class `:groq` driver, including the data migration of existing rows and the maintainer discussion that entails).
- Behavior changes beyond Groq's own copy, stated up front: services whose name was changed, or that use a custom URL, now receive their driver backend's standard error message instead of the old message that interpolated the service name, and the tools gate now consults each backend's provider policy, so configurations that previously bypassed the name-based gate may see tool calls declined. Copy for default-named services is byte-identical; Anthropic's backend additionally translates the gem's own configuration error (raised directly for nil tokens on custom-URL services).
- Groq inherits `response_format` JSON handling from the OpenAI backend. This is pre-verified: JSON Object Mode is supported on all Groq models, and the title prompt meets its prompt requirement. The title job was publicly tested against Groq in #583.
- Every commit is green on its own (full suite verified at each), so any commit can be reverted independently.

## Follow-ups

- Groq tool support: Groq's docs say all hosted models support function calling. The backend's pinned denial stays until someone debugs this app's tool loop against Groq; the intent is recorded on the class.
- Groq still rides the `openai` driver with driver+URL dispatch; a first-class `:groq` driver (enum value, data migration, settings row) is deliberately left for a maintainer discussion. The new backend class documents this path and its migration lineage.

### DIFF
--- a/app/jobs/autotitle_conversation_job.rb
+++ b/app/jobs/autotitle_conversation_job.rb
@@ -27,9 +27,7 @@ class AutotitleConversationJob < ApplicationJob
     @conversation.update!(title: topic)
   rescue Faraday::Error,
          ::Timeout::Error,
-         ::OpenAI::ConfigurationError,
-         ::Anthropic::ConfigurationError,
-         ::Gemini::Errors::ConfigurationError => e
+         AIBackend::ConfigurationError => e
     Rails.logger.warn("[AutotitleConversationJob] Provider failure for conversation #{@conversation.id}: #{e.class}")
     true
   end
--- a/app/jobs/get_next_ai_message_job.rb
+++ b/app/jobs/get_next_ai_message_job.rb
@@ -1,8 +1,6 @@
 include ActionView::RecordIdentifier
 require "nokogiri/xml/node"
 
-class ::Gemini::Errors::ConfigurationError < ::Gemini::Errors::GeminiError; end
-
 class GetNextAIMessageJob < ApplicationJob
   include ActionView::Helpers::RenderingHelper
   class ResponseCancelled < StandardError; end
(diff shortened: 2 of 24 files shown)


strivedi183:
Context for anyone connecting this to the RubyLLM and tooling work:

- @fluxgravity, heads-up since #775 is affected here: Phase 5 reworks the job error branches this PR also touches. This error contract lands first, and your Phase 5 can rebase onto it. The job's provider arms you'd have replaced now collapse into a single `AIBackend::ConfigurationError` rescue, and the routing side of the house (driver plus canonical URL pair in `APIService#ai_backend`) mirrors how your `provider_for_url` resolves providers inside RubyLLM. Happy to pair on the rebase.
- @mattlindsey, this is #740 groundwork in the same shape as #780: provider policy out of the job and out of the model, into the backends.
- One note on the tools gate: the old name-string check meant a renamed Groq service could silently regain tool calls. The backend pins that denial until the tool loop is debugged; Groq's docs say all its models support function calling, so this is a deliberate hold with the intent recorded on the class.
- One note on the new backend selection: it checks the driver and the canonical Groq URL. A Groq service whose URL was hand-edited still resolves to the OpenAI backend. The plan is to wait until Groq needs anything outside OpenAI compatibility, and elevate Groq to its own driver then; that is recorded as a future path on the class and is a maintainer discussion, not something this PR decides.

mattlindsey:
Looks good!

Issue #433, opened 2024-06-24
Implement OpenRouter backend for better support of OSS LLMs


krschacht:
https://github.com/obie/hostedgpt/commit/250467d887a5c268e42fa79d1cdf606020b43aa7

PR #788, opened 2026-08-31
Add openrouter backend

strivedi183:
## Add OpenRouter support

Groundwork for [#433](https://github.com/AllYourBot/hostedgpt/issues/433). Follows the Groq precedent from #787 in the same shape: a new OpenAI-dialect backend, driver+URL dispatch, and per-provider facts owned by the backend.

~~**This PR is stacked on #787 (Groq backend error contract) and depends on it: the branch builds on #787's error-contract seams, so until that PR merges, the commits tab here shows #787's 9 Groq commits plus this PR's 5 OpenRouter commits. The OpenRouter work is the last 5 commits; once #787 merges, this diff collapses to just those.**~~ (merged)

## Summary

OpenRouter aggregates hundreds of models behind a single OpenAI-compatible endpoint, so it rides the `openai` driver the way Groq does:

- `AIBackend::OpenRouter` subclasses `AIBackend::OpenAI` and carries OpenRouter's error contract: its own key-error copy and `openrouter.ai/credits` for quota errors, instead of OpenAI's billing page. It also sends OpenRouter's optional attribution headers (`HTTP-Referer`, `X-Title`) so requests are attributed to this app in OpenRouter's rankings. `TestClient::OpenAI` mirrors the real client's `add_headers`, so the header behavior is asserted in tests rather than guarded out of them.
- Backend selection checks the driver and canonical URL pair: a service on OpenRouter's URL gets the OpenRouter backend; OpenAI services and openai-dialect custom URLs resolve exactly as before.
- `URL_OPENROUTER` joins the `requires_token?` allowlist, so a blank OpenRouter key fails fast with the friendly error at chat, at title generation, and at the settings Test button.
- The default-key ladder falls back to a shared `DEFAULT_OPENROUTER_KEY` when `default_llm_keys` is enabled and the user has no personal token. The variable is passed through in `docker-compose.yml` and listed in the README key list, per the convention every other provider follows.
- New users get an OpenRouter service on registration; a data migration provisions the same service for existing users. A user who pre-connected OpenRouter through the settings UI under a custom name is adopted and renamed to the canonical name (their token is untouched), so the seeded models bind for them too; a service at that URL with a different driver is left alone and the canonical row is created beside it. The migration is forward-only: `down` raises `IrreversibleMigration` per the AddGroq precedent, since language models hold a non-cascading foreign key to the service.
- `models.yml` gains the top 25 most-popular text-output models from OpenRouter's live API ranking (image-input support recorded per model). Routed models are distinct rows from native-provider models, since model IDs differ (`openai/gpt-4o` vs `gpt-4o`).
- OpenRouter joins the official providers in the settings form, with a four-step "How do I get?" instructions block (English and German), following the Groq precedent.

## Follow-ups

- **RubyLLM:** this PR needs no RubyLLM; OpenRouter works through the built-in SDK today. The settings "AI backends" choice (SDK vs RubyLLM, #775) derives its options from drivers, so once `AIBackend::RubyLLM` supports the `openai` driver, every OpenRouter service picks up the RubyLLM choice automatically, and RubyLLM speaks `:openrouter` natively there. Nothing in this PR blocks or duplicates that work.
- Image generation resolves its token via `find_by(driver: :openai)`, which now also matches OpenRouter services; an OpenRouter-only user's image request can send the wrong key to OpenAI. Pre-existing ambiguity (the same was true of Groq before this PR); needs its own fix with a red-first test.
- OpenRouter's out-of-credits response (HTTP 402) has no dedicated branch in the job's error contract, so an exhausted user sees retries and a generic error instead of the quota copy + credits URL the backend already declares. Same shape likely covers 403 moderation flags. Worth a small follow-up.
- OpenRouter rides the `openai` driver with driver+URL dispatch. If a third OpenAI-dialect service needs its own backend, that is the moment for the first-class driver split discussion recorded in the seams roadmap.


strivedi183:
- **Depends on #787**: this branch is stacked on it and rebases cleanly onto `main` once it merges. `AIBackend::OpenRouter` subclasses the backend error-fact seams #787 introduced.
- @mattlindsey, this is #433 in the same shape as the Groq work in #787: an OpenAI-dialect provider gets its own backend, dispatch, and error contract rather than string comparisons. Nothing here touches the RubyLLM migration path; when #775 lands, OpenRouter services pick up the SDK-vs-RubyLLM choice automatically through the existing driver-derived settings control, and RubyLLM speaks `:openrouter` natively.
- One deliberate non-decision: OpenRouter stays on the `openai` driver with driver+URL dispatch. Elevating it to a first-class driver means a data migration of user rows and a maintainer discussion, the same call #787 made for Groq.
- On the gem question #433 raises: the OpenAI-compatible route was the easiest path to start with and needs no community gem. The full decision is recorded on the backend class.
- One note on the model list: the 25 models are a point-in-time snapshot of OpenRouter's popularity ranking (August 2026), captured from their public API. Refreshing is a manual `models.yml` edit + `rails models:import` for now.
- Live-verified with a real key before opening: settings save, Test button (success + 401), streamed chat with usage capture, and a structured tool-call round-trip.

mattlindsey:
Hi @strivedi183.  Is this ready?

strivedi183:
@mattlindsey It's ready for review. Let me know if you see anything or run into anything.

PR #789, opened 2026-09-02
Clean up and normalize the views

mattlindsey:
A consistency pass over every file in `app/views`, plus the handful of real bugs it turned up along the way.

## Structure

- Renamed the six `layouts/` partials that were `.erb` to `.html.erb`, matching every other view.
- Extracted the two byte-identical `test.turbo_stream.erb` files into `settings/_test_result`, written with `turbo_stream.replace` rather than hand-rolled `<turbo-stream>` markup.
- Replaced the duplicated inline row-click `<script>` in the two settings index tables with a `clickable-row` Stimulus controller — these were the only inline scripts left in the views. Behavior is unchanged (still `window.location.href`).
- Collapsed the six near-identical field blocks in `documents/_document` into a loop.
- Normalized `render partial: "x", locals: {…}` to the short form, and `<%= content_for … do %>` to `<% … %>`.

## Bugs found along the way

- `_head` linked `stylesheet_link_tag "application"` twice.
- `app.generic.yes` / `app.generic.no` never resolved. The keys actually live under `app.settings.generic`, **and** YAML parses bare `yes:` / `no:` as booleans, so they were unreachable by name either way. Quoted the keys and fixed the lookup — German now translates instead of silently falling back to the English `default:`.
- `conversations/index` had hard tabs and a stray `%` inside the composer's `data-action` string, emitting an invalid Stimulus action token.
- `_main_column` never closed its `<section>` (tags now balance 1:1).
- `id="modal-backdrop"` was rendered once per image message, colliding with the keyboard-shortcuts dialog that owns that id — and that `composer_test` reaches for via `querySelector`. Dropped from the per-message copy so the id is unique.
- Removed `dark:dark:` duplicate variants and empty `data: { }` / `class=""` attributes.

## Consistency

- Moved the last hardcoded English out of the views into `en.yml` / `de.yml` (Assistants, Show All, the share-button titles, the assistant-deleted sentence, two shortcut labels, the PDF link text, the documents form labels).
- Repaired ~60 lines whose indentation had been flattened to column 2 by an earlier i18n conversion.
- Single → double quotes, per the repo's RuboCop `EnforcedStyle: double_quotes`.

## Verification

- All 64 views compile through Rails' actual ERB handler.
- All 268 `t()` keys used in views resolve in both `en` and `de` (worth checking explicitly — `raise_on_missing_translations` is commented out, so missing keys render silently).
- `bin/rails test` — 809 runs, 0 failures, 0 errors.
- `bin/rails test:system` — 122 runs, 0 failures, 0 errors.

## Notes

- This branch also carries a few small in-progress tweaks that were already sitting in the working tree: the nav column moving from `fixed` to `absolute`, `pb-safe-or-4` on the nav footer, and the settings Cancel button relabelled to Back. Happy to split those out if you'd rather they land separately.
- Left alone deliberately: the `value: User.last.id` / `Assistant.last.id` / `Message.last.id` defaults in `documents/_form` are scaffold residue that query the DB from a view, but removing them changes the form's behavior. The whole `documents/` CRUD scaffold looks unused by the real UI — both feel like a separate decision.


mattlindsey:
Fixes #360

PR #797, opened 2026-09-09
Provider identity dispatch

strivedi183:
## Dispatch AI backends by provider identity

Follows the provider-facts-own-backends shape of [#787](https://github.com/AllYourBot/hostedgpt/pull/787) and [#788](https://github.com/AllYourBot/hostedgpt/pull/788) into the transport layer, and connects the AI-backend choice panel from [#788](https://github.com/AllYourBot/hostedgpt/pull/788) to dispatch.

## Summary

- **Dispatch resolves by provider identity.** A service's identity is the driver plus, for openai-dialect vendors, the canonical URL, and one registry (`APIService.sdk_backends`) maps each identity to the backend that owns its wire dialect, error copy, and provider policy. With the `use_ruby_llm` flag defaulting on, Groq and OpenRouter conversations previously routed to `AIBackend::RubyLLM` because both ride the `openai` driver, and error copy fell to the generic base. Quota errors rendered "You are using RubyLLM so go here " with a blank billing URL.
- **The stored backend choices go live.** An explicit per-identity choice (`ruby_llm` / `sdk`) always wins; silent users follow the site-wide flag, gated on RubyLLM supporting the identity. The choice panel renders a row for every chat provider (Groq joins, brave stays out), all radios selectable, the tri-state enforced server-side, and a form submit writes all choices in one save where each choice used to be a full model save.
- **Error facts survive the transport switch.** The job's error arms read key copy, billing URL, and provider name from the service's identity backend, so a Groq conversation served by RubyLLM still says Groq and points at `console.groq.com/keys`.
- **The RubyLLM transport honors compat-vendor configuration.** Contexts for non-canonical openai URLs set `openai_use_system_role`: Groq, OpenRouter, and custom servers expect the traditional system role the previous SDK always sent, while RubyLLM defaults to OpenAI's `developer` role. OpenRouter's attribution headers (`HTTP-Referer`, `X-Title`) ride only the OpenRouter identity, from one declaration shared with the SDK backend. A context-capturing test double asserts the wiring instead of assuming it.
- **Image generation resolves only the canonical OpenAI service**, closing the follow-up recorded on [#788](https://github.com/AllYourBot/hostedgpt/pull/788): `find_by` is scoped to the URL, so a Groq or OpenRouter row can no longer send its token to `api.openai.com`. The tool labels the image from the executor, and the key error is composed in the toolbox instead of the provider layer reaching through `Current.message`.
- **Along the same seams:** the generated-image attach plumbing moves out of the job into a `GeneratedImage` object with its first tests (a tool result with a nil payload used to wedge the turn); search deep links are owned by the tool, whose result carries `link_url`, rendered https-only with the legacy link bytes preserved; Groq and OpenRouter get their service logos and the llama name-sniff retires (it stamped the Meta logo on DeepSeek's distilled model); the settings sidebar renders the shared assistant avatar partial instead of an empty circle; the AI-backends radio columns align at any width and every cell is clickable; AGENTS.md, CLAUDE.md, and `.github/copilot-instructions.md` document the identity model and an add-a-provider recipe that no longer lands nil backends.

## Follow-ups

- **Native `:openrouter` is deliberately untaken.** RubyLLM serves OpenRouter over the openai-compatible path (base-URL override), which is what the tests pin; the gem's native `:openrouter` provider stays a one-line switch in `provider_slug` if OpenRouter-specific error parsing is ever needed.
- **OpenRouter's HTTP 402 quota arm**, carried over from [#788](https://github.com/AllYourBot/hostedgpt/pull/788)'s follow-ups, still open.
- **Groq tool calling stays pinned denied**, now transport-independent: the policy follows the identity whichever backend serves. Debugging this app's tool loop against Groq remains the gate.
- **The RubyLLM transport for Groq and OpenRouter is pinned by a context-capturing test double, not live keys.** The site flag already defaults to RubyLLM, so a real-key smoke across the five providers is the remaining gate before leaning on it broadly.
- **The registry and the identity mapping must stay in lockstep**: an unmapped driver yields nil backends silently. An invariant test covers a missing backend, and the registration touchpoints are enumerated on the class.

Let me know if anyone runs into any issues. This is meant to be more of tidy up PR and not to have any big breaking changes.


strivedi183:
- **This is the consumer half of [#788](https://github.com/AllYourBot/hostedgpt/pull/788)'s choice panel:** the per-identity choices that PR stored (with its disabled OpenRouter radio) go live here, and [#788](https://github.com/AllYourBot/hostedgpt/pull/788)'s recorded follow-up (image generation resolving its token by driver alone) is fixed with a red-first test.
- @mattlindsey, this continues the provider-facts-own-backends shape from [#787](https://github.com/AllYourBot/hostedgpt/pull/787)/[#788](https://github.com/AllYourBot/hostedgpt/pull/788) into the transport layer: dispatch by provider identity, per-identity choices that steer dispatch, and error facts that survive a transport switch.
- @fluxgravity, heads-up since [#775](https://github.com/AllYourBot/hostedgpt/pull/775)'s transport is what this builds on: nothing about the RubyLLM backend's public surface changes, but dispatch now consults stored per-identity choices before the site flag, contexts for non-canonical openai URLs set the system role your compat vendors expect, and the tool policy follows the identity rather than a URL check inside the backend.
- One deliberate non-decision: RubyLLM serves Groq and OpenRouter over the openai-compatible path; native `:openrouter` stays a recorded one-line switch.
- Another: Groq tool calling remains pinned denied: same gate as before, now enforced the same way under both transports.

Let me know if either you run into any issues.

mattlindsey:
@strivedi183 This looks good, however I get an error after asking Google Gemini to create an image.  I get the image back fine, but after the image where a comment should be I get the following error:
```
[image here]
(There is a configuration error with the Gemini API Service. Maybe you have an invalid API key? Click your Profile in the bottom left and then Settings and then API Services. You will find Gemini there.)
```

fluxgravity:
@strivedi183 -
## Review notes — nothing blocking, take what's useful

I ran a multi-persona review of this branch (9 lenses, findings deduplicated and independently validated). Headline: **no correctness or security defects were found** — the security-relevant changes look sound (canonical-URL scoping for image generation, https-only link sink, allowlisted backend choices, payload redaction preserved). Nothing below needs to block this PR; they're optional suggestions, and it's entirely your call which (if any) to take.

For each item: what it is, why someone might care, the fix if you want it, and when you might reasonably skip it.

### 1. Vendor name hard-coded in the shared RubyLLM transport — P2, optional

`app/services/ai_backend/ruby_llm.rb:129`

```ruby
chat.with_headers(**AIBackend::OpenRouter.attribution_headers) if @api_service.provider_identity == :openrouter
```

**The observation:** the branch's own design says per-provider facts live on the identity's backend class and dispatch never names a vendor. This one line names `AIBackend::OpenRouter` directly. It works correctly today — this is purely about the next vendor that wants per-transport headers having to discover `build_chat` on their own, since the add-a-provider checklist doesn't mention it.

**If you want it:** base-class default plus polymorphic dispatch —

```ruby
# AIBackend base
def self.attribution_headers; {}; end

# build_chat
chat.with_headers(**@api_service.sdk_backend.attribution_headers)
```

Same headers, byte-identical on the wire; the existing headers test plus an empty-hash case covers it.

**Reasonable to skip if:** you'd rather leave one explicit line than add an indirection for a single vendor.

### 2. `batch_merge` pre-fetch is redundant with `deep_merge` — P3, optional

`app/models/user/features.rb:42-43`

```ruby
feature = user.preferences[:feature] || user.preferences["feature"] || {}
user.update!(preferences: user.preferences.deep_merge(feature: feature.merge(updates)))
```

**The observation:** `deep_merge(feature: updates)` alone already gives `updates` precedence and preserves siblings, and the `["feature"]` fallback can never hit (`JsonSerializer.load` always symbolizes keys — JsonSerializer.rb:21-23). So the first line is dead work plus a branch suggesting a case that isn't real. Behavior-preserving either way.

**If you want it:**

```ruby
user.update!(preferences: user.preferences.deep_merge(feature: updates))
```

**Reasonable to skip if:** the defensive style feels safer to you than the proof that it's dead.

### 3. openai-family `api_key` stanza unpinned in RubyLLM context tests — P3, optional

`test/services/ai_backend/ruby_llm_test.rb:312-337`

**The observation:** the new Groq/OpenRouter/canonical-OpenAI context tests pin `openai_api_base` and `openai_use_system_role` but not `context.openai_api_key` — while the anthropic (:388) and gemini (:396) siblings assert theirs. Today's code is fine; the gap is only that a future typo in the `#{provider}_api_key=` assignment would pass the suite and fail authentication for every RubyLLM-served openai-family chat in production.

**If you want it:** add to each of the three new context tests (:312, :321, :330), matching the sibling pattern:

```ruby
assert_equal @assistant.language_model.api_service.effective_token, context.openai_api_key
```

**Reasonable to skip if:** you prefer the live-key smoke test as the credential gate.

### Decision item — no change proposed, just a question — P2

`app/models/api_service.rb:102` (also :59-60, :98-99, :131-134)

`provider_identity`, `requires_token?`, `logo_filename`, and `default_llm_key` each re-compare the same URL constants. This diff itself paid the tax (Groq/OpenRouter arms added to `provider_identity` and `logo_filename` in lockstep), and a new provider must touch five sites the AGENTS.md checklist doesn't document. Two directions, your call — or neither, now:

- **Consolidate:** one canonical identity table (driver, url, logo, default-key setting, token-required; brave included, flagged non-chat) and derive all four call sites from it.
- **Cheaper:** extend the AGENTS.md/CLAUDE.md registration checklist with the three extra call sites.

This is a design call, not a defect — fine to defer to a follow-up given #797 is framed as a tidy-up.

---

strivedi183:
> @strivedi183 This looks good, however I get an error after asking Google Gemini to create an image. I get the image back fine, but after the image where a comment should be I get the following error:
> 
> ```
> [image here]
> (There is a configuration error with the Gemini API Service. Maybe you have an invalid API key? Click your Profile in the bottom left and then Settings and then API Services. You will find Gemini there.)
> ```

Thanks for trying it @mattlindsey ! Gemini is the one I don't have that I can manually try (Groq is another) so I depended on the tests.

The error message is helpful, would you be comfortable sharing any logs with that? Maybe we can figure out what is happening from there. I'll look into it myself.

Thank you again!

mattlindsey:
Hi @strivedi183.  I used Claude to discover the problem where Gemini needs tool_signature passed.
Looks like we're not doing tools and tool calling exactly like the RubyLLM documentation says to do it, but that should be OK I think.
Do you think this PR is ready to merge with the fix? I think if there are any other minor problems we can fix them later.

strivedi183:
> Hi @strivedi183. I used Claude to discover the problem where Gemini needs tool_signature passed. Looks like we're not doing tools and tool calling exactly like the RubyLLM documentation says to do it, but that should be OK I think. Do you think this PR is ready to merge with the fix? I think if there are any other minor problems we can fix them later.

Thank you so much @mattlindsey ! It's good to merge with me!

PR #799, opened 2026-09-11
feat(rubyllm): generate images via Context#paint behind use_ruby_llm flag

fluxgravity:
## RubyLLM image generation behind the `use_ruby_llm` flag

Completes the optional Phase 6 item skipped in #775: `AIBackend::RubyLLM.generate_image` wraps a keyed `RubyLLM.context.paint` instead of inheriting the #780 delegation to `AIBackend::OpenAI.generate_image`. Same `{ b64_json:, model: }` contract and dispatch seam — flag off keeps the ruby-openai path byte-identical.

### Related PRs

- **#775** — completes its optional Phase 6 item
- **#797 (provider-identity-dispatch)** — both directions:
  - #797 scopes the old path to the canonical OpenAI URL; this branch keeps the name-based lookup, so reconcile both to URL-based selection when #797 merges
  - Until #797 lands, its stored per-backend choices are inert at dispatch, so image-gen routing here follows only the `use_ruby_llm` flag.

### Known trade-off (deferred to #797)

A renamed OpenAI service keeps image gen flag-off but gets the "OpenAI API key not found" error flag-on. The URL-based reconciliation above resolves it.

### Testing

- `bin/rails test`: 912 runs, 0 failures; old-path tests untouched.
- Live dogfood: image generated via a Claude chat; `images/generations` POST confirmed.
- Phase 7 (if implemented) can remove `ruby-openai`.

### DIFF
--- a/app/services/ai_backend/ruby_llm.rb
+++ b/app/services/ai_backend/ruby_llm.rb
@@ -5,6 +5,8 @@ class AIBackend::RubyLLM < AIBackend
   class ConfigurationError < AIBackend::ConfigurationError; end
   class ToolCallIntercepted < StandardError; end
 
+  IMAGE_MODEL = "gpt-image-1"
+
   CONFIGURATION_ERRORS = [
     ::RubyLLM::UnauthorizedError, ::RubyLLM::ConfigurationError,
     ::RubyLLM::BadRequestError, ::RubyLLM::ForbiddenError,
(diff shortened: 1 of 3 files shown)


fluxgravity:
@strivedi183 do you have any feedback on this? I deferred to your #797 for the URL-based selection refactor. If that gets merged first, then this branch will likely need a minor rebase tweak.

